PR #2, opened 2019-05-17
auto-publish attempt 1

tpolecat:
I think I followed the instructions at 

- https://github.com/olafurpg/sbt-ci-release

correctly. We'll see.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,22 @@ language: scala
 jdk:
   - oraclejdk8
 
-script:
-  - sbt headerCheck compile test
+before_install:
+  - git fetch --tags
+
+stages:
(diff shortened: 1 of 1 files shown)


cquiroz:
Why not buildkite?

tpolecat:
I don't feel strongly about it but I think what makes sense is

- Travis for pure library code. We treat these builds like any other open-source project and (hopefully) publish to sonatype on every push to master.
- Heroku CI for services we're deploying to Heroku. This lets us deploy to staging for every push to master, and promote applications straight from GH. These builds run on big machines that are billed by the minute.
- Buildkite for local services that we need to deploy to vsphere, or that need access to local resources for some reason.

PR #3, opened 2019-06-03
Add modify to SplitEpi and SplitMono

swalker2m:
Would it make sense to add `modify` to these optics?  If so, what do you recommend for testing? They don't appear to make sense as fundamental laws.

For example, to halve an `Angle`:

```scala
   Angle.signedMicroarcseconds.modify(_/2)(θ)
```

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/optics/SplitEpi.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/optics/SplitEpi.scala
@@ -3,6 +3,7 @@
 
 package gsp.math.optics
 
+import cats.Functor
 import cats.arrow.Category
 import monocle.{ Iso, Prism }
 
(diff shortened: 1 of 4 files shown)


modules/math/shared/src/main/scala/gsp/math/optics/SplitEpi.scala
```diff
@@ -22,6 +23,14 @@ import monocle.{ Iso, Prism }
  */
 final case class SplitEpi[A, B](get: A => B, reverseGet: B => A) {
 
+  /** Modify the target of the SplitEpi using a function. */
+  def modify(f: B => B): A => A =
+    a => reverseGet(f(get(a)))
+
+  /** Modify the target of the SplitEpi using Functor. */
+  def modifyF[F[_]: Functor](f: B => F[B])(a: A): F[A] =
```
tpolecat: It seems inconsistent to have these methods curried differently but that's how Monocle does it do it seems fine to me.

swalker2m: Right, I thought it was odd too but I decided it would be best to be consistent with Monocle.

swalker2m:
I looked into what Monocle did to test `modify` on `Lens` and this seems to be everything:

```scala
  test("modify") {
    x.modify(_ + 1)(Point(9, 2)) shouldEqual Point(10, 2)
  }
```

I just added a basic sanity check as well.

PR #4, opened 2019-06-07
Test artifact

swalker2m:
I'd like to be able to reuse the arbitraries that we define in `gsp-math`, but I'm not sure how.  If I add 

```
publishArtifact in Test := true
```

to the build then it produces a `gsp-math_2.12-tests.jar` to which presumably I could add a "test" dependency from another project.  Is this the right thing to do?  Unfortunately it will package the test classes themselves as well which is unnecessary but harmless I suppose.


tpolecat:
Ah right, I think the right way to do this is to pull out `gsp-math-testkit` or something like that, with the laws and arbitraries in `src/main`. Then the tests themselves will need to be in another (unpublished) module that depends on it in the test configuration. Downstream users can add `"edu.gemini" %% "gsp-math-testkit" % <version> % "test"` if they want to use the arbitraries or laws. Does that make sense?

PR #9, opened 2019-06-17
Use shared publish settings

swalker2m:
Updates the build to get publish settings from the `sbt-gsp` plugin instead of defining them locally.


swalker2m:
I didn't go as far as including the `sbt-ci-release` plugin in every build because I don't think we'll be publishing the applications?  This simply defines the required settings that are common across builds so there is only one place to update it.

PR #10, opened 2019-06-18
Use Shared JS Settings

swalker2m:
Small update to make use of the shared JS settings defined in `sbt-gsp`.


swalker2m:
I think this is fine but it would be best for @cquiroz or someone who knows scala-js to take a quick look.

PR #26, opened 2019-08-19
Cross-compile for 2.13

tpolecat:
This adds support for 2.13

- Upgrading to cats 2.x breaks compatibility for tests because `CatsSuite` is gone from cats-testkit in order to eliminate the ScalaTest dependency (we get it via monocle-law). It's still used internally in cats however, so I just copy/pasted those files. We need a better solution eventually but for now this is ok.
- `scala.collection.JavaConverters` is now `scala.jdk.CollectionConverters` but you can paper over it by using `scala-collection-compat` which makes it available for 2.12
- Scaladoc is pickier in 2.13 which is good, it caught some errors.


tpolecat:
Good catch.

.travis.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   include:
     # stage="test" if no stage is specified
     - env: TEST="build"
-      script: sbt headerCheck compile test
+      script: sbt headerCheck +compile +test
     # run ci-release only if previous stages passed
     - stage: release
       script: sbt ci-release
```
cquiroz: Will `ci-release`also "cross-release"?

tpolecat: Yes, `ci-release` does `+publishSigned`

tpolecat:
> Other libraries and apps will get the CatsSuite from here?

Hm, no. Good idea. I will move them into `gsp-math-testkit`.

tpolecat:
Ok I moved `CatsSuite` into `gsp-math-testkit` so anyone pulling that in will get it now. Need re-approval from someone.

tpolecat:
Ok, merged. I am inclined not to tag this one because it's still on RC versions of cats and monocle. Anyone have strong opinions on this?

swalker2m:
No strong opinion at all, but I'm more inclined to just tag them all since they are really only for internal use.

PR #33, opened 2019-09-05
switch to cats-testkit-scalatest

tpolecat:
This removes the `cats.test` classes we copy/pasted in favor of the brand new `cats-testkit-scalatest` which does the same thing.


build.sbt
```diff
@@ -1,11 +1,12 @@
...
-lazy val catsVersion          = "2.0.0-RC1"
-lazy val collCompatVersion    = "2.1.2"
-lazy val kindProjectorVersion = "0.10.3"
-lazy val monocleVersion       = "2.0.0-RC1"
+lazy val attoVersion                 = "0.7.0-RC1"
+lazy val catsVersion                 = "2.0.0-RC1"
+lazy val collCompatVersion           = "2.1.2"
+lazy val kindProjectorVersion        = "0.10.3"
+lazy val monocleVersion              = "2.0.0-RC1"
+lazy val catsTestkitScalaTestVersion = "1.0.0-M1"
```
cquiroz: Does this bring scalatest as well? If so should we be explicit here on what version?

tpolecat: Good question. Scalatest is getting pulled in by monocle-law … I'll investigate.

tpolecat: Ok, so the deal is, `monocle-law` and `cats-testkit-scalatest` both pull in `discipline-scalatest`, which is where scalatest itself is coming from. I don't want to push `cats-testkit-scalatest` into Monocle because it doesn't need the extra dependencies. So I think we just need to manage this like any other diamond inheritance scenario. We'll get eviction notices if there's a problem.

image

Regarding scalatest version I think we should just use whatever is pulled in here. It's hard to find versions of all this stuff that work together so I'd prefer that it be someone else's problem.

cquiroz: I agree that we should use the same version as used by the dependencies. The other day I was looking into this and couldn't find exactly what scalatest version we were using

Though it would be redundant it may be good to be explicit what version are we using, especially considering that there so many scalatest variants

tpolecat: Ok I can add it. We just need to be vigilant about eviction warnings. Maybe I can turn it into a build error.

tpolecat: Oh man it's tangled up in scaltestplus-scalacheck as well, we would basically be copy-pasting disciple-scalatest's dependencies and would have them three times. I'd prefer to just let it get pulled in transitively. We shouldn't care on our end.

cquiroz: Ok, if it is complex never mind. I desire more and more a non-scalatest solution :D

tpolecat: Yeah, I want to use whatever Cats uses because it lets us inherit all the law-checking stuff, but it's really complicated. I hope someone is able to simplify at some point.

PR #44, opened 2019-09-17
fix scala-js tests

tpolecat:
This fixes the dependencies such that the scala-js stuff gets built correctly and tests run. It's crazy that this was just failing silently … confidence high :-\


build.sbt
```diff
@@ -7,6 +7,7 @@ lazy val collCompatVersion           = "2.1.2"
 lazy val kindProjectorVersion        = "0.10.3"
 lazy val monocleVersion              = "2.0.0"
 lazy val catsTestkitScalaTestVersion = "1.0.0-M1"
+lazy val scalaJavaTimeVersion      = "2.0.0-RC2"
```
cquiroz: There is an RC3

PR #51, opened 2019-10-17
Refactor Offset in order to define its Component elements once.

rpiaggio:


modules/math/shared/src/main/scala/gsp/math/Offset.scala
```diff
@@ -49,122 +51,104 @@ object Offset extends OffsetOptics {
...
-      Q(Angle.Angle0)
-
-    /** Q forms a commutative group. */
-    implicit val CommutativeGroupQ: CommutativeGroup[Q] =
-      new CommutativeGroup[Q] {
-        val empty: Q = Zero
-        def combine(a: Q, b: Q) = a + b
-        def inverse(a: Q) = -a
-      }
+  object P {
```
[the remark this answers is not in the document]
rpiaggio: Done.

modules/math/shared/src/main/scala/gsp/math/Offset.scala
```diff
@@ -49,122 +51,104 @@ object Offset extends OffsetOptics {
...
-    /** This Q component, reflected around the 0 .. 180° axis. Exact, invertable. */
-    def unary_- : Q =
-      Q(-toAngle)
+  // P, Q types and objects defined here for source compatibility.
 
-    /** Some of this Q component and `p`. Exact. */
-    def +(p: Q): Q =
-      Q(toAngle + p.toAngle)
+  /** @deprecated("Use Component[Axis.P] instead", "gsp-math 0.1.7") */
+  type P = Component[Axis.P]
```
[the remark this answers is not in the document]
rpiaggio: Ok, I've undeprecated them. It does seem convenient to keep them.

modules/math/shared/src/main/scala/gsp/math/Offset.scala
```diff
@@ -49,122 +51,87 @@ object Offset extends OffsetOptics {
...
+    /** Component[A] forms a commutative group. */
+    implicit def CommutativeGroupComponent[A]: CommutativeGroup[Component[A]] =
+      new CommutativeGroup[Component[A]] {
+        val empty: Component[A] = Zero[A]
+        def combine(a: Component[A], b: Component[A]) = a + b
+        def inverse(a: Component[A]) = -a
       }
 
-    implicit val ShowP: Show[P] =
+    implicit def ShowComponent[A]: Show[Component[A]] =
```
cquiroz: @tpolecat Are we moving in the direction of not doing `Show`?

tpolecat: My current thinking is that it's ok to define `Show` but it should always be the same thing as `toString`, which should always be for debugging and never shown to the user. Changing a `Show` or `toString` should never affect visible behavior. Use specific methods like `formatUTC` or `formatHmsDms` if you need to get a string with specific expected behavior; and define a `Format[String, A]` if there is an accompanying parser for that format. All of this is to prevent the situation we have in OCS where you never know if it's ok to change a toString, and you never know whether parsers and toStrings are consistent.

modules/math/shared/src/main/scala/gsp/math/Offset.scala
```diff
@@ -49,122 +51,87 @@ object Offset extends OffsetOptics {
     Order.by(o => (o.p, o.q))
 
   /** P component of an angular offset.. */
```
swalker2m: It's a trivial detail but maybe we should update the comment (and others below) to not refer to `P` in particular.

modules/math/shared/src/main/scala/gsp/math/Offset.scala
```diff
@@ -49,122 +51,87 @@ object Offset extends OffsetOptics {
...
-  final case class P(toAngle: Angle) {
+  final case class Component[A](toAngle: Angle) {
 
     /** This P component, reflected around the 0 .. 180° axis. Exact, invertable. */
-    def unary_- : P =
-      P(-toAngle)
+    def unary_- : Component[A] =
+      Component[A](-toAngle)
 
     /** Some of this P component and `p`. Exact. */
```
swalker2m: Not part of the PR but "Sum" I think we meant to say.

modules/math/shared/src/main/scala/gsp/math/Offset.scala
rpiaggio: Good catch. I changed some of the comments but I totally missed a few.

swalker2m:
> Code looks fine. I guess my only question is whether we want to continue to support Offset.P and Offset.Q as a convenience since most of the time we don't want to abstract over them and the parameterization is kind of verbose. What do you think @swalker2m?

My inclination is yes.  It just seems more natural to use.  The PR looks great to me.

PR #52, opened 2019-11-12
Add difference method

cquiroz:
For the seqexec I added a method to calculate the minimum angle difference between 2 angles
This PR adds this code to gsp-math and eventually we'd call it from the seqexec

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/Angle.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/Angle.scala
@@ -110,6 +110,11 @@ sealed class Angle protected (val toMicroarcseconds: Long) {
   override final def hashCode =
     toMicroarcseconds.toInt
 
+  /**
+   * angle difference or separation between two angles in the range [0 .. π)
+   */
+  def difference(a: Angle): Angle =
+    Angle.difference(this, a)
 }
 
 object Angle extends AngleOptics {
@@ -236,6 +241,17 @@ object Angle extends AngleOptics {
       degrees.toLong         * 1000 * 1000 * 60 * 60
     )
 
+  /**
+   * Calculate the angle difference or separation between two angles.
+   * The calculation is such that you get the minimal angle in the range [0 .. π)
+   */
+  def difference(α: Angle, ϐ: Angle): Angle = {
+    import cats.implicits._ // To get order syntax
+    implicit val order: Order[Angle] = AngleOrder
+
+    val δ: Angle = (α - ϐ)
+    if (δ > Angle.Angle180) δ.mirrorBy(Angle.Angle180) else δ
+  }
 }
 
 trait AngleOptics extends OpticsHelpers { this: Angle.type =>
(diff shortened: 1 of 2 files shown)


modules/math/shared/src/main/scala/gsp/math/Angle.scala
```diff
@@ -110,6 +110,11 @@ sealed class Angle protected (val toMicroarcseconds: Long) {
   override final def hashCode =
     toMicroarcseconds.toInt
 
+  /**
+   * angle difference or separation between two angles in the range [0 .. π)
```
tpolecat: [0 .. π]

modules/math/shared/src/main/scala/gsp/math/Angle.scala
```diff
@@ -236,6 +241,17 @@ object Angle extends AngleOptics {
       degrees.toLong         * 1000 * 1000 * 60 * 60
     )
 
+  /**
+   * Calculate the angle difference or separation between two angles.
+   * The calculation is such that you get the minimal angle in the range [0 .. π)
```
tpolecat: [0 .. π]

modules/tests/shared/src/test/scala/gsp/math/AngleSpec.scala
```diff
@@ -135,4 +134,29 @@ final class AngleSpec extends CatsSuite {
...
+    }
+  }
+
+  test("Difference with reverse is 180") {
+    forAll { (a: Angle) =>
+      a.difference(a + Angle.Angle180) shouldEqual Angle.Angle180
+    }
+  }
+
+  test("Difference is in the range [0 .. π)") {
```
tpolecat: [0 .. π] … this test should fail for (0, 180)

modules/tests/shared/src/test/scala/gsp/math/AngleSpec.scala
```diff
@@ -135,4 +134,29 @@ final class AngleSpec extends CatsSuite {
...
+    implicit val order: Order[Angle] = Angle.AngleOrder
+    forAll { (a: Angle, b: Angle) =>
+      (a.difference(b) >= Angle.Angle0 && a.difference(b) < Angle.Angle180) shouldBe true
+    }
+  }
+
+  test("Difference is commutative") {
+    forAll { (a: Angle, b: Angle) =>
+      a.difference(b) shouldEqual a.difference(b)
```
tpolecat: One of these should be `b.difference(a)` ;-)

modules/tests/shared/src/test/scala/gsp/math/AngleSpec.scala
cquiroz: Are you using `[]` for an inclusive range, i.e `0 <= x <= pi`?

modules/tests/shared/src/test/scala/gsp/math/AngleSpec.scala
cquiroz: Good catch

modules/tests/shared/src/test/scala/gsp/math/AngleSpec.scala
tpolecat: Yes, `[` for inclusive, `(` for exclusive.

tpolecat:
👍

PR #70, opened 2020-02-20
downgrade sbt-ci-release

tpolecat:
sbt-ci-release 1.5.2 added a `clean` step that may be causing the 2.13 jarfiles to end up empty … i want to try a tag with this downgrade and see if it fixes the issue.


tpolecat:
oops there was no 1.5.1 …  stepping down to 1.5.0

PR #77, opened 2020-02-26
minor cleanup

swalker2m:
Trivial updates and some header copyright year changes (to avoid mixing these in with a future PR).

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
@@ -11,7 +11,7 @@ import monocle.Prism
 /**
  * Exact wavelengths represented as unsigned integral picometers in the range [0 .. Int.MaxValue]
  * which means the largest representable wavelength is 2.147483647 mm.
- * @param toPicometers This wavelength in integral picometers (10^-12 of a meter).
+ * @param toPicometers This wavelength in integral picometers (10^-12^ of a meter).
  */
 sealed abstract case class Wavelength private (toPicometers: Int) {
   // Sanity check … should be correct via the companion constructor.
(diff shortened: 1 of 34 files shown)


modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
```diff
@@ -11,7 +11,7 @@ import monocle.Prism
 /**
  * Exact wavelengths represented as unsigned integral picometers in the range [0 .. Int.MaxValue]
  * which means the largest representable wavelength is 2.147483647 mm.
- * @param toPicometers This wavelength in integral picometers (10^-12 of a meter).
+ * @param toPicometers This wavelength in integral picometers (10^-12^ of a meter).
```
tpolecat: My intent was `10^-12` ... does `^foo^` make a superscript in Scaladoc?

modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
```diff
@@ -79,25 +80,19 @@ object Wavelength {
   }
 
   /**
-   * Creates a Wavelength from an Int in Å.
-   * @param Å in the range (0 to 21474836)
-   * @return a Wavelength provided Å is in range
+   * Returns a `UnitConverter` for angstroms in the range 0 to 21474836.
    */
   def fromAngstroms: UnitConverter =
```
tpolecat: While we're here is there any reason for these to not be `val`s?

modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
swalker2m: Yes.  You could argue though that mostly the documentation is read in the source code and so this change is more confusing. I can put it back.

modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
swalker2m: Yeah that makes sense.

modules/math/shared/src/main/scala/gsp/math/Wavelength.scala
tpolecat: Ok. Either way. I tried `10⁻¹²` but it's unreadable in my terminal font.

PR #94, opened 2020-05-07
Minor updates to the build

cquiroz:
Updates the cache for actions and sbt-gsp

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -7,10 +7,22 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Set up JDK 1.8
-        uses: olafurpg/setup-scala@v2
+        uses: olafurpg/setup-scala@v7
         with:
           java-version: 1.8
+      - name: Cache dependencies
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/coursier/v1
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -12,6 +12,7 @@ lazy val scalaJavaTimeVersion        = "2.0.0"
 inThisBuild(Seq(
   homepage := Some(url("https://github.com/gemini-hlsw/gsp-math")),
   addCompilerPlugin("org.typelevel" %% "kind-projector" % kindProjectorVersion),
+  Global / onChangedBuildSource := ReloadOnSourceChanges
```
cquiroz, line 15: Will this be a good default to set on sbt-gsp?

tpolecat: Yeah, if you think it's good then let's do it. I can't think of a situation when I *don't* want to reload immediately.

.github/workflows/scala.yml
```diff
@@ -7,10 +7,22 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
       - name: Set up JDK 1.8
-        uses: olafurpg/setup-scala@v2
+        uses: olafurpg/setup-scala@v7
```
tpolecat, line 12: We need something like scala-steward to tell us when these get updated, I guess.

cquiroz: Right, I discovered the update just by chance

PR #96, opened 2020-05-07
Add geom module

swalker2m:
This is an old PR that was originally offered for discussion in the `ocs3` repo.  I'm bringing it back to life for discussion particularly because it came up during the "sprint review" meeting.  The JavaScript implementation using JSTS is missing but presumably not difficult?  The focus was on calculations for AGS, but you can convert the shapes into AWT Shapes (which we don't really need) and presumably into something that can be drawn with a JavaScript library.  I'm not completely convinced by any of this and would like a better alternative but I don't know of any short of writing all the geometry code in Scala.

--

The issue is defining shapes for use in AGS, etc.  This would include probe arms, probe ranges, and detector areas at least.  There are two main operations needed for AGS:

1) Determine whether a point is contained in a shape (guide star reachability)
2) Determine the area of a shape relative to another shape (vignetting comparisons)

```scala
trait Shape {

  /**
   * Determines whether the given position is contained in the shape.
   */
  def contains(o: Offset): Boolean

  /**
   * Area of the shape. The value is in microarcseconds angular separation^2
   * but the absolute value is less useful than comparing across multiple
   * shapes (to determine minimum vignetting).
   */
  def area: Area

}
```

Instead of using AWT though, I wanted to find a geometry library either in Scala (fail) or else with Java and JavaScript implementations (see [JTS](https://github.com/locationtech/jts)/[JSTS](https://github.com/bjornharrtell/jsts)) .  So the idea is to describe the Shape in an ADT and then interpret it down to a specific shape implementation in a language/library-specific interpreter.

```scala
sealed trait ShapeExpression {

  def eval(implicit ev: ShapeInterpreter): Shape =
    ev.interpret(this)

}

object ShapeExpression {

  final case class Ellipse(a: Offset, b: Offset)                         extends ShapeExpression
  final case class Polygon(os: List[Offset])                             extends ShapeExpression
  final case class Rectangle(a: Offset, b: Offset)                       extends ShapeExpression

  final case class Difference(a: ShapeExpression, b: ShapeExpression)    extends ShapeExpression
  final case class Intersection(a: ShapeExpression, b: ShapeExpression)  extends ShapeExpression
  final case class Union(a: ShapeExpression, b: ShapeExpression)         extends ShapeExpression

  final case class Rotate(e: ShapeExpression, a: Angle)                  extends ShapeExpression
  final case class Translate(e: ShapeExpression, o: Offset)              extends ShapeExpression

}
```

Note that the shapes are defined with `Offset` positions, which seemed like the natural way to do it.  Usually they are specified in terms of arcsec sizes and at offset position locations.  With some syntax (not shown) here's the JTS/JVM implementation of the interpreter:

```scala
object JtsShapeInterpreter extends ShapeInterpreter {

  private def toGeometry(e: ShapeExpression): Geometry =
    e match {
      // Constructors
      case Ellipse(a, b)      => a.shapeFactory(b).createEllipse
      case Polygon(os)        => Jts.geometryFactory.createPolygon(os.map(_.coordinate).toArray)
      case Rectangle(a, b)    => a.shapeFactory(b).createRectangle

      // Combinations
      case Difference(a, b)   => toGeometry(a).difference(toGeometry(b))
      case Intersection(a, b) => toGeometry(a).intersection(toGeometry(b))
      case Union(a, b)        => toGeometry(a).union(toGeometry(b))

      // Transformations
      case Rotate(e, a)       =>
        AffineTransformation
          .rotationInstance(a.toDoubleDegrees)
          .transform(toGeometry(e))

      case Translate(e, o)    =>
        val c = o.coordinate
        AffineTransformation
          .translationInstance(c.x, c.y)
          .transform(toGeometry(e))
    }

  override def interpret(e: ShapeExpression): Shape =
    JtsShape(toGeometry(e))
}
```
Hiding some syntax support, the GMOS OIWFS probe arm can be defined as a `ShapeExpression`:
```scala
  val PickoffArmLength: Angle      = 358460.mas
  val PickoffMirrorSize: Angle     =  20000.mas
  val ProbeArmLength: Angle        = PickoffArmLength - (PickoffMirrorSize / 2)
  val ProbeArmTaperedWidth: Angle  =  15000.mas
  val ProbeArmTaperedLength: Angle = 180000.mas

  private val arm: ShapeExpression = {
    val hm  = PickoffMirrorSize    / 2
    val htw = ProbeArmTaperedWidth / 2

    val p0 = (-hm,                           htw  )
    val p1 = (p0._1 - ProbeArmTaperedLength, hm   )
    val p2 = (p0._1 - ProbeArmLength,        p1._2)
    val p3 = (p2._1,                        -hm   )
    val p4 = (p1._1,                         p3._2)
    val p5 = (p0._1,                        -htw  )

    ShapeExpression.polygon(p0, p1, p2, p3, p4, p5, p0)
  }

  private val pickoff: ShapeExpression = {
    val s = PickoffMirrorSize / 2
    ShapeExpression.rectangle((s, s), (s - PickoffMirrorSize, s - PickoffMirrorSize))
  }

  /**
   * Description of the GMOS OIWFS probe arm with the pickoff mirror centered
   * at the base position.
   */
  val shape: ShapeExpression =
    arm ∪ pickoff
```

Note `arm ∪ pickoff` to combine the probe arm proper with the pickoff mirror.  Here's an AWT plot obtained from the resulting shape with the pickoff mirror centered at the origin (marked with a +).

screen shot 2019-01-15 at 16 17 02

where the `toAwt` below is only defined in the `JtsShape` implementation and only needed to visualize the shape for the demo:

```scala
    import gem.geom.jts.interpreter._

    override def paint(g: Graphics): Unit = {
      val g2d = g.asInstanceOf[Graphics2D]

      GmosOiwfsProbeArm.shape.eval match {
        case jts: JtsShape => g2d.draw(jts.toAwt)
        case x             => sys.error(s"Whoa unexpected shape type: $x")
      }
  }
```


cquiroz:
Should this go into a separate `geom` module (not project)?

cquiroz:
Does it make sense to try building the js version right away?

swalker2m:
> Should this go into a separate geom module (not project)?

I waffled for a while on how (and where) it should be set up, and that commit title is admittedly misleading.  I'm open for suggestions on the best way to do this.

swalker2m:
> Does it make sense to try building the js version right away?

I think it's worth investigating yes and it probably makes sense to progress a bit on that before merging anything.

modules/math/shared/src/main/scala/gsp/math/geom/syntax/ShapeExpression.scala
```diff
@@ -0,0 +1,166 @@
...
+  def union(that: ShapeExpression): ShapeExpression =
+    Union(self, that)
+
+  def -(that: ShapeExpression): ShapeExpression =
+    difference(that)
+
+  def ∩(that: ShapeExpression): ShapeExpression =
+    intersection(that)
+
+  def ⟲(a: Angle): ShapeExpression =
```
tpolecat, line 35: Haha nice

PR #104, opened 2020-06-02
Support scala.js on the geom module using gpp-jts

cquiroz:
This PR makes the geom package work on scala.js using the port of JTS to scala on gpp-jts
Some notes:
* Very little code is jvm/js specific and only the pieces related to rendering to awt and web
* There isn't yet any implementation to render on a web context, that would be dealt on a later PR
* gpp-jts could be buggy due to the way it was constructed. Proceed with caution


build.sbt
```diff
@@ -28,18 +27,20 @@ lazy val math = crossProject(JVMPlatform, JSPlatform)
...
     libraryDependencies ++= Seq(
-      "org.locationtech.jts" % "jts-core" % jtsVersion
+      "edu.gemini"                 %%% "gpp-jts-awt"             % jtsVersion
     )
   )
   .jsSettings(gspScalaJsSettings: _*)
   .jsSettings(
-    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % scalaJavaTimeVersion
+    libraryDependencies += "io.github.cquiroz" %%% "scala-java-time" % scalaJavaTimeVersion,
+    scalaJSUseMainModuleInitializer := true,
```
swalker2m, line 43: What do we do with this setting?

modules/math/js/src/main/scala/gsp/math/geom/jts/JtsShape.scala
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * JS / JTS implementation of Shape.
+ */
+final case class JtsShape(g: Geometry) extends Shape {
+
+  override def contains(o: Offset): Boolean =
+    g.contains(o.point)
+
+  override def area: Area =
+    Area.fromMicroarcsecondsSquared.getOption(g.getArea.round).getOrElse(Area.MinArea)
```
swalker2m, line 20: Do we need JavaScript and JVM-specific implementations of `Shape` now?  The translation to AWT needs to be separated of course but the `Shape` implementation itself doesn't differ?

build.sbt
cquiroz: We probably don't need it. It lets us build a runnable example on js, equivalent to JTSDemo but if we move the "demo" code to another module we can remove it

modules/math/js/src/main/scala/gsp/math/geom/jts/JtsShape.scala
cquiroz: I agree, we should preserve the common bits and put the AWT/SVG parts onto the jvm/js specific parts

PR #107, opened 2020-06-04
Added GMOS science area geometry + bug fixes

swalker2m:
Adds a `GmosScienceAreaGeometry`, fixes a couple of bugs, and adds a new `ShapeExpression.Empty`.  This is an interim PR.  Ultimately the GMOS stuff (probe arm, science area and probe range) should be moved into, presumably, `gsp-core`.

### DIFF
--- a/modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/GmosScienceAreaGeometry.scala
+++ b/modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/GmosScienceAreaGeometry.scala
@@ -23,14 +23,19 @@ object GmosScienceAreaGeometry {
   private def offsetInP(p: Angle): Offset =
     Offset(Offset.P(p), Offset.Q.Zero)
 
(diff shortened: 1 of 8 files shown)


modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/GmosScienceAreaGeometry.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import gsp.math.syntax.int._
+
+/**
+ * GMOS science area geometry.
+ */
+object GmosScienceAreaGeometry {
+
+  val imaging: ShapeExpression =
+    imagingFov(330340.mas, 33840.mas)
```
cquiroz, line 18: Where do these numbers come from? Are they sepecified in some documento or come from the old codebase?

swalker2m: Unfortunately I don't have a great answer.  I took them from the old codebase and I'm not sure where to find GMOS documentation for this.

PR #113, opened 2020-06-16
Support rendering of geometries to svg

cquiroz:
This is a first pass as exporting geometries to svg so we can use them on the web side.
A js library `svg.js` has been used an an API to interpret `ShapeExpressions` into svg. So far it seems to work though there are some rough edges around scaling and positioning.
The svgs produced are intentionally non-styled so we can apply styles with css.
This is an example rendered onto a web page

image


cquiroz:
I'll close this as I changed the build quite a bit

PR #120, opened 2020-07-17
Add SkyCalc.

rpiaggio:
`ImprovedSkyCalc(Methods)` was migrated from `ocs` and API was adapted to use `Angle` and `Coordinates` from `gsp-math` as well as `java.time` classes.

The logic was otherwise untouched, since conversion from Java to Scala was done automatically using IntelliJ.

For testing purposes, results are compared to the `ocs` output. For this reason, testing is only possible in the JVM (although the code runs in JS). A version of `ImprovedSkyCalc` was created in testing to allow easily calling the logic of `ocs`'s `ImprovedSkyCalcMethods` but with the new API.

I implemented the new test with Weaver. It looks like a promising testing framework to adopt due to it parallelizing everything and its integration with `cats-effect`. But we can evaluate other options or go back to ScalaTest.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -73,10 +73,10 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform)
   .jvmSettings(
     resolvers += "Gemini Repository" at "https://github.com/gemini-hlsw/maven-repo/raw/master/releases",
     libraryDependencies ++= Seq(
-      "edu.gemini.ocs" %%% "edu-gemini-util-skycalc" % "2020001.1.7",
-      "com.disneystreaming" %%% "weaver-framework" % "0.4.1",
-      "com.disneystreaming" %%% "weaver-scalacheck" % "0.4.1",
-      "com.47deg" %%% "scalacheck-toolbox-datetime" % "0.3.5"
+      "edu.gemini.ocs" %% "edu-gemini-util-skycalc" % "2020001.1.7",
+      "com.disneystreaming" %% "weaver-framework" % "0.4.1",
+      "com.disneystreaming" %% "weaver-scalacheck" % "0.4.1",
+      "com.47deg" %% "scalacheck-toolbox-datetime" % "0.3.5"
     ),
     testFrameworks += new TestFramework("weaver.framework.TestFramework")
   )
--- a/build.sbt
+++ b/build.sbt
@@ -67,18 +67,21 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform)
   .dependsOn(math, testkit)
   .settings(
     name := "gsp-math-tests",
-    skip in publish := true
+    skip in publish := true,
+    libraryDependencies ++= Seq(
+      "com.disneystreaming" %%% "weaver-framework" % "0.4.1"
+    ),
+    testFrameworks += new TestFramework("weaver.framework.TestFramework"),
+    scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.CommonJSModule))
   )
   .jvmConfigure(_.enablePlugins(AutomateHeaderPlugin))
   .jvmSettings(
     resolvers += "Gemini Repository" at "https://github.com/gemini-hlsw/maven-repo/raw/master/releases",
     libraryDependencies ++= Seq(
       "edu.gemini.ocs" %% "edu-gemini-util-skycalc" % "2020001.1.7",
-      "com.disneystreaming" %% "weaver-framework" % "0.4.1",
       "com.disneystreaming" %% "weaver-scalacheck" % "0.4.1",
       "com.47deg" %% "scalacheck-toolbox-datetime" % "0.3.5"
-    ),
-    testFrameworks += new TestFramework("weaver.framework.TestFramework")
+    )
   )
   .jsSettings(gspScalaJsSettings: _*)
 
--- a/build.sbt
+++ b/build.sbt
@@ -9,12 +9,15 @@ lazy val catsTestkitScalaTestVersion = "1.0.1"
 lazy val scalaJavaTimeVersion        = "2.0.0"
 lazy val jtsVersion                  = "0.0.9"
 lazy val svgdotjsVersion             = "0.0.1"
+lazy val newType                     = "0.4.4"
 
-inThisBuild(Seq(
-  homepage := Some(url("https://github.com/gemini-hlsw/gsp-math")),
-  addCompilerPlugin("org.typelevel" %% "kind-projector" % kindProjectorVersion),
-  Global / onChangedBuildSource := ReloadOnSourceChanges
-) ++ gspPublishSettings)
+inThisBuild(
+  Seq(
+    homepage := Some(url("https://github.com/gemini-hlsw/gsp-math")),
+    addCompilerPlugin("org.typelevel" %% "kind-projector" % kindProjectorVersion),
+    Global / onChangedBuildSource := ReloadOnSourceChanges
+  ) ++ gspPublishSettings
+)
 
 skip in publish := true
 
@@ -76,12 +79,13 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform)
   )
   .jvmConfigure(_.enablePlugins(AutomateHeaderPlugin))
   .jvmSettings(
-    resolvers += "Gemini Repository" at "https://github.com/gemini-hlsw/maven-repo/raw/master/releases",
+    resolvers += "Gemini Repository".at(
+      "https://github.com/gemini-hlsw/maven-repo/raw/master/releases"
+    ),
     libraryDependencies ++= Seq(
-      "edu.gemini.ocs" %% "edu-gemini-util-skycalc" % "2020001.1.7",
-      "com.disneystreaming" %% "weaver-scalacheck" % "0.4.1",
-      "com.47deg" %% "scalacheck-toolbox-datetime" % "0.3.5"
+      "edu.gemini.ocs"      %% "edu-gemini-util-skycalc"     % "2020001.1.7",
+      "com.disneystreaming" %% "weaver-scalacheck"           % "0.4.1",
+      "com.47deg"           %% "scalacheck-toolbox-datetime" % "0.3.5"
     )
   )
   .jsSettings(gspScalaJsSettings: _*)
-
--- a/build.sbt
+++ b/build.sbt
@@ -70,5 +70,15 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform)
     skip in publish := true
   )
   .jvmConfigure(_.enablePlugins(AutomateHeaderPlugin))
+  .jvmSettings(
+    resolvers += "Gemini Repository" at "https://github.com/gemini-hlsw/maven-repo/raw/master/releases",
+    libraryDependencies ++= Seq(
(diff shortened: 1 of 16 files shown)


build.sbt
```diff
@@ -70,5 +70,15 @@ lazy val tests = crossProject(JVMPlatform, JSPlatform)
     skip in publish := true
   )
   .jvmConfigure(_.enablePlugins(AutomateHeaderPlugin))
+  .jvmSettings(
+    resolvers += "Gemini Repository" at "https://github.com/gemini-hlsw/maven-repo/raw/master/releases",
+    libraryDependencies ++= Seq(
+      "edu.gemini.ocs" %%% "edu-gemini-util-skycalc" % "2020001.1.7",
```
cquiroz, line 76: I suppose we don't need three `%` for the jvm side

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
```diff
@@ -0,0 +1,224 @@
...
+  private var lunarPhaseAngle                      = .0
+  private var sunAltitude                          = .0
+  private var lunarElevation                       = .0
+
+  // caching for calculate()
+  private var cachedCoordinates: Option[Coordinates] = none
+  private var cachedInstant: Option[Instant]         = none
+  private var cachedCalculateMoon: Option[Boolean]   = none
+
+  private implicit val eqInstant: Eq[Instant] = Eq.by(i => (i.getEpochSecond, i.getNano))
```
cquiroz, line 45: I wonder if this is defined in some of the other libs. On the other hand it is private, we probably don't need to worry

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  private def getCircumstances(
+    objra:         Double,
+    objdec:        Double,
+    objepoch:      Double,
+    curep:         Double,
+    sid:           Double,
+    lat:           Double,
+    jdut:          DoubleRef,
```
cquiroz, line 87: What is `DoubleRef`?

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalcMethods.scala
```diff
@@ -0,0 +1,1413 @@
...
+trait ImprovedSkyCalcMethods {
+
+  // defined quantities for apparent place transforms ..
+  protected val XFORM_FROMSTD      = 1
+  protected val XFORM_TOSTDEP: Int = -1
+  protected val XFORM_JUSTPRE      = 1
+  protected val XFORM_DOAPPAR      = 0
+
+  // some (not all) physical, mathematical, and astronomical constants used are defined here.
+  protected val TWOPI     = 6.28318530717959
```
cquiroz, line 22: Maybe we should eventually migrate these constants to other places in the lib

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalcMethods.scala
```diff
@@ -0,0 +1,1413 @@
...
+
+  final protected class DoubleRef(var d: Double) {
+    def this() = {
+      this(0.0)
+    }
+
+    override def toString: String = d.toString
+  }
+
+  protected case class DateTime(
```
cquiroz, line 65: This is a nitpick but there is another `DateTime` in `java.unit` maybe we could avoid a name clash

modules/tests/jvm/src/test/scala/gsp/math/skycalc/ImprovedSkyCalcSpec.scala
```diff
@@ -0,0 +1,61 @@
...
+  implicit val showZDT: Show[ZonedDateTime] = Show.fromToString
+
+  private val zdtFrom  = ZonedDateTime.of(
+    LocalDate.of(1901, 1, 1),
+    LocalTime.MIDNIGHT,
+    ZoneOffset.UTC
+  )
+  private val zdtRange = Duration.ofDays(Period.ofYears(1000).getDays.toLong)
+
+  simpleTest("ImprovedSkyCalcSpec: Arbitrary sky calculations") {
```
cquiroz, line 33: Neat

build.sbt
rpiaggio: Good catch. Fixed.

rpiaggio:
> Great work, I'm glad you could make this
> One quick extra thing I'd like is to have a "test" on the js side. It doesn't need to compare with the java side but just a calculation against a known number or even just Test runnable app. The reason for that is to ensure there are no linking errors when building the js app

Ok, we can generate some known results.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
rpiaggio: I considered using https://github.com/ChristopherDavenport/cats-time. I don't know if it's worth adding a dependency just for this.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
rpiaggio: It's a wrapper for a mutable double. It's used in a number of places for output values. It's a far cry from FP but I don't know if we want to rewrite the whole thing in FP style.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalcMethods.scala
rpiaggio: Hmm.... Let me see if we can get to actually use `java.time.LocalDateTime` and discard this class...

rpiaggio: I actually replaced its use for `Instant`, which seems like the right thing to do.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
cquiroz: I found it below in the code, I thought it was something external

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalcMethods.scala
cquiroz: 👍

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
tpolecat: I think this is fine as-is, but you might check to see what `.equals` does for `Instant` … we may be able to define this as `Eq.fromUniversalEquals`.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
```diff
@@ -0,0 +1,222 @@
...
+  private var cachedInstant: Option[Instant]         = none
+  private var cachedCalculateMoon: Option[Boolean]   = none
+
+  private implicit val eqInstant: Eq[Instant] = Eq.by(i => (i.getEpochSecond, i.getNano))
+
+  def calculate(
+    coords:        Coordinates,
+    instant:       Instant,
+    calculateMoon: Boolean
+  ): Unit = { // Early exit if the parameters haven't changed.
```
tpolecat, line 51: Eventually it would be nice to return a struct here rather than side-effecting, and add an LRU cache. But for now this is ok I think.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
rpiaggio: Ok. `Instant.equals` does the same thing. Changing it.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
rpiaggio: That sounds good and it should be straightforward. I'll try to take it a notch further and encapsulate init/calc/get in one static function and expose only that.

BTW, I think it would be nice to have a `Location` class that encapsulates (latitude, longitude, altitude) and use that in the input. In that case we could make `Site` in `gsp-core` extend this class. I accept suggestions for a more accurate name other than `Location`.

rpiaggio: About the cache: I'm not sure it would be useful unless we specify a really big cache. For example, for a night, we compute a point for every minute, which is in the order of 600 points. Caching would be useful if the user switches back and forth between nights/targets/sites, so it would make sense if we store data for a bunch of charts. Do we want, say, a 6000 entry cache?

Also, maybe in that case we want to cache at a higher level, having one cache entry store the entire plot data (QPT does that).

I'm only thinking in terms of plots here, I don't know yet if this code is used in other scenarios.

WDYT?

cquiroz: Do we really need a cache? If we were I'd make it external to the code here so we can use it only if needed

tpolecat: That's also fine, but we do want a way to do it because these calculations are expensive.

rpiaggio: I removed the cache for now, mainly because it didn't seem thread safe. Let's implement it on callers as needed.

modules/tests/shared/src/test/scala/gsp/math/skycalc/ImprovedSkyCalcSpec.scala
```diff
@@ -0,0 +1,42 @@
...
+import cats.Monoid
+import cats.Show
+import java.time._
+import gsp.math.Coordinates
+import gsp.math.Angle
+
+// This is just a basic case, mostly to test linking in JS.
+// Property based testing is in ImprovedSkyCalcSpecJVM, where output
+// is compared to the one from {edu.gemini.skycalc} in Java.
+object ImprovedSkyCalcSpec extends SimpleIOSuite {
```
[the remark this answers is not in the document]
rpiaggio: Yes, in CI we run `+test` so it runs it in JVM and JS.

modules/math/shared/src/main/scala/gsp/math/skycalc/ImprovedSkyCalc.scala
rpiaggio: I'd appreciate another look at this. I added a case class to return the results with typed accessors, but I'm not 100% sure I got them right. I had to keep the raw double values there in order to compare them with the output from the OCS version, but they are only visible from the package.

PR #121, opened 2020-07-22
Migrate twilight calculations from OCS.

rpiaggio:
* Methods from `ImprovedSkyCalcMethods` that were used to compute twilights are now in `SunCalc` trait. These used `JulianDate`, which was substituted for the new one.
* Added methods to `JulianDate` to convert to `Instant` and to construct (approximately) from a `Double`.
* Computations previously performed in `TwilightBoundedNight` are now in `TwilightCalc`. The idea is to keep the math in `gsp-math`. `gsp-core` will contain a `TwilightBoundedNight` which is trivial to implement (already done, will send PR to `gsp-core` after we publish this).
* Twilight computations return an `Option` in case there's no sunset or sunrise for the specified parameters.
* JVM-only tests were added comparing output to OCS. JVM and JS tests were added computing some known results.
* Some formatting.

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/skycalc/TwilightCalc.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/skycalc/TwilightCalc.scala
@@ -1,0 +1,87 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gsp.math.skycalc
+
+import cats.implicits._
+import gsp.math.JulianDate
+import gsp.math.Place
+import java.time.Instant
+import java.time.LocalDate
+
+trait TwilightCalc extends SunCalc {
+
+  /** Compute start and end of night for a particular date and place.
+    *
+    * The night will be bounded by twilight as defined by
+    * [[gsp.math.skycalc.TwilightBoundType]]. It will be the night that starts
+    * on the given date and ends on the following day.
+    *
+    * @param boundType twilight bound type to use
+    * @param date date when the night starts
+    * @param place place on Earth
+    *
+    * @return A tuple of [[java.time.Instant]]s containing sunset and sunrise
+    * respectively, or None if there's no sunset or sunrise for the
+    * provided parameters.
+    */
+  def calculate(
+    boundType: TwilightBoundType,
+    date:      LocalDate,
+    place:     Place
+  ): Option[(Instant, Instant)] = {
+    val nextMidnight = date.atStartOfDay(place.zoneId).plusDays(1)
+    val jdmid        = JulianDate.ofInstant(nextMidnight.toInstant)
+
+    // Sunrise/set take altitude into account whereas the twilights don't.
+    //
+    // "The various twilights (6,12,18) describe how bright the sky is, and this does not depend on the altitude of
+    // the observer, however, the time of sunrise and sunset does.  For example, consider Hilo and Maunakea.  The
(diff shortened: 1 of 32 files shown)


modules/math/shared/src/main/scala/gsp/math/skycalc/TwilightCalc.scala
```diff
@@ -0,0 +1,87 @@
...
+
+import cats.implicits._
+import gsp.math.JulianDate
+import gsp.math.Place
+import java.time.Instant
+import java.time.LocalDate
+
+trait TwilightCalc extends SunCalc {
+
+  /** Compute start and end of night for a particular date and place.
```
cquiroz, line 14: Will it be in local time right?

modules/tests/shared/src/test/scala/gsp/math/skycalc/TwilightBoundTypeSpec.scala
```diff
@@ -0,0 +1,26 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gsp.math
+
+import cats.tests.CatsSuite
+import cats.{ Eq, Show }
+import gsp.math.arb._
+import gsp.math.skycalc.TwilightBoundType
+
+@SuppressWarnings(Array("org.wartremover.warts.ToString", "org.wartremover.warts.Equals"))
```
cquiroz, line 11: We don't have WartRemover anymore

modules/tests/shared/src/test/scala/gsp/math/skycalc/package.scala
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gsp.math
+
+import java.time.ZoneId
+
+package object skycalc {
+  val GN =
```
cquiroz, line 9: It seems this would be generally useful, are they going to be in `gsp-core`?

modules/math/shared/src/main/scala/gsp/math/skycalc/TwilightCalc.scala
rpiaggio: It takes a `LocalDate` as input and returns `Instant`s.

modules/tests/shared/src/test/scala/gsp/math/skycalc/TwilightBoundTypeSpec.scala
rpiaggio: Ah, thanks for the heads up.

modules/tests/shared/src/test/scala/gsp/math/skycalc/package.scala
rpiaggio: `gsp-core` already has `Site`, and instances for `GN` and `GS`. But we don't have access to those from here. These here are only used for testing against known results

cquiroz: 👍

modules/tests/shared/src/test/scala/gsp/math/skycalc/TwilightBoundTypeSpec.scala
cquiroz: We should remove it everywhere

rpiaggio: Ok, done

PR #126, opened 2020-08-03
Make toMicrosexigesimal public

cquiroz:
I'd like to use this method in explore. I'm not sure why is it protected


tpolecat:
Curious what you need it for. It's private because it just returns a tuple with no indication of what the elements mean. The wrapper types `Angle.DMS` and `HourAngle.HMS` pack them into structs with named fields.

cquiroz:
I want to do a custom formatting for the ui and don't necessarily want to keep changing n modules to do then

tpolecat:
Ok, can you add a comment explaining what the fields mean? I.e., that they mean HMS for `HourAngle` and DMS for `Angle`?

cquiroz:
Found a workaround

PR #127, opened 2020-08-07
Use refined and coulomb to define Wavelength

cquiroz:
This PR updates `Wavelength` to use `coulomb` for units and experimentally use `refined` for the values.
`Wavelength` is now defined as a quantity in picometers. It only makes sense if the value is greater than zero thus we use `refined` to define the numeric type wavelength can use.

This PR took me a while to realize some subtle details. For example to convert exactly to nm or angstrom, you need a different numeric type, in this case a `Rational`.

`UnitConversion` was removed as we can convert directly using `coulomb`

Using `refined` adds a bit of ceremony in some cases

### DIFF
--- a/modules/tests/shared/src/test/scala/gsp/math/WavelengthSpec.scala
+++ b/modules/tests/shared/src/test/scala/gsp/math/WavelengthSpec.scala
@@ -7,22 +7,21 @@ import cats.tests.CatsSuite
 import cats.{ Eq, Order, Show }
 import cats.kernel.laws.discipline._
 import gsp.math.arb._
+import gsp.math.units._
 import monocle.law.discipline._
-import org.scalacheck.Arbitrary._
-import org.scalacheck.Gen
+import coulomb.refined._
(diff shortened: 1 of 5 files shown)


modules/tests/shared/src/test/scala/gsp/math/WavelengthSpec.scala
```diff
@@ -43,64 +42,30 @@ final class WavelengthSpec extends CatsSuite {
...
+    Wavelength.fromAngstrom(Wavelength.MaxAngstrom + 1) shouldEqual none
   }
 
-  private def pow10(exp: Int): Int =
-    BigInt(10).pow(exp).toInt
-
-  private def conversionTo(u: Wavelength.UnitConverter): org.scalatest.Assertion =
-    forAll { (a: Wavelength) =>
-      u.reverseGet(a) shouldEqual a.toPicometers / pow10(u.exp)
+  test("toAmstrong") {
```
tpolecat, line 54: typo

PR #131, opened 2020-08-18
Make RadialVelocity use coulomb

cquiroz:
Convert Radial Velocity

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/RadialVelocity.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/RadialVelocity.scala
@@ -8,42 +8,48 @@ import cats.implicits._
 import gsp.math.PhysicalConstants.SpeedOfLight
 import gsp.math.optics.Format
 import gsp.math.syntax.prism._
-import scala.math.sqrt
+import gsp.math.units._
 import monocle.Prism
+import coulomb._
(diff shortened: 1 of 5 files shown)


modules/math/shared/src/main/scala/gsp/math/RadialVelocity.scala
```diff
@@ -8,42 +8,48 @@ import cats.implicits._
...
   // Sanity check
-  assert(toMetersPerSecond.abs <= SpeedOfLight, "Radial velocity exceeds the speed of light.")
+  assert(toMetersPerSecond.value.abs <= SpeedOfLight, "Radial velocity exceeds the speed of light.")
 
-  def toDoubleKilometersPerSecond: Double =
-    toMetersPerSecond.toDouble / 1000.0
+  def toKilometersPerSecond: Quantity[Rational, KilometersPerSecond] =
+    toMetersPerSecond.to[Rational, KilometersPerSecond]
 
-  def toRedshift: Double = {
```
cquiroz, line 28: We have a `Redshift` on `gsp-core` that's why I'm removing these

cquiroz:
Hmm my last comment made me look again, we have `RadialVelocity` in two places, in `gsp-math` and `gsp-core`. Rather than this update we should consider moving

PR #132, opened 2020-08-18
Move classes from gsp-core to gsp-math

cquiroz:
`RadialVelocity` from gsp-core really belongs in `gsp-math` as `ProperMotion` uses it. The classes were copied more or less exactly but some methods were added to conform to the `gsp-math` style

these will be removed from `gsp-core` once this is released

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/RadialVelocity.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/RadialVelocity.scala
@@ -3,70 +3,85 @@
 
 package gsp.math
 
-import cats.{ Order, Show }
-import cats.implicits._
-import gsp.math.PhysicalConstants.SpeedOfLight
-import gsp.math.optics.Format
-import gsp.math.syntax.prism._
-import scala.math.sqrt
+import java.math.MathContext
+
+import cats._
+import coulomb._
+import coulomb.cats.implicits._
+import coulomb.si._
+import coulomb.siprefix._
+import gsp.math.PhysicalConstants._
+import gsp.math.units._
 import monocle.Prism
+import spire.std.bigDecimal._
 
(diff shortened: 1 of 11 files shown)


modules/math/shared/src/main/scala/gsp/math/Redshift.scala
```diff
@@ -0,0 +1,54 @@
...
+    s"Redshift($z)"
+}
+
+object Redshift {
+
+  /**
+    * The `No redshift`
+    * @group Constructors
+    */
+  val Zero: Redshift = new Redshift(0)
```
swalker2m, line 44: I suppose we don't need the `new`.

modules/math/shared/src/main/scala/gsp/math/RadialVelocity.scala
```diff
@@ -3,70 +3,85 @@
...
+    Prism[BigDecimal, RadialVelocity](b =>
+      Some(b).filter(_.abs <= CValue).flatMap(v => RadialVelocity(v.withUnit[MetersPerSecond]))
+    )(_.rv.value)
+
+  /**
+    * Construct a RadialVelocity if the value is in the allowed range
+    * @group Constructors
+    */
+  def apply(rv: Quantity[BigDecimal, MetersPerSecond]): Option[RadialVelocity] =
+    if (rv.value.abs < CValue) Some(new RadialVelocity(rv)) else None
```
swalker2m, line 64: I think we can drop the `new`, fwiw, since we no longer have `sealed abstract`.

cquiroz: In this case we need `new` or it will try to call itself as `apply`

PR #133, opened 2020-08-19
Add MagnitudeValue constructor from Double

cquiroz:
This PR adds a constructor of `MagnitudeValue` out of `Double`


cquiroz:
Yes we could add more constructors. Should they be optics if possible? Unfortunately scaling tends to mess with optic laws.

swalker2m:
> Yes we could add more constructors. Should they be optics if possible? Unfortunately scaling tends to mess with optic laws.

We can always just add them as we need them.  That's probably best so we don't have a bunch of stuff we don't end up needing.  But maybe an unscaled `Int` constructor so you can make, say, a `MagnitudeValue(17)` that means mag 17.  A `Format[BigDecimal, MagnitudeValue]` might make sense too, I don't know.

PR #134, opened 2020-08-20
Migration of constraint solver classes.

rpiaggio:
Migrated classes:
* `Interval`
* `Schedule` (ex `Solution`)
* `Calculator` and subtypes.
* `Solver` and subtypes.
* `Constraint` and subtypes.
* `TargetCalculator` (no `MoonCalculator` or `SunCalculator` for the moment).

Property testing was added for `Interval` and `Schedule`, as well as optics and typeclasses. Other tests were migrated or added.

The PR introduces the use of `cats-time`~~, but we have to wait for its release with Scala.js 1.x. Therefore, I'm opening the PR as draft for the moment.~~


modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Schedule.scala
```diff
@@ -0,0 +1,327 @@
...
+  def diff(interval: Interval): Schedule = diff(Schedule.single(interval))
+
+  /** Compute a Schedule covering all instants in the same timespan but not covered by this Schedule. */
+  def gaps: Schedule =
+    if (intervals.size < 2) Schedule.Never
+    else
+      Schedule.unsafe(
+        intervals
+          .sliding(2)
+          .map({
```
cquiroz, line 102: Double parentheses?

modules/tests/shared/src/test/scala/gsp/math/skycalc/solver/ConstraintSpec.scala
```diff
@@ -0,0 +1,95 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gsp.math.skycalc.solver
+
+import cats.tests.CatsSuite
+import gsp.math.skycalc.SkyCalcResults
+import java.time.Instant
+import java.time.Duration
+import gsp.math.Declination
+import implicits._
```
cquiroz, line 11: It is hard To tell to which package this belongs to

modules/tests/jvm/src/test/scala/gsp/math/skycalc/ImprovedSkyCalcSpecJVM.scala
```diff
@@ -8,8 +8,6 @@ import weaver._
 import weaver.scalacheck._
```
cquiroz, line 8: Is there a js version of this test?

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Schedule.scala
rpiaggio: Good catch

modules/tests/shared/src/test/scala/gsp/math/skycalc/solver/ConstraintSpec.scala
rpiaggio: Agreed. Let's make the implicits import explicit.

modules/tests/jvm/src/test/scala/gsp/math/skycalc/ImprovedSkyCalcSpecJVM.scala
rpiaggio: Not this test, since it compares the output to the one from the original OCS version. There is another test added by your suggestion that cross compiles and just checks a couple of known values.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/implicits.scala
```diff
@@ -0,0 +1,82 @@
...
+import gsp.math.HourAngle
+import io.chrisdavenport.cats.time._
+
+object implicits {
+
+  implicit def exactGetter[A]: CalcGetter[GetterStrategy.Exact, A] =
+    new CalcGetter[GetterStrategy.Exact, A] {
+      def get[T](calc: Calculator[GetterStrategy.Exact, T])(field: T => A)(instant: Instant): A =
+        field(calc.timedResults(calc.toIndex(instant))._2)
+    }
```
swalker2m, line 19: If a value isn't defined exactly at `instant`, this will return a value at the previous instant for which a value exists?  If so, should it instead be the closest instant?

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/implicits.scala
```diff
@@ -0,0 +1,82 @@
...
+  implicit val interpolatedHourAngleGetter
+    : CalcGetter[GetterStrategy.LinearInterpolating, HourAngle] =
+    new CalcGetter[GetterStrategy.LinearInterpolating, HourAngle] {
+      def get[T](
+        calc:  Calculator[GetterStrategy.LinearInterpolating, T]
+      )(field: T => HourAngle)(instant: Instant): HourAngle =
+        HourAngle.fromDoubleDegrees(
+          interpolatedDoubleGetter.get(calc)(field.andThen(_.toDoubleDegrees))(instant)
+        )
+    }
```
swalker2m, line 69: I wonder if you could generalize this adding something like

```scala
def imap[B](f: A => B)(g: B => A): CalcGetter[G, B] =
  new CalcGetter[G, B] {
    def get[T](
      calc: Calculator[G, T]
    )(field: T => B)(instant: Instant): B =
      f(this.get(calc)(field.andThen(g))(instant))
  }
```
to `CalcGetter`.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/CalcGetter.scala
```diff
@@ -0,0 +1,32 @@
...
+  * @tparam A type of the value to get
+  */
+trait CalcGetter[G, A] {
+
+  /** Get value from [[Calculator]]
+    *
+    * @tparam T the results that the [[Calculator]] holds
+    * @param calc [[Calculator]] holding the timed results of type [[T]]
+    * @param field how to extract a partial result of type [[A]] from a [[T]]
+    * @param instant the desired [[java.time.Insant]] at which to get the value of type [[A]]
```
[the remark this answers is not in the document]
rpiaggio: Thank you

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/CalcGetter.scala
```diff
@@ -0,0 +1,32 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package gsp.math.skycalc.solver
+
+import java.time.Instant
+
+object GetterStrategy {
+  sealed trait Exact
+  sealed trait LinearInterpolating
+}
```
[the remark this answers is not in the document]
rpiaggio: These are phantom types. I like to think of them as `case object`s at the type level. In this case, they are used to specify which `CalcGetter` implicit to invoke.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/CalcGetter.scala
```diff
@@ -0,0 +1,32 @@
...
+
+/** Typeclass defininig how to get a value from a [[Calculator]] using a certain [[GetterStrategy]].
+  *
+  * @tparam G [[GetterStrategy]] to use
+  * @tparam A type of the value to get
+  */
```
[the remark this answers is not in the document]
rpiaggio: Ok, I wasn't aware of this. I will change throughout.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Schedule.scala
```diff
@@ -0,0 +1,325 @@
...
+  def diff(interval: Interval): Schedule = diff(Schedule.single(interval))
+
+  /** Compute a Schedule covering all instants in the same timespan but not covered by this Schedule. */
+  def gaps: Schedule =
+    if (intervals.size < 2) Schedule.Never
+    else
+      Schedule.unsafe(
+        intervals
+          .sliding(2)
+          .map { case List(i, j) => Interval.unsafe(i.end, j.start) }
```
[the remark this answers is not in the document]
rpiaggio: Nice trick to get `sliding(2)` as tuples! I'll adopt it. I think I'm using `sliding(2)` in some other place too.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Constraint.scala
```diff
@@ -0,0 +1,105 @@
...
+import cats.implicits._
+import java.time.Duration
+import java.time.Instant
+import gsp.math.Angle
+import gsp.math.Declination
+import gsp.math.HourAngle
+import gsp.math.skycalc.SkyCalcResults
+
+trait Constraint[T, A] {
+  protected val solver: Solver[A]
```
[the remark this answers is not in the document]
rpiaggio: I think the idea is to define specific specific constraints, like `ElevationConstraint`, etc. which already have a preferred solver for that specific constraint.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/TargetCalculator.scala
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+  * Target calculator that allows to calculate different attributes of a target for a given set of instants.
+  *
+  * If in doubt use {@link isDefinedAt} to make sure that values for a given time are actually calculated before
+  * accessing them, otherwise an out of bounds exception will be thrown.
+  */
+trait TargetCalculator[G] extends Calculator[G, SkyCalcResults] {
+
+  val place: Place
```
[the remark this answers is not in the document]
rpiaggio: Yes, I introduced this when I ported `SkyCalc` since `Site` is defined in `gsp-core`. If we end up moving `Enumerated` and `Site` to this library then `Place` may not be necessary. Or a `Site` will be a `Place`.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/implicits.scala
rpiaggio: It will depend on what `toIndex` returns.

Right now, it is only used in `SingleValueCalculator` which holds a single value.

But let's do that. I was thinking of adding a `GetterStrategy.Closest` already, so let's implement it and ditch `Exact`, which may or may not be exact, depending on environment.

Also, I'm thinking of reworking things so different strategies can be specified when asking the same calculator for a value. (ie: Strategy not tied to the calculator)

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/implicits.scala
rpiaggio: Yes! I'm actually working on this. And was planning to generalize to the case where you have some optic that allows you to go back and forth (`Iso`, `Split*`, `Wedge`).

I'm actually trying to write the interpolator function using `spire`'s `Number` and have instances for `Int`, `Long`, `Double`, etc. by using this mechanism.

Thank you for the `imap` name, I didn't know what to call this, and for the actual code.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/CalcGetter.scala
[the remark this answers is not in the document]
swalker2m: That's a pretty cool trick!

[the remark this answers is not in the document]
rpiaggio: I learnt it from @tpolecat

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Interval.scala
```diff
@@ -0,0 +1,194 @@
...
+import io.chrisdavenport.cats.time._
+
+/**
+  * Representation of an interval between two points in time, including the start time and excluding the end time
+  * (i.e. the interval [start, end)).
+  *
+  * Note that an interval can not represent a single point in time (i.e. start == end) because such an interval
+  * could not contain any time t for which t >= start && t < end.
+  */
+sealed abstract case class Interval protected (start: Instant, end: Instant) {
```
tpolecat, line 26: It would be nice if we could replace this with `spire.math.Interval` in a followup PR.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Schedule.scala
```diff
@@ -0,0 +1,326 @@
...
+import io.chrisdavenport.cats.time._
+import scala.annotation.tailrec
+
+/**
+  * A sequence of [[Interval]]s.
+  *
+  * The intervals are sorted by their start time and don't overlap or abut, i.e. the Schedule is always represented
+  * by the smallest possible set of Intervals.
+  */
+sealed abstract case class Schedule protected (intervals: List[Interval]) {
```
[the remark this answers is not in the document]
tpolecat: We can probably replace this with a Spire `IntervalSeq`.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/SampleRounder.scala
```diff
@@ -0,0 +1,157 @@
...
+  sealed trait LinearInterpolating
+}
+
+/**
+  * Typeclass defininig how to round values associated with `Instant`s using a certain [[RoundStrategy]].
+  *
+  * @tparam R [[RoundStrategy]] to use
+  * @tparam A type of the value
+  */
+trait SampleRounder[R, A] { self =>
```
tpolecat, line 32: I have some suggestions at https://github.com/rpiaggio/lucuma-core/pull/2

modules/math/shared/src/main/scala/gsp/math/optics/Spire.scala
```diff
@@ -0,0 +1,45 @@
...
+        java.lang.Double.isNaN(f) || java.lang.Double.isInfinite(f)
+      ),
+      _.flatMap(f => Try(Number(f)).toOption)
+    )
+
+  val numberBigInt: SplitEpi[Number, BigInt] = SplitEpi(_.toBigInt, Number.apply)
+
+  val numberSafeLong: SplitEpi[Number, SafeLong] = SplitEpi(n => SafeLong(n.toBigInt), Number.apply)
+
+  val numberBigDecimal: Iso[Number, BigDecimal] = Iso((n: Number) => n.toBigDecimal)(Number.apply)
```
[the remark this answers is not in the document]
rpiaggio: That sounds right and was my first attempt, but I couldn't get `SplitEpi`'s `coverage` law to pass.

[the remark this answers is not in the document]
tpolecat: Yeah introduce `1/3` to your generator. It's kind of a weird test, it requires a failing example.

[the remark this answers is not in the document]
tpolecat: Or I'll bet if you tested it with Rational it would probably be sufficient because you're bound to hit one with an infinite decimal expansion.

modules/math/shared/src/main/scala/gsp/math/skycalc/solver/Interval.scala
rpiaggio: I'm taking a look at Spire. We could definitely use that. It'd be nice if the bound types (open or closed) were encoded in the type, since here we want to enforce [start, end).

tpolecat:
Looks good,  thanks. The `Spire.numberBigDecimal` thing is weird because Spire's equality is defined such that `1/3 === 0.3333333333333333333333333333333333` which is why the `Iso` tests pass.

rpiaggio:
> Looks good, thanks. The `Spire.numberBigDecimal` thing is weird because Spire's equality is defined such that `1/3 === 0.3333333333333333333333333333333333` which is why the `Iso` tests pass.

Forgot to look into that, thanks for reminding. Let me try generating specific non-decimal rationals and see what happens.

rpiaggio:
> > Looks good, thanks. The `Spire.numberBigDecimal` thing is weird because Spire's equality is defined such that `1/3 === 0.3333333333333333333333333333333333` which is why the `Iso` tests pass.
> 
> Forgot to look into that, thanks for reminding. Let me try generating specific non-decimal rationals and see what happens.

Well, it seems that `Rational` is using its decimal representation for comparison. So for the moment I'll leave the `Iso`.

PR #135, opened 2020-08-20
Replace cache actions

cquiroz:
We can replace the manual cache task with a simple cache action
I'm also updating the jdk to 14 instead of being stuck in 8. I think it is ok to do so but feel free to comment otherwise


tpolecat:
Thanks, I need to try this one too.

PR #140, opened 2020-08-23
Place altitude as a coulomb quantity

cquiroz:
This PR converts `Place`'s `altitude` field to be a quantity in meters.
I previously had type aliases for PositiveInt but I noted that refined comes already with
some aliases so I'm using them instead
I also added munit and converted `PlaceSpec` to `PlaceSuite`

### DIFF
--- a/modules/math/shared/src/main/scala/gsp/math/Place.scala
+++ b/modules/math/shared/src/main/scala/gsp/math/Place.scala
@@ -5,12 +5,23 @@ package gsp.math
 
 import cats.Eq
 import cats.Show
+import coulomb.Quantity
+import coulomb.si.Meter
+import eu.timepit.refined.types.numeric.NonNegInt
 import monocle.Lens
 import monocle.macros.GenLens
 import java.time.ZoneId
 
 /** A point on Earth, given latitude, longitude and altitude in m above sea level. */
-final case class Place(latitude: Lat, longitude: Lon, altitude: Double, zoneId: ZoneId)
+final case class Place(
+  latitude:  Lat,
+  longitude: Lon,
+  altitude:  Quantity[NonNegInt, Meter],
+  zoneId:    ZoneId
+) {
+  // This is needed for the JVM based test. The JVM doesn't like interacting with the refined type
+  val altitudeDouble: Double = altitude.value.value.toDouble
+}
 
 object Place {
 
@@ -29,7 +40,7 @@ object Place {
     GenLens[Place](_.longitude)
 
   /** @group Optics */
-  val altitude: Lens[Place, Double] =
+  val altitude: Lens[Place, Quantity[NonNegInt, Meter]] =
     GenLens[Place](_.altitude)
 
   /** @group Optics */
--- a/modules/testkit/shared/src/main/scala/gsp/math/arb/ArbPlace.scala
+++ b/modules/testkit/shared/src/main/scala/gsp/math/arb/ArbPlace.scala
@@ -3,6 +3,12 @@
 
 package gsp.math.arb
 
+import coulomb._
+import coulomb.si.Meter
+import eu.timepit.refined._
+import eu.timepit.refined.auto._
+import eu.timepit.refined.numeric._
+import eu.timepit.refined.types.numeric.NonNegInt
 import gsp.math.{ Angle, Lat, Lon, Place }
 import org.scalacheck._
 import org.scalacheck.Arbitrary._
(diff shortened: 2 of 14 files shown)


modules/tests/shared/src/test/scala/gsp/math/skycalc/package.scala
```diff
@@ -3,22 +3,25 @@
...
+import eu.timepit.refined.numeric._
 import java.time.ZoneId
 
 package object skycalc {
   val GN =
     Place(
       Lat.fromAngleWithCarry(Angle.fromDoubleDegrees(19.8238068))._1,
       Angle.fromDoubleDegrees(-155.4690550),
-      4213.0,
+      4213.0.withRefinedUnit[NonNegative, Meter],
```
rpiaggio, line 16: It looks like we could remove the `.0` here and in GS. Or `withRefinedUnit` requires a `Double`?

modules/math/shared/src/main/scala/gsp/math/Place.scala
```diff
@@ -5,12 +5,23 @@ package gsp.math
...
 import monocle.Lens
 import monocle.macros.GenLens
 import java.time.ZoneId
 
 /** A point on Earth, given latitude, longitude and altitude in m above sea level. */
-final case class Place(latitude: Lat, longitude: Lon, altitude: Double, zoneId: ZoneId)
+final case class Place(
+  latitude:  Lat,
+  longitude: Lon,
+  altitude:  Quantity[NonNegInt, Meter],
```
rpiaggio, line 19: What do you think of using alias/newtype for `Quantities`? eg in this case `type Altitude = Quantity[NonNegInt, Meter]`.

modules/math/shared/src/main/scala/gsp/math/Place.scala
```diff
@@ -5,12 +5,23 @@ package gsp.math
...
+  zoneId:    ZoneId
+) {
+  // This is needed for the JVM based test. The JVM doesn't like interacting with the refined type
+  val altitudeDouble: Double = altitude.value.value.toDouble
```
rpiaggio, line 23: Can we move this to the JVM test?

modules/tests/shared/src/test/scala/gsp/math/skycalc/package.scala
cquiroz: Thanks, that's right I'll change it

modules/testkit/shared/src/main/scala/gsp/math/arb/ArbPlace.scala
```diff
@@ -14,24 +20,29 @@ trait ArbPlace {
...
+  private val MinAltitude: Int = 0
+  private val MaxAltitude: Int = 8000
 
-  val genEarthAlt: Gen[Double] =
-    Gen.chooseNum(MinAltitude, MaxAltitude)
+  val genEarthAlt: Gen[NonNegInt] =
+    Gen.chooseNum(MinAltitude, MaxAltitude).map { a =>
+      refineV[NonNegative](a)
+        .getOrElse(sys.error("Min altitude is non negative"))
+    }
```
rpiaggio, line 30: Maybe we can use `Gen(..).suchThat` here.

modules/math/shared/src/main/scala/gsp/math/Place.scala
cquiroz: We have been back and forth with type aliases. I thought about adding one here
but `Altitude` as a concept doesn't necessarily means `meters`.
I recall @tpolecat was not a big fan of type aliases but I don't recall the exact reasons
Maybe it applied more to nested types

modules/testkit/shared/src/main/scala/gsp/math/arb/ArbPlace.scala
cquiroz: I didn't know that construct seems good though it is probably not applicable in this case.
`refineV` has to be called on the value and it returns an `Either`

modules/math/shared/src/main/scala/gsp/math/Place.scala
tpolecat: Yeah I don't think an alias would be useful here.

rpiaggio: Does coulomb allow us to express the concept of length independently of what units are used?

modules/testkit/shared/src/main/scala/gsp/math/arb/ArbPlace.scala
rpiaggio: Ah, does refineV transform the value or just checks it's non negative? If it's the latter we can use .suchThat{ a => refineV[NonNegative](a).isRight} ?

cquiroz: That was my first instinct but (if I understand `suchThat` it will produce an `A` rather than a `A Refined NonNegative`

modules/math/shared/src/main/scala/gsp/math/Place.scala
cquiroz: No, unlike squants there is no type for `Length` or `Velocity`. I was expecting that but there are only `Quantity`s

PR #141, opened 2020-08-24
Model Angular velocity and proper velocity

cquiroz:
`ProperMotion` models Proper Velocity as an `Offset` but it maybe better if it gets its own
set of classes including `ProperVelocity` and Ra/Dec angular velocities

Not  that `[RightAscension|Declination]AngularVelocity` uses `Long` and it has no practical limits,
though in practice it is often a small number.


modules/math/shared/src/main/scala/gsp/math/RightAscencionAngularVelocity.scala
```diff
@@ -0,0 +1,67 @@
...
+import gsp.math.optics.SplitMono
+import monocle.Iso
+import spire.math.Rational
+import spire.std.long._
+
+/**
+  * Representation of Right Ascension angular velocity in microarcseconds per year
+  * It is used to represent proper velocity in RA
+  */
+final case class RightAscensionAngularVelocity(μasy: Quantity[Long, MicroArcSecondPerYear]) {
```
swalker2m, line 19: In the end `DeclinationAngularVelocity` is identical to `RightAscensionAngularVelocity`, though I agree we want to have distinct types for them.  One option would be to copy the pattern used for `Offset` and have a `AngularVelocity.Component` with an `Axis` type parameter (having `RA` and `Dec` options).

cquiroz: That's an interesting approach. I'll take a look.
It would be nice to have the same logic as `Offset` even though they are different beasts

cquiroz:
Ok I implemented your suggestion and indeed it is quite a bit less code

PR #144, opened 2020-08-25
Move enums to gsp-math

cquiroz:
This PR moves the enums from gsp-core to gsp-math. It includes three modules:

`schema` The db schema with the data to generate enums
`gen` Generator code
`enum` The actual generated enums

Some notes:
* I changed the package name of enums to `gsp` to match `gsp-math`. This will break downstream code. If we want to change the package name this would be a good opportunity
* The schema is exactly the same as in `gsp-core` it includes not only enums but other entities
* I'm including `Enumeration` in the `enum` package.


tpolecat:
Per the conversation in Gitter, I don't think we want gsp-math to depend on the database. I think I would prefer to just copy the enums themselves and then figure out enum generation as a follow-up. @swalker2m what do you think?

cquiroz:
The PR has been changed quite a bit based on the Slack conversation.

Changes include:
* Remove `gem` and `schema` and keep `enum` only as generated scala classes
* Rename all projects as `lucuma`, we got `lucuma-math`, `lucuma-enum`, `lucuma-testkit` and `lucuma-tests`.
* Rename the top level package to `lucuma`

cquiroz:
Ok, another try. This time with a single `core` module containing three packages:

`lucuma.math`, `lucuma.util` and `lucuma.enum`

tpolecat:
I'm seeing `lucuma.core.math`, `lucuma.core.util` and `lucuma.core.enum` ... ?

build.sbt
```diff
@@ -66,11 +66,11 @@ lazy val math = crossProject(JVMPlatform, JSPlatform)
...
 
 lazy val testkit = crossProject(JVMPlatform, JSPlatform)
-  .crossType(CrossType.Full)
+  .crossType(CrossType.Pure)
   .in(file("modules/testkit"))
-  .dependsOn(math)
+  .dependsOn(core)
   .settings(
-    name := "gsp-math-testkit",
+    name := "lucuma-testkit",
```
swalker2m, line 73: Since there will be other lucuma projects, I guess this should be `lucuma-core-testkit` (and `lucuma-core-tests` below)?

tpolecat: Good catch, agree.

cquiroz: 👍

PR #146, opened 2020-08-26
Make Place a member of Site. Remove weaver.

rpiaggio:
* Change weaver test suites to munit (and remove weaver dependency).
* Make `Site` a subclass of `Place`.

I'm on the fence whether to keep the concept of `Place` or not, now that we have `Site` here. It does serve a purpose to exercise code that should work for any (latitude, longitude, altitude), but that's about it. It doesn't seem we will be needing it for the moment from client code.


cquiroz:
I'm not sure if we should keep `Place` seems a bit redundant with `Site` and there is a single implementation.

tpolecat:
The `Place`/`Site` distinction seems ok to me, if only to give us a generic geo location to pass to Skycalc. I'm less convinced about the subtyping relationship however, because natural equality is inconsistent; i.e., `p == s` but `s != p` … in general this is very hard to remedy. What if `Site` had a `Place` member?

rpiaggio:
Ok, I think that's it: `Place` is all the abstraction we need at the lower layer of calculation.

I'll se how it looks with composition then.

modules/core/shared/src/main/scala/lucuma/core/math/Place.scala
```diff
@@ -3,47 +3,59 @@
...
+import coulomb.cats.implicits._
+import eu.timepit.refined.cats._
 
 /** A point on Earth, given latitude, longitude and altitude in m above sea level. */
-final case class Place(
-  latitude:  Lat,
-  longitude: Lon,
-  altitude:  Quantity[NonNegInt, Meter],
-  zoneId:    ZoneId
+class Place(
```
swalker2m, line 19: Should this maybe go back to being a `final case class` since we're not trying to extend it anymore?

rpiaggio: Yup, good catch. It can be a final case class again and we don't need constructor and can derive Eq and Show.

swalker2m: :+1:

PR #147, opened 2020-08-27
Model package

cquiroz:
This PR moves several classes from gsp-core to lucuma-core. The code is essentially the same with the following changes:

 * Moved to the `lucuma.core.model` package
 * Converted `scalatest` tests to `munit`
 * Removal of unused plugins


modules/core/shared/src/main/scala/lucuma/core/instances/TimeInstances.scala
```diff
@@ -0,0 +1,33 @@
...
+  implicit val YearOrder: Order[Year]                   = naturalOrder(_ isBefore _)
+  implicit val LocalTimeOrder: Order[LocalTime]         = naturalOrder(_ isBefore _)
+  implicit val LocalDateOrder: Order[LocalDate]         = naturalOrder(_ isBefore _)
+  implicit val LocalDateTimeOrder: Order[LocalDateTime] = naturalOrder(_ isBefore _)
+
+  implicit val DurationEq: Eq[Duration] =
+    Eq.by(d => (d.getSeconds, d.getNano))
+
+}
+object time extends TimeInstances
```
rpiaggio, line 33: These are in `cats-time` which we are pulling in now.

modules/tests/shared/src/test/scala/lucuma/core/model/EphemerisKeySuite.scala
```diff
@@ -0,0 +1,39 @@
...
+import lucuma.core.model.arb._
+import lucuma.core.util.arb._
+
+import cats.{ Eq, Show }
+import cats.kernel.laws.discipline._
+import lucuma.core.math.laws.discipline._
+import munit._
+import org.scalacheck.Prop._
+
+final class EphemerisKeySuite extends FunSuite with DisciplineSuite with ScalaCheckSuite {
```
rpiaggio, line 15: `DisciplineSuite` already extends `ScalaCheckSuite`, which extends `FunSuite`.

tpolecat:
:+1: if we use Skunk (which I hope we will) we won't need `Timestamp` but we can leave it in for now.

PR #149, opened 2020-08-31
Move Display typeclass to it's own file & modify.

toddburnside:
Modifications from the slack conversation.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
@@ -1,0 +1,13 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.util
+
+/**
+  * Typeclass for things that can be shown in a user interface.
+  * @group Typeclasses
+  */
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
```diff
@@ -0,0 +1,13 @@
...
+package lucuma.core.util
+
+/**
+  * Typeclass for things that can be shown in a user interface.
+  * @group Typeclasses
+  */
+trait Display[A]     {
+  def shortName(a: A): String
+  def longName(a:  A): String = shortName(a)
+}
```
swalker2m, line 13: Sorry, late to the party.

Should we add some syntax to permit `a.shortName` and `a.longName` in a future PR?  And an

```scala
object Display {
  def apply[A](implicit ev: Display[A]): Display[A] = ev
}
```

for convenient summons?

cquiroz: Great idea

tpolecat: Yes, good idea.

toddburnside: Guess I was a little quick with the merge button. I'll create another PR with this.

PR #150, opened 2020-09-01
Some Display typeclass related renaming.

toddburnside:
The commit creating the Display syntax, constructors and summoner was inadvertently pushed directly to master. Please review all files in their entirety.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
@@ -17,23 +17,23 @@ object Display {
 
   /** Create an instance of Display using the provided functions. 
    * 
-   * @param toShort function that maps A to the shortName
-   * @param toLong function that maps A to the longName
+   * @param toShortName function that maps A to the shortName
+   * @param toLongName function that maps A to the longName
   */
-  def from[A](toShort: A => String, toLong: A => String): Display[A] = 
+  def from[A](toShortName: A => String, toLongName: A => String): Display[A] = 
     new Display[A] {
-      def shortName(a: A) = toShort(a)
-      override def longName(a: A)  = toLong(a)
+      def shortName(a: A) = toShortName(a)
+      override def longName(a: A)  = toLongName(a)
     }
 
   /** Create an instance of Display using the provided function 
    * for the shortName. The longName will be the same as the 
    * shortName.
    * 
-   * @param toShort function that maps A to the shortName
+   * @param toShortName function that maps A to the shortName
    */
-  def fromShort[A](toShort: A => String): Display[A] = 
+  def fromToShortName[A](toShortName: A => String): Display[A] = 
     new Display[A] {
-      def shortName(a: A) = toShort(a)
+      def shortName(a: A) = toShortName(a)
     }
 }
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
@@ -3,6 +3,8 @@
 
 package lucuma.core.util
 
+import cats.Contravariant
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
```diff
@@ -17,23 +17,23 @@ object Display {
 
   /** Create an instance of Display using the provided functions. 
    * 
-   * @param toShort function that maps A to the shortName
-   * @param toLong function that maps A to the longName
+   * @param toShortName function that maps A to the shortName
+   * @param toLongName function that maps A to the longName
   */
-  def from[A](toShort: A => String, toLong: A => String): Display[A] = 
+  def from[A](toShortName: A => String, toLongName: A => String): Display[A] =
```
tpolecat, line 23: Contravariant instance constructors like this are conventionally called `by` so I suggest `by` and `byShortName` for these.

modules/core/shared/src/main/scala/lucuma/core/util/syntax/package.scala
```diff
@@ -3,6 +3,11 @@
 
 package lucuma.core.util
```
tpolecat, line 4: Suggest moving this up to `lucuma.core.syntax` so we don't have to dig for it. This ended up being a problem in OCS.

modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
toddburnside: Will do. Wasn't certain about the naming, anyway.

modules/core/shared/src/main/scala/lucuma/core/util/syntax/package.scala
toddburnside: Makes sense.

modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
cquiroz: @tpolecat What does it make it `Contravariant`?

modules/core/shared/src/main/scala/lucuma/core/util/syntax/package.scala
cquiroz: Do we want to do `import lucuma.core.syntax._` or `import lucuma.core.syntax.all._`

tpolecat: I think `import lucuma.core.syntax.all._` … I think should do what cats does in these cases.

modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
toddburnside: @tpolecat Do you want just the constructors themselves renamed, or the parameters as well?

modules/core/shared/src/main/scala/lucuma/core/util/syntax/package.scala
toddburnside: There is currently a lot of syntax under `lucuma.core.math.syntax`. Should we move that, too?

tpolecat: Hm. I was initially thinking that everything in `gsp.math` would move into `lucuma.core` rather than `lucuma.core.math` but I guess we need to discuss it more. It's nice to have all the instances and syntax in one place.

toddburnside: We can discuss in slack, and I'll move the `lucuma.core.math` syntax to `lucuma.core` in another PR if we decide to do so.

PR #151, opened 2020-09-01
Migrate Night classes + unification.

rpiaggio:
* Migrate `Night`, `ObservingNight`, `LocalObservingNight` and `TwilightBoundedNight` from `gsp-core`.
* `TwilightType` is now an `Enumerated`.
* Unification of `Constants`.
* Add `Prism[String, A]` generator to `Enumerated` object.
* Use `Interval` and `Schedule` when possible instead of `start`/`end` `Instant`s.
* `Interval` and `Schedule` due to wider adoption were moved to `math` package. (The idea is still to see if in the future we can substitute them by `spire`'s `Interval` and `IntervalSeq`.)


modules/core/shared/src/main/scala/lucuma/core/math/Constants.scala
```diff
@@ -0,0 +1,67 @@
...
+  val KZen: Double = 0.172
+
+  /** Used in numerical differentiation to find Earth velocity. */
+  val EarthDiff: Double = 0.05
+
+  /*
+   * Physical Constants
+   */
+
+  /** Speed of light in meters per second. Exact. */
```
cquiroz, line 39: There is already a `PhysicalConstants` class I think

cquiroz: Oh I see, it was deleted

modules/core/shared/src/main/scala/lucuma/core/model/LocalObservingNight.scala
```diff
@@ -0,0 +1,148 @@
...
+}
+
+object LocalObservingNight extends LocalObservingNightOptics {
+
+  /** The hour, in the local time zone, at which the night is considered to
+    * officially start.
+    *
+    * @group Constants
+    */
+  val StartHour: Int =
```
cquiroz, line 65: Maybe it should be private, it maybe confusing just if used as an `Int`

modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
```diff
@@ -44,11 +45,12 @@ object Enumerated {
       def tag(a: A): String = a.productPrefix
     }
 
+  def fromTag[A](implicit ev: Enumerated[A]): Prism[String, A] =
```
cquiroz, line 48: Nice, were you able to test it?

modules/tests/jvm/src/test/scala/lucuma/core/math/skycalc/TwilightTypeJVM.scala
```diff
@@ -0,0 +1,14 @@
...
+
+package lucuma.core.math.skycalc
+
+import lucuma.core.enum.TwilightType
+
+// For use from Java when testing SkyCalc (enum is a reserved keyword)
+case class TwilightTypeJVM(wrapped: TwilightType) {
+  val horizonAngle = wrapped.horizonAngle.toAngle.toSignedDoubleDegrees
+}
+object TwilightTypeJVM {
```
cquiroz, line 12: Missing line?

modules/core/shared/src/main/scala/lucuma/core/model/TwilightBoundedNight.scala
```diff
@@ -0,0 +1,232 @@
...
+
+trait TwilightBoundedNightOptics {
+
+  /** @group Optics */
+  val twilightType: Lens[TwilightBoundedNight, TwilightType] =
+    GenLens[TwilightBoundedNight](_.twilightType)
+
+  /** @group Optics */
+  val observingNight: Lens[TwilightBoundedNight, ObservingNight] =
+    GenLens[TwilightBoundedNight](_.toObservingNight)
```
swalker2m, line 219: I see this is in the old code and maybe I'm confused, but is this safe?  The constructor is marked `private` but I assume a set will make a `copy` swapping out the `ObservingNight` in the `TwilightBoundedNight` (i.e., they become inconsistent).

modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
toddburnside: The only thing I don't like here is that we have a `fromTag` on the trait. too. And the meaning of the 2 methods is quite different.

modules/core/shared/src/main/scala/lucuma/core/enum/TwilightType.scala
```diff
@@ -0,0 +1,60 @@
...
+  /** Select the member of TwilightType with the given tag, if any. */
+  def fromTag(s: String): Option[TwilightType] =
+    all.find(_.tag === s)
+
+  /** Select the member of TwilightType with the given tag, throwing if absent. */
+  def unsafeFromTag(s: String): TwilightType =
+    fromTag(s).getOrElse(throw new NoSuchElementException(s"TwilightType: Invalid tag: '$s'"))
+
+  /** @group Typeclass Instances */
+  implicit val SiteEnumerated: Enumerated[TwilightType] =
```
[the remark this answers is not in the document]
tpolecat: I have taken to calling them `EnumeratedTwilightType`, i.e., the same as the type, without the brackets. It's boring but at least I can remember it.

modules/core/shared/src/main/scala/lucuma/core/enum/TwilightType.scala
```diff
@@ -0,0 +1,60 @@
...
+    new Enumerated[TwilightType] {
+      def all = TwilightType.all
+      def tag(a:                    TwilightType)         = a.tag
+      override def unsafeFromTag(s: String): TwilightType =
+        TwilightType.unsafeFromTag(s)
+    }
```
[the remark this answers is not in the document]
toddburnside: I think the argument against using `Enumerated.of` here is that it uses the `productPrefix` as the tag, which is not safe wrt refactoring.

modules/core/shared/src/main/scala/lucuma/core/model/TwilightBoundedNight.scala
```diff
@@ -0,0 +1,232 @@
...
+import java.time.Instant
+import java.time.LocalDate
+import java.time.LocalDateTime
+
+/** The start and end of a particular night at a [[lucuma.core.enum.Site]]. */
+final case class TwilightBoundedNight private (
+  twilightType:     TwilightType,
+  toObservingNight: ObservingNight,
+  interval:         Interval
+) extends Night {
```
swalker2m, line 26: In general, `TwilightType` and `ObservingNight` are enough to define the interval so there is opportunity to make these inconsistent.  I wonder if we should remove the redundancy?  Maybe we could move the `Interval` into a `val` calculated from the `TwilightType` and `Interval` instead of supplied as a constructor argument.

swalker2m: Then there's no need to make the constructor `private` either, I think.

modules/core/shared/src/main/scala/lucuma/core/enum/TwilightType.scala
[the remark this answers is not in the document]
swalker2m: Yeah, that's a good point. :+1:

modules/core/shared/src/main/scala/lucuma/core/enum/TwilightType.scala
```diff
@@ -0,0 +1,60 @@
...
+import lucuma.core.math.syntax.all._
+import lucuma.core.util.Enumerated
+
+/**
+  * Definition for how the range from sunset to sunrise should be defined for
+  * a night. There are various standard options for definition where the night
+  * begins and ends which are represented as static constants in side this
+  * class.
+  */
+sealed abstract class TwilightType(val tag: String, val horizonAngle: Declination)
```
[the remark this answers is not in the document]
rpiaggio: I went with this concept because it seemed to me that it's an elevation, which is an angle that can only have values between [-90, 90] degrees, and `Declination` models that accurately. We are also using `Declination` for target/sun/moon elevation in `SkyCalcResults`.

To model `Latitude` we made an alias `type Latitude = Declination`, maybe we should define another `Elevation` alias and use it in these cases?

modules/core/shared/src/main/scala/lucuma/core/enum/TwilightType.scala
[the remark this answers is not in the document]
rpiaggio: That was a copy/paste that I forgot to rename :). Anyway, it's good to have a common criteria for naming typeclasses.

modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala
```diff
@@ -56,21 +56,21 @@ final case class ProperMotion(
 }
 
 object ProperMotion extends ProperMotionOptics {
-  import PhysicalConstants.{ AstronomicalUnit, TwoPi }
+  import Constants.{ AstronomicalUnit, TwicePi }
```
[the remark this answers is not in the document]
rpiaggio: Ok, I changed it back. `TwicePi` seemed consistent with `HalfPi` ;)

modules/core/shared/src/main/scala/lucuma/core/math/skycalc/ImprovedSkyCalcMethods.scala
```diff
@@ -402,17 +403,17 @@ trait ImprovedSkyCalcMethods {
...
     accusun(jd2, 0.0, 0.0, ras, decs, dists, topora, topodec, x2, y2, z2)
     accusun(jd, 0.0, 0.0, ras, decs, dists, topora, topodec, x, y, z)
     Xdot =
-      KmSInAUDay * (x2.d - x1.d) / (2.0 * EarthDiff * SpeedOfLight) /* numerical differentiation */
+      KmSInAUDay * (x2.d - x1.d) / (2.0 * EarthDiff * SpeedOfLight / 1000) /* numerical differentiation */
     Ydot =
-      KmSInAUDay * (y2.d - y1.d) / (2.0 * EarthDiff * SpeedOfLight) /* crude but accurate */
-    Zdot = KmSInAUDay * (z2.d - z1.d) / (2.0 * EarthDiff * SpeedOfLight)
+      KmSInAUDay * (y2.d - y1.d) / (2.0 * EarthDiff * SpeedOfLight / 1000) /* crude but accurate */
+    Zdot = KmSInAUDay * (z2.d - z1.d) / (2.0 * EarthDiff * SpeedOfLight / 1000)
```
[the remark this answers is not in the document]
rpiaggio: SkyCalc used a `Double` constant in km/s, and RadialVelocity uses an `Int` in m/s. I unified towards the latter, so in SkyCalc I added the division to get km/s.

That said, we could change the `KmSInAUDay` constant to `MetersPerSecondInAUPerDay` to get the same result, so I'll do that.

modules/core/shared/src/main/scala/lucuma/core/model/TwilightBoundedNight.scala
rpiaggio: Thank you for catching this. This is code is not that old, I introduced it a while ago when I migrated this class to `gsp-core`.

I'm tempted to move the `Interval` to a lazy val in the class and make it an `Option`, but then we cannot inherit from `Night` anymore. I'll try fixing the `Lens`es so that they recompute.

modules/core/shared/src/main/scala/lucuma/core/model/TwilightBoundedNight.scala
rpiaggio: This ties in with the previous comment. I completely agree but the thing is that when you take into account that given a `TwilightType`, a `Place` and a `LocalDate`, the sun may not set or rise.

This does not happen for our 2 current `Site`s. So, should we: a) Do an `unsafe` get? b) Make `interval`, `start`, `end` `Option`al and not inherit from `Night` anymore?

modules/core/shared/src/main/scala/lucuma/core/enum/TwilightType.scala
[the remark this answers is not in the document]
swalker2m: Allowing for negative values, the `horizonAngle` is the Sun's altitude. At any rate the Sun will have a real declination at that time that won't be the horizon angle (though it will be related to the horizon angle through the observer's latitude).

I think I might make it an `Angle` and call it good, but if you want to investigate making `Declination` have or be a kind of more general [-90, 90] thing that might be good.  What do you think @tpolecat?

modules/core/shared/src/main/scala/lucuma/core/model/TwilightBoundedNight.scala
rpiaggio: Just to document here what we talked in Slack: 

- It seems that we should always allow for the creation of the object and compute the interval in-class. We could allow for empty `Interval`s. We should look into this when we evaluate switching to spire's `Interval`.
- For the moment, we are preventing inconsistencies by disallowing `copy` and not having `Lens`es, just `Getter`s.

PR #153, opened 2020-09-02
Migrate geom classes from gsp-core

rpiaggio:


modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/JtsDemo.scala
```diff
@@ -4,38 +4,40 @@
...
 
   // TODO: should come from the FPUnit enum in core
-  val ifuOffset: Offset       =
+  val ifuOffset: Offset =
     Offset.Zero
 
   // TODO: when we move this core there should be an enum
-  val sideLooking: Boolean    =
+  val sideLooking: Boolean =
     true
```
swalker2m, line 41: This code does differ a bit in `gsp-core`, referencing a `PortDisposition` enumeration here for example.  Will those changes get picked up as part of this process?

cquiroz: We had two versions for GMOS geometries. This PR will unify them using the one form gsp core which has a proper use of enumerations

cquiroz:
This is JTS demo:
JTS Demo 2020-09-03 09-26-36

modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/JtsDemo.scala
rpiaggio: Yes, @cquiroz was migrating this. Thank you @cquiroz !

cquiroz:
@swalker2m what do you think about putting all the geometries and geometry related classes in a separate geom project inside lucuma-core

swalker2m:
> What do you think about putting all the geometries and geometry related classes in a separate `geom` project inside `lucuma-core`

I went back and forth on that back when I was writing this code and had a discussion with @tpolecat about it too. In the end we decided to keep it all part of the same module but I'm not opposed to the idea.  It would be pretty small.

Ultimately we'll need the geometries to do calculations like determining whether a guide star is reachable by a guide probe.  I'm not sure where that logic will be needed, but if we split `geom` out then that logic couldn't be in the main part of `lucuma-core`?  Maybe we should leave it all together for now until it becomes clear where we're going?  What do you think?

cquiroz:
That's fine with me too. I guess the only concern was making test a bit slower.
I noted that locally but on CI seems ok.

modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/JtsDemo.scala
```diff
@@ -4,45 +4,49 @@
...
 
   // Shape to display
   val shapes: List[ShapeExpression] =
     List(
-      GmosOiwfsProbeArm.shapeAt(posAngle, guideStarOffset, offsetPos, ifuOffset, sideLooking),
-      GmosOiwfsProbeArm.patrolFieldAt(posAngle, offsetPos, ifuOffset, sideLooking),
+      GmosOiwfsProbeArm.shapeAt(posAngle, guideStarOffset, offsetPos, fpu, port),
+      GmosOiwfsProbeArm.patrolFieldAt(posAngle, offsetPos, fpu, port),
+      GmosScienceAreaGeometry.imaging ⟲ posAngle,
       GmosScienceAreaGeometry.imaging ↗ offsetPos ⟲ posAngle
```
swalker2m, line 50: This all looks good, but I am curious whether all the changes in the old `gsp-core` got picked up.  It evolved a bit in `gsp-core` after I added it there.  It _really doesn't matter_ for `JtsDemo` but it'd be good to know all the substantive changes to the rest are there.

Here for example in line 50 it is

```scala
GmosScienceAreaGeometry.shapeAt(posAngle, offsetPos, fpu)
```

in `gsp-core`.

modules/core/shared/src/main/scala/lucuma/core/math/geom/GmosOiwfsProbeArm.scala
```diff
@@ -0,0 +1,149 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math.geom
```
cquiroz, line 4: We may want this to be `lucuma.core.geom`? The instrument definitions may not be exactly math.

modules/core/shared/src/main/scala/lucuma/core/math/geom/GmosScienceAreaGeometry.scala
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math.geom
```
cquiroz, line 4: Same as above.

cquiroz (approved):
LGTM

modules/tests/jvm/src/main/scala/gsp/math/geom/jts/demo/JtsDemo.scala
rpiaggio: `GmosOiwfsProbeArm` and `GmosScienceAreaGeometry` were copied from `gsp-core`. The only changes were the imports. Are there other classes from `gsp-core` that I'm missing?

modules/core/shared/src/main/scala/lucuma/core/math/geom/GmosOiwfsProbeArm.scala
swalker2m: Yeah, maybe let's put all the geom stuff in `lucuma.core.geom` for now.

PR #157, opened 2020-09-03
Package reorganization and DST considerations

rpiaggio:
* Move `optics` package out of `math`.
* Move `geom` package out of `math`.
* When generating arbitrary `ZonedDateTime`s, consider the special case of DST transition instants.
* Fix `Interval.toFullDays` to consider potential DST changes in the middle.
* Upgrade `cats` to `2.2.0`.

There's some auto formatting that happened when I saved the modified files. If this bothers anyone, I can revert. In particular, infix notation was changed to dot notation for all method calls.


modules/core/js/src/main/scala/lucuma/core/math/geom/svg/package.scala
```diff
@@ -1,7 +1,7 @@
 // Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
```
cquiroz, line 1: Should the file be moved as well?

modules/core/shared/src/main/scala/lucuma/core/geom/Shape.scala
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.geom
+
+import lucuma.core.math.Offset
```
cquiroz, line 6: Shape seems added and deleted. Could instead be git moved?

cquiroz (changes requested):
It seems correct but some files should not only be removed but moved to the correct directory.

modules/core/js/src/main/scala/lucuma/core/math/geom/svg/package.scala
rpiaggio: Ooops, I overlooked the js/jvm files. Moving them too.

cquiroz (approved):
LGTM

swalker2m (approved):
It looks great to me.  I think we should re-agree on a comment format though or a lot of PRs may be just reformatting comments.  I'm fine with any of the options.

modules/core/shared/src/main/scala/lucuma/core/geom/Shape.scala
rpiaggio: Git didn't pick it up as a move, only for this file, probably because it's a small file and all the scaladoc was reformatted, going over the similarity threshold. I agree this is far from ideal since it breaks continuity.

We *could* split the commit in two, one with the move and another one with the modification and then git will probably pick it up.

rpiaggio:
> It looks great to me. I think we should re-agree on a comment format though or a lot of PRs may be just reformatting comments. I'm fine with any of the options.

I've reconfigured scalafmt to do JavaDoc style.

swalker2m:
> Adhere to Javadoc fomat in Scaldoc comments.

:+1:

PR #158, opened 2020-09-04
Migrate Semester. Rearrange syntax/parser/arb.

rpiaggio:
* Migrate  from `gsp-core`: `Semester`, specific enum implicits and parsers.
* Move some `syntax`, `parser`, `arb` implementations out of `math` package and under `model`, `optics` or root.
* `import lucuma.core.syntax.all._` is available.
* Use `syntax.time` when possible. Added `ZonedDateTime` `+`/`-`.

Some formatting.


modules/core/shared/src/main/scala/lucuma/core/enum/package.scala
```diff
@@ -0,0 +1,85 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core
+
+import java.time.Month
+import java.time.Month._
+import cats.implicits._
```
swalker2m, line 8: Will this get the cats updates, (e.g. to `import cats.syntax.all._`)?

modules/core/shared/src/main/scala/lucuma/core/enum/package.scala
```diff
@@ -0,0 +1,85 @@
...
+import lucuma.core.optics.syntax.ToPrismOps
+
+// The members of this package are generated from database tables, which are the source of truth.
+// See project/gen2.scala for details. Associations with other model types, as needed, are provided
```
swalker2m, line 12: We can delete the first two sentences here since they no longer apply.

swalker2m (approved):
Looks great. :+1:

modules/core/shared/src/main/scala/lucuma/core/enum/package.scala
rpiaggio: Yup, slipped through the cracks. Sending another commit.

modules/core/shared/src/main/scala/lucuma/core/enum/package.scala
rpiaggio: :+1:

rpiaggio: Hmm... Actually we could move the extra methods to the enums themselves then. I'll do that.

swalker2m: Oh, right, that makes more sense now.

swalker2m (approved):
Looks good.  :+1:

PR #162, opened 2020-09-09
GID typeclass

tpolecat:
This adds a **typeclass** for Gemini internal ids that have a canonical string representation and a type hint to help with debugging; i.e., `u-26fd21b3` identifies a `User`, at least in the SSO database. I don't know that there's a huge need to make the tags globally unique but we might have a wiki page for them or something.

The idea is that `Gid[A]` witnesses that
- `A` is isomorphic to `PosLong`
- `A` has a letter tag
- `A` has a canonical string representation `<letter>-<positive long in lowercase hex>`

Declare an instance thus:

```scala
case class Foo(id: Foo.Id, ...)
object Foo {
  case class Id(value: PosLong)
  object Id {
    implicit val GidId: Gid[Id] =
      Gid.instance(refineMV('f'), _.value, apply) // tag, to PosLong, from PosLong
  }
}
```

Within a database we can keep id types distinct by declaring domains, and auto-generate them with sequences.

```sql
CREATE DOMAIN foo_dom AS VARCHAR CHECK(VALUE ~ '^f-([1-9a-f][0-9a-f]*)$');
CREATE SEQUENCE foo_seq START WITH 256;
CREATE TABLE foo (
  foo_id  varchar NOT NULL PRIMARY KEY DEFAULT 'f-' || to_hex(nextval('foo_seq')),
  ...
)
```

Via the `fromString` prism provided by `Gid` we can derive codecs for Circe, doobie, skunk, etc., as well as GraphQL custom scalars.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Gid.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Gid.scala
@@ -12,6 +12,7 @@ import eu.timepit.refined.types.numeric.PosLong
 import eu.timepit.refined.numeric.Positive
 import eu.timepit.refined._
 import eu.timepit.refined.api.Refined
+import _root_.cats.kernel.BoundedEnumerable
 
 /**
  * A typeclass for Lucuma identifiers, which are of the form T-26fd21b3 where T is a constant,
@@ -43,7 +44,8 @@ import eu.timepit.refined.api.Refined
 final class Gid[A](
   val tag:        Char Refined Letter,
   val isoPosLong: Iso[A, PosLong]
-) extends Order[A]
+) extends BoundedEnumerable[A]
+     with Order[A]
      with Show[A] {
 
   // We use this in a few places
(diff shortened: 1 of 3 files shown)


swalker2m (approved):
I've been using (an earlier copy of) this for a while and it works out well.  How would you feel about adding a `one` or whatever to get the first valid id and a `next` to get the next one?  Or is that nuts?

Modulo however you work with refined:

```scala
  val one: A =
    isoPosLong.reverseGet(1L)

  def next(a: A): A =
    isoPosLong.reverseGet(isoPosLong.get(a) + 1L)
```

tpolecat:
> How would you feel about adding a one or whatever to get the first valid id and a next to get the next one? Or is that nuts?

My thinking was that we would only construct these in the database to ensure uniqueness … but I guess if you have an in-memory database where you're faking things it could be useful. I'll add that.

cquiroz:
Looks very interesting. We'd likely need to use the ids for urls as well so they should be stable.

modules/core/shared/src/main/scala/lucuma/core/util/Gid.scala
```diff
@@ -0,0 +1,93 @@
...
+
+  // Order
+  final override def compare(a: A, b: A): Int =
+    isoPosLong.get(a).value compare isoPosLong.get(b).value
+
+}
+
+object Gid {
+
+  def apply[A](implicit ev: Gid[A]): ev.type = ev
```
cquiroz, line 87: I normally write summons with a return type `Gid[A]`. Why is it `ev.type` here?

tpolecat: It doesn't matter really, this just means it has to return the one it gets through implicit scope, rather than constructing a new one or something. And if it happens to have a more precise type that's what you'll get back.

cquiroz (approved):
LGTM

tpolecat:
@swalker2m added a `BoundedEnumerable` instance, which gives you min/max/prev/next

PR #164, opened 2020-09-10
add circe to core, add codec for Gid

tpolecat:


modules/tests/shared/src/test/scala/lucuma/core/util/GidSuite.scala
```diff
@@ -27,6 +29,7 @@ final class GidSuite extends DisciplineSuite {
   checkAll("Gid[Id]", OrderTests[Id].order)
   checkAll("Gid[Id].isoPosLong", IsoTests(Gid[Id].isoPosLong))
   checkAll("Gid[Id].fromString", PrismTests(Gid[Id].fromString))
+  checkAll("Gid[Id]", CodecTests[Id].unserializableCodec)
```
cquiroz, line 32: Why not `CodecTest[Gid[Id]]`

tpolecat: We're testing the codec for `Id`, same as `OrderTest` above.

tpolecat: So `OrderTests` and `CodecTests` are checking that the derived `Order` and `Encoder/Decoder` instances for `Id` are correct.

cquiroz: Ok, now I got it, thanks

tpolecat:
Actually I'm following up with `GidLaws` because this is a typeclass and testing against an example isn't really the way to do it. I'm going to close this one.

PR #165, opened 2020-09-10
add SSO data types

tpolecat:
This adds some SSO data types to `lucuma.core.model`

- `Access` for access levels (these are comparable)
- `Role` for roles (which have an access level)
- `User` for users, guests, services (which have one or more `Role`s)
- `OrcidId` and `OrcidProfile` (standard users have these)

It also adds Discipline suites for `Gid` and `Enumerated` so you can test that your instances are ok.


swalker2m (approved):
This looks great! :+1:

PR #167, opened 2020-09-11
Move string case transformation to StringOps

swalker2m:
Adds the various `String` case transformations to `StringOps` to make them generally available.  In particular, I need the "screaming snake case" transformation to generate a GraphQL enum type from `Enumerated` instances.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/syntax/String.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/syntax/String.scala
@@ -43,34 +43,22 @@ object StringOps {
   // Case transformation implementation adapted from:
   // https://github.com/circe/circe-generic-extras/blob/master/generic-extras/src/main/scala/io/circe/generic/extras/Configuration.scala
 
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/syntax/String.scala
```diff
@@ -16,6 +18,60 @@ final class StringOps(val self: String) extends AnyVal {
...
+    s => {
+      val partial = basePattern.matcher(s).replaceAll(replacement)
+      caze.convert(swapPattern.matcher(partial).replaceAll(replacement))
+    }
+
+  private val snakeCaseTransformation: String => String =
+    transformation("$1_$2", LowerCase)
+
+  private val screamingSnakeCaseTransformation: String => String =
+    transformation("$1_$2", UpperCase)
```
swalker2m, line 70: I don't know, I guess we could drop all the `Case` stuff and just do `s.toSnakeCase.toUpperCase` where needed.

tpolecat (approved):

PR #173, opened 2020-09-28
Add magnitudes field to target

cquiroz:
This PR adds several smallish changes that are required for catalogues, these includes:

* Adds fields for magnitudes and catalogue Id
* Add target optics
* Increase precision of Magnitude Values
* Port some tests to munit


modules/core/shared/src/main/scala/lucuma/core/math/MagnitudeValue.scala
```diff
@@ -3,26 +3,28 @@
...
+import cats.Order
+import cats.Show
 import lucuma.core.optics.Format
+import spire.math.Rational
 
 /**
  * Exact magnitude value represented as an int with the original value scaled up
  *
- * @param scaledValue This magnitude integral value, as the original multiplied by 100. value is dimensionless
+ * @param scaledValue This magnitude integral value, as the original multiplied by 1000. value is dimensionless
```
cquiroz, line 16: Simbad provides mag errors with 3 decimals precision

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -3,16 +3,32 @@
...
+import monocle.function.FilterIndex
+import monocle.macros.GenLens
+import monocle.std.either._
+import monocle.std.option
 
 /** A target of observation. */
-@Lenses final case class Target(name: String, track: Either[EphemerisKey, ProperMotion])
+final case class Target(
+  name:       NonEmptyString,
+  catalogId:  Option[NonEmptyString],
```
cquiroz, line 26: This field is to keep a link with the catalogue as the `name` can be changed by the user

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -3,16 +3,32 @@
...
+  track:      Either[EphemerisKey, ProperMotion],
+  magnitudes: SortedMap[MagnitudeBand, Magnitude]
```
cquiroz, line 28: This assumes we only have a single value per band.

cquiroz:
Note this keeps the magnitude as a constant value rather than time varying one.
That needs to be added later on

modules/core/shared/src/main/scala/lucuma/core/enum/MagnitudeBand.scala
```diff
@@ -65,4 +67,12 @@ object MagnitudeBand {
         MagnitudeBand.unsafeFromTag(s)
     }
 
+  /** @group Typeclass Instances */
+  implicit val MagnitudeBandOrder: Order[MagnitudeBand] =
+    Order.by(_.center)
```
swalker2m, line 72: `Enumerated` extends `Order` of course so this sets up a situation where there are two implicit instances for `Order[MagnitudeBand]`.  Is it important that they be ordered by wavelength?  Maybe we should change the `Enumerated` instance such that the ordering is by wavelength?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
swalker2m: If it is non-sidereal then the horizons id is in the `EphmerisKey`.  I'm not sure it makes sense for a non-sidereal target.  If this is a sidereal target coming from a catalog then the proper motion data will be coming from the catalog so maybe it makes sense to include it in `ProperMotion`?

Also, should we pair the catalog id with the catalog instead of just a `NonEmptyString`?  Is there an enumeration of catalogs that we could use?  For example:

```
final case class CatalogId(catalog: HypotheticalCatalogEnum, id: NonEmptyString)
```

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -24,18 +40,94 @@ object Target {
...
+
   /** A target order based on tracking information.  For sidereal targets this
     * roughly means by base coordinate without applying proper motion.  For
     * non-sidereal this means by `EphemerisKey`.
+    *
+    * Not implicit.
     */
-  implicit val TargetTrackOrder: Order[Target] =
-    Order.by(t => (t.track, t.name))
+  val TargetTrackOrder: Order[Target] =
```
swalker2m, line 52: Is this fallout from having multiple `Order` implicits for the band?

swalker2m (commented):
I think this looks great.  I'm just not sure about multiple implicit `Order`s.

modules/core/shared/src/main/scala/lucuma/core/enum/MagnitudeBand.scala
cquiroz: Good catch. I didn't think about it. I'd remove this, as the order is really needed only for `SortedMap`

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
cquiroz: It doesn't seem possible to define an order for `SortedMap` (which is kind of ironic), thus the `implicit Order` wouldn't be consistent with `Eq`. I preferred then to make it non implicit and create a separate `Eq`

cquiroz:
There is a `CatalogName` enum in `lucuma-catalog` that I could move here and create `CatalogId`.

cquiroz:
Added a few fixes as discussed in comments and slack:

* Renamed `ProperMotion` to `SiderealTracking`
* Added `CatalogName` enum and `CatalogId` class
* Include the `CatalogId` only for Sidereal targets

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
tpolecat: Why do we need to know the catalog id for sidereal targets?

modules/core/shared/src/main/scala/lucuma/core/enum/CatalogName.scala
```diff
@@ -0,0 +1,19 @@
...
+
+sealed trait CatalogName extends Product with Serializable
+
+object CatalogName {
+  case object Simbad extends CatalogName
+  case object Horizon extends CatalogName
+  case object Gaia extends CatalogName
+
+  /** @group Typeclass Instances */
+  implicit val SectionVisibilityStateEnumerated: Enumerated[CatalogName] =
```
tpolecat, line 16: `EnumeratedCatalogName`?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
tpolecat: Seems fine. I think this is out the window anyway when we get time-varying magnitudes.

modules/core/shared/src/main/scala/lucuma/core/enum/CatalogName.scala
tpolecat: I'm a little skeptical that this needs to be in the model since it's unused, and I'm not sure it would ever need to be in the database. Where do you see this going?

cquiroz:
@tpolecat The idea with the catalogId is to keep the original source of a search in case you need to re run it

Let's say a user search for target X, in the response we get an id we can use to requery and we let the user change the name of the target at will

There are scenarios when e.g. the user will be able to fetch the magnitudes again and we want to do it with the object used originally

tpolecat:
> There are scenarios when e.g. the user will be able to fetch the magnitudes again and we want to do it with the object used originally

Right, that's my question I guess. Do we need to refetch because the user knows it's been updated in the catalog?

tpolecat:
Do other catalogs actually have fixed ids or are they all insane like Horizons?

cquiroz:
I think a more common scenario is:

the user changes magnitudes, and screws up. Then he decides to refetch to start again

cquiroz:
I'm not very aware of horizons ids but Simbad has stable ids

tpolecat:
> the user changes magnitudes, and screws up. Then he decides to refetch to start again

:+1: that makes sense, although it seems like a potential UX complication "only enable the refresh button if they have changed it"

Anyway seems fine. :shipit:

cquiroz:
I think it will be an always enabled "reset" button that will do a fresh fetch

modules/core/shared/src/main/scala/lucuma/core/enum/MagnitudeBand.scala
```diff
@@ -65,4 +67,12 @@ object MagnitudeBand {
         MagnitudeBand.unsafeFromTag(s)
     }
 
+  /** @group Typeclass Instances */
+  val MagnitudeBandWavelengthOrder: Order[MagnitudeBand] =
+    Order.by(_.center)
+
+  /** @group Typeclass Instances */
+  implicit val MagnitudeBandOrdering: Ordering[MagnitudeBand] =
+    MagnitudeBandEnumerated.toOrdering
```
swalker2m, line 76: I vaguely think with the right imports you get this for free.

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
swalker2m: I'm not sure I understand.  `Eq` is implied by `Order`.

swalker2m (approved):
Looks great. :+1:

cquiroz:
I maybe mistaken here, but I think that for an `Order` to be a proper `Eq` it should take into account all fields. 
Or at least to be consistent with `==`
In this case the order is ignoring the `magnitudes` field as there is not really a sensible way to order two `SortedMap`s

tpolecat:
@cquiroz you are correct. That `Order` instance is not valid.

tpolecat:
I think we may be overdesigning the model here. Do magnitudes really need to be a property of targets, or can they just be something associated with targets instead, that we look up on demand?

tpolecat:
You're right that `Order` needs to consider all fields. Maybe just define `Eq`?

cquiroz:
That's why I removed the implicit form Order and added an Eq instead

PR #180, opened 2020-10-09
change orcid a bit

tpolecat:
ORCID iD is technically a URI but in practice the portion after `https://orcid.org/` (which I call the `value` here) is what we use. So this change adds support for `value` and ensures that you can round-trip through it (as well as through the full URI).


cquiroz (approved):
LGTM

toddburnside (approved):
LGTM.

PR #183, opened 2020-10-13
Add abbreviation to Partner

toddburnside:
I have need in the UI for an abbreviation, as well as a shorter name. While the tag is essentially the abbreviation and I could just uppercase it, that seemed like a conflation of concerns. However, I guess having a bunch of UI concerns in lucuma-core is also a conflation...


cquiroz (approved):
LGTM

PR #184, opened 2020-10-14
Add Eq instance for OrcidProfile and tests

cquiroz:
Add Eq instance for `OrcidProfile`


cquiroz:
@tpolecat Should we use refined types in `OrcIdProfile`?

swalker2m (commented):
Seems good to me.  I'll leave it to @tpolecat to approve though since it's his thing.

tpolecat:
Sorry, I missed this one. @cquiroz what did you want to use refined types for? Did you mean using newtypes for the components so it's not just a bunch of strings?

cquiroz:
Yes basically NonEmptyString in some places, but it is not very important

PR #189, opened 2020-10-20
Epoch format

cquiroz:
This PR adds a `Format` that is more lenient letting you ignore the scheme part and with a different formatting removing the scheme and trailing zeros.

I expect to use this format to let users enter the value on the UI

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/parser/Epoch.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/parser/Epoch.scala
@@ -7,6 +7,7 @@ import cats.syntax.all._
 import atto._, Atto._
 import lucuma.core.math.Epoch
 import lucuma.core.parser.MiscParsers
+import lucuma.core.math.Epoch.Julian
 
 /** Parser for [[lucuma.core.math.Epoch]]. */
 trait EpochParsers {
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/parser/Epoch.scala
```diff
@@ -30,5 +31,15 @@ trait EpochParsers {
       }
       .named("epoch")
 
+  /** Parser for an `Epoch` with a default Julian schemme. */
+  val epochLenient: Parser[Epoch] =
```
tpolecat, line 35: Can we put "Julian" in the name somewhere? `julianEpoch` maybe?

tpolecat (approved):
Looks good, one comment that you may wish to ignore.

swalker2m (approved):
This seems good, pending confirmation from science on the no-scheme concept and what it would mean.  I guess that will fall out of testing though.

modules/core/shared/src/main/scala/lucuma/core/math/parser/Epoch.scala
cquiroz: Sure, is the name used in error messages?

PR #191, opened 2020-10-21
Add ObsStatus enum

rpiaggio:
Also, dial back CI JVM to 11.


swalker2m (approved):
Looks good. :+1:  Do you need a `Display` instance as well?

PR #192, opened 2020-10-21
Shareable Ids

swalker2m:
The various user interfaces and backend services will need to identify top-level objects like targets and observations.  As discussed each kind of identifier should have its own Scala type.  Ideally we can all share the definitions of these identifiers so that they are not repeated.  I'm not sure this is the place for it, but wanted to offer it for comments.

This PR is straightforward, to put it charitably.  Perhaps with `opaque type`s it can be done better in the future?  Or perhaps there is already a preferred `newtype` alternative?  Everything I could think of seemed no better than simply repeating the same definitions each time.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Id.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Id.scala
@@ -1,0 +1,55 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.model
+
+import eu.timepit.refined.auto._
+import eu.timepit.refined.types.numeric.PosLong
+import lucuma.core.util.Gid
+
+/**
+ * Top-level shareable object ids.
+ */
+object Id {
+
+  final case class Asterism(value: PosLong) {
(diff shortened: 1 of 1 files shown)


cquiroz (approved):
LGTM, I guess we don't support a 0 index? Is that why we use PosInt?

tpolecat:
I have been doing these like `User.Id` instead of `Id.User` … not sure which is better but this is inconsistent as it stands.

swalker2m:
> I have been doing these like User.Id instead of Id.User … not sure which is better but this is inconsistent as it stands.

That's what I did too (_typename_`.Id`) in the API service and I like how it reads.  I reversed it here though because otherwise we'd need to add an `Observation` object for example though we aren't anticipating a rich model.  We'd need to add `Id` to the existing `Target` class as well though it is not directly related.

I'd also prefer _typename_.`Id`. How would you do it?

swalker2m:
> LGTM, I guess we don't support a 0 index? Is that why we use PosInt?

I'd be happy with `NonNegLong` if we want to update `Gid`.

cquiroz:
*I'd be happy with NonNegLong if we want to update Gid.*
I don't think there is a need for that, I was just wondering.

swalker2m:
Raul has a better id for this that keeps the natural order (e.g., `Target.Id`).

PR #197, opened 2020-10-26
Add Eq instances for Role and User classes

cquiroz:
Useful for the UI


rpiaggio:
Great, thank you! Do you think we should add `Eq[Role]` too?

rpiaggio (approved):
LGTM

PR #200, opened 2020-10-29
Optional (as in monocle) Ops

swalker2m:
Adds the same syntax that @cquiroz brought over for `Lens` but for `Optional`.  I renamed these to `LensOps` and `OptionalOps` because the editor referred to something in the API codebase and maybe we'll want to add other syntax at some point.


cquiroz (approved):
LGTM

PR #206, opened 2020-11-02
Fixes definition of cms-per-second and use provided physical constants

cquiroz:
This fixes the definition of cms per seconds and also uses the new coulomb's physical constants module.


modules/core/shared/src/main/scala/lucuma/core/math/Constants.scala
```diff
@@ -4,27 +4,49 @@
...
+import lucuma.core.math.units.MetersPerSecond
+import spire.math.Rational
+import coulomb.define.DerivedUnit
+import coulomb.si.Meter
+import coulomb.physicalconstants.infra.PhysicalConstantQuantity
 
 object Constants {
+  object infra {
+    trait AstronomicalUnit
+    implicit val defineAU = {
```
rpiaggio, line 18: Should we explicit the types of implicit vals, or are they too convoluted?

modules/core/shared/src/main/scala/lucuma/core/math/Constants.scala
```diff
@@ -4,27 +4,49 @@
...
+import coulomb.physicalconstants._
+import lucuma.core.math.units.MetersPerSecond
+import spire.math.Rational
+import coulomb.define.DerivedUnit
+import coulomb.si.Meter
+import coulomb.physicalconstants.infra.PhysicalConstantQuantity
 
 object Constants {
+  object infra {
+    trait AstronomicalUnit
```
rpiaggio, line 17: Can we make this a `type` alias so that it doesn't generate a `.class` ?

Otherwise, can it be `sealed` ?

cquiroz: Maybe I'll make object `infra` private, these definitions shouldn't really be available outside, you want to only see the `def`s.

modules/core/shared/src/main/scala/lucuma/core/math/Constants.scala
```diff
@@ -4,27 +4,49 @@
...
 
-  /** Equatorial radius of Earth in meters. */
-  val EquatorialRadiusMeters: Int = 6378137
-
-  /** 1 AU in meters. */
-  val MetersInAU: Double = 1.4959787066e11
+  /** Equatorial radius of Earth. */
+  def EquatorialRadius[V](implicit
+    pcq: PhysicalConstantQuantity[V, EquatorialRadius]
+  ): Quantity[V, pcq.QU] = pcq.q
```
rpiaggio, line 50: Why can't this be `Quantity[Int, Meters]` ?

cquiroz: This is the idiomatic encoding for constants used in coulomb, the idea is you can select the type most appropriate for your particular caller but the units are defined in the constant

modules/core/shared/src/main/scala/lucuma/core/math/Constants.scala
cquiroz: I can replace it with `type` and it compiles fine, `coulomb` uses a `trait` in all these cases so I'm a bit reluctant to change it unless there is a clear benefit

rpiaggio (approved):
LGTM

PR #211, opened 2020-11-05
Rename ProperVelocity

cquiroz:
It was suggested to rename `ProperVelocity` to `ProperMotion` as it maybe clearer this way to science.
I also added a few optics created for `explore`


modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
```diff
@@ -73,24 +73,24 @@ object SiderealTracking extends SiderealTrackingOptics {
...
     * @param epoch           the epoch
-    * @param properVelocity  proper velocity per epoch year
+    * @param properMotion  proper velocity per epoch year
     * @param radialVelocity  radial velocity (km/sec, positive if receding)
     * @param parallax        parallax
     * @param elapsedYears    elapsed time in epoch years
     * @return Coordinates corrected for proper motion
     */
-  def properMotion(
+  def properMotionOn(
```
cquiroz, line 82: I renamed this to avoid a name clash with generated lenses, but maybe a better name would be `coordinatesOn`?

swalker2m: Right.  We're getting `Coordinates` back so maybe `correctedCoordinates` or `adjustedCoordinates`?

swalker2m (approved):
Minor naming question aside, this looks good to me. :+1:

PR #227, opened 2020-11-22
Add missing gnirs enums

cquiroz:
Somehow we missed porting these enums from gsp-core

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enum/GnirsPrism.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enum/GnirsPrism.scala
@@ -1,0 +1,50 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core
+package enum
+
+import cats.instances.string._
+import cats.syntax.eq._
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for GNIRS Prism Turret.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GnirsPrism(
+  val tag: String,
+  val shortName: String,
+  val longName: String
+) extends Product with Serializable {
+  type Self = this.type
+}
+
+object GnirsPrism {
+
+  type Aux[A] = GnirsPrism { type Self = A }
+
+  /** @group Constructors */ case object Mirror extends GnirsPrism("Mirror", "Mirror", "Mirror")
+  /** @group Constructors */ case object Sxd extends GnirsPrism("Sxd", "Short XD", "Short cross dispersion")
+  /** @group Constructors */ case object Lxd extends GnirsPrism("Lxd", "Long XD", "Long cross dispersion")
+
+  /** All members of GnirsPrism, in canonical order. */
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/enum/GnirsPrism.scala
```diff
@@ -0,0 +1,50 @@
...
+package lucuma.core
+package enum
+
+import cats.instances.string._
+import cats.syntax.eq._
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for GNIRS Prism Turret.
+ * @group Enumerations (Generated)
```
swalker2m, line 13: Are these still generated?  I thought we abandoned that idea.

modules/core/shared/src/main/scala/lucuma/core/enum/GnirsPrism.scala
```diff
@@ -0,0 +1,50 @@
...
+  val tag: String,
+  val shortName: String,
+  val longName: String
+) extends Product with Serializable {
+  type Self = this.type
+}
+
+object GnirsPrism {
+
+  type Aux[A] = GnirsPrism { type Self = A }
```
swalker2m, line 25: Remind me what this is about?

modules/core/shared/src/main/scala/lucuma/core/enum/GnirsReadMode.scala
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Enumerated type for GNIRS Read Mode.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GnirsReadMode(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val count: Int,
+  val minimumExposureTime: Int,
```
swalker2m, line 20: Can this be a `Duratation` or something now?

modules/core/shared/src/main/scala/lucuma/core/enum/GnirsReadMode.scala
```diff
@@ -0,0 +1,55 @@
...
+  val readNoise: Int,
+  val readNoiseLow: Int
```
swalker2m, line 22: Should we represent these with their units now?

modules/core/shared/src/main/scala/lucuma/core/enum/GnirsPrism.scala
tpolecat: Not needed. It was from that thing where I was experimenting with indexing thpes on the singleton types of enum values, but it was too complicated.

modules/core/shared/src/main/scala/lucuma/core/enum/GnirsPrism.scala
tpolecat: Not generated anymore. I think we agreed that the Scala enum is truth and database apps need to do a some kind of startup check to ensure consistency (not implemented).

cquiroz:
These were generated ages ago and we copied them to Lucuma core, for some
Reason we didn't copy thes two and we need them in the seqexec 
I copied them only changin the package name

swalker2m (approved):
> These were generated ages ago and we copied them to Lucuma core

I see.  We probably need to go through all of them then at some point and clean them up.

PR #232, opened 2020-11-26
Add catalogId to SiderealTracking Order

swalker2m:
A minor bug fix that adds `catalogId` to `Order[SiderealTracking]` so that two `SiderealTracking` instances that differ only by catalog id count as distinct according to `===`.


cquiroz (approved):
LGTM

PR #238, opened 2020-12-02
Use refined value for Epoch.toMilliYear

toddburnside:
I will add a comment about the changes to SiderealTracking.scala below.


modules/core/shared/src/main/scala/lucuma/core/math/parser/Epoch.scala
```diff
@@ -25,18 +25,18 @@ trait EpochParsers {
...
+    (epochScheme, int.refined[Epoch.Year] <~ char('.'), intN(3))
       .mapN { (s, y, f) =>
-        s.fromMilliyears(y * 1000 + f)
+        s.fromMilliyearsUnsafe(y.value * 1000 + f)
       }
       .named("epoch")
 
   /** Parser for an `Epoch` with a default Julian schemme. */
-  val epochLenient: Parser[Epoch] =
-    (opt(epochScheme), int <~ opt(char('.')), MiscParsers.frac(3))
```
toddburnside, line 36: I removed the optional scheme here because the Format that uses this Parser doesn't include the scheme in the reverseGet. This made the Format unlawful.

modules/core/shared/src/main/scala/lucuma/core/math/Epoch.scala
```diff
@@ -37,13 +44,13 @@ final class Epoch private (val scheme: Epoch.Scheme, val toMilliyears: Int) {
...
   def untilJulianDay(jd: Double): Double =
-    untilEpochYear(scheme.fromJulianDay(jd).epochYear)
+    untilEpochYear(scheme.fromJulianDayToEpochYears(jd))
 
   /** Offset in epoch-years from this `Epoch` to the given epoch year under the same scheme. */
   def untilEpochYear(epochYear: Double): Double =
     epochYear - this.epochYear
 
-  def plusYears(y: Double): Epoch =
+  def plusYears(y: Double): Option[Epoch] =
```
cquiroz, line 53: Would it make sense to make this always take a type positive such that it returns always a valid `Epoch`?

modules/core/shared/src/main/scala/lucuma/core/math/Epoch.scala
```diff
@@ -53,60 +60,75 @@ final class Epoch private (val scheme: Epoch.Scheme, val toMilliyears: Int) {
...
   override def toString: String =
-    Epoch.fromString.taggedToString("Epoch", this)
+    Epoch.fromString.asFormat.taggedToString("Epoch", this)
 
 }
 
 object Epoch extends EpochOptics {
+  type Year         = RefinedInterval.Closed[1900, 3000]
+  type MilliYear    = RefinedInterval.Closed[1900000, 3000999]
+  type IntYear      = Int Refined Year
```
cquiroz, line 73: If the base unit is `IntMilliYear` would it make sens for this to be e.g. `RationalYear`?

modules/core/shared/src/main/scala/lucuma/core/math/parser/Epoch.scala
cquiroz: 👍

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/Epoch.scala
tpolecat: Seems reasonable to me. Not urgent but maybe add a story for this.

tpolecat (approved):
👍

PR #241, opened 2020-12-09
Make Gid regex public. Add tests.

rpiaggio:


modules/core/shared/src/main/scala/lucuma/core/util/Gid.scala
```diff
@@ -47,37 +47,43 @@ final class Gid[A](
...
+    with Show[A]
+    with Encoder[A]
+    with Decoder[A] {
 
   // We use this in a few places
   private final val TagString: String = tag.value.toString
 
+  private final val AnchoredRegex: Regex = raw"$TagString-([1-9a-f][0-9a-f]*)".r
+
+  final val regex: Regex = AnchoredRegex.unanchored
```
swalker2m, line 60: Huh, I didn't realize it was anchored by default.  Why do we need the unanchored version?

swalker2m (approved):
This looks good to me.  :+1:

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/util/Gid.scala
rpiaggio: We are going to pass the regex to `scalajs-react`'s router, which will use it to build a larger regex to parse the URL.

However, I just checked and said router actually needs the regex pattern as a `String`, so I'm going to slightly modify the PR.

PR #242, opened 2020-12-09
Allow testing Format/Split*/Wedge laws without coverage check.

rpiaggio:
The coverage check allows us to make sure we are using the right optic, but it's not a law per se.

This PR keeps the coverage check when testing laws as usual, but also allows testing of just the optic laws without the coverage check. To do the latter, the syntax would be, for example:

`FormatTests(...).formatLaws`, or
`FormatTests(...).formatLawsWith(...)`


cquiroz (approved):
LGTM

swalker2m (approved):
I've forgotten what prompted this change, but it makes sense and looks good. :+1:

PR #244, opened 2020-12-11
Restrict Parallax to [0°, 180°]

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/math/Parallax.scala
```diff
@@ -14,48 +19,48 @@ import spire.std.long._
...
   val mas: Quantity[Rational, MilliArcSecond] = μas.to[Rational, MilliArcSecond]
 
   override def toString: String =
     s"Parallax(${μas.show})"
 }
 
 object Parallax extends ParallaxOptics {
-  val MaxValue: Parallax = apply(648000000000L)
-  val MinValue: Parallax = apply(-648000000000L)
+  type Parallaxμas     = Interval.Closed[0, 648000000000L]
```
tpolecat, line 36: At some point it might be good to define these singletons on `Angle` (i.e., `type Angle180µas = 648000000000L`) so we don't have so many magic numbers floating around. But this is fine for now.

modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
```diff
@@ -197,11 +197,11 @@ object SiderealTracking extends SiderealTrackingOptics {
...
-    order(_.baseCoordinates)  |+|
-      order(_.epoch)          |+|
-      order(_.properMotion)   |+|
+    order(_.baseCoordinates) |+|
+      order(_.epoch) |+|
+      order(_.properMotion) |+|
       order(_.radialVelocity) |+|
-      order(_.parallax)       |+|
+      order(_.parallax) |+|
       order(_.catalogId)
```
tpolecat, line 206: See, this is why code formatters are bad.

tpolecat (approved):
Seems fine to me. If the optics laws pass we're good I think.

modules/core/shared/src/main/scala/lucuma/core/math/Parallax.scala
```diff
@@ -14,48 +19,48 @@ import spire.std.long._
...
   * Normally parallax is expressed in milliarcseconds but simbad reports them
-  * with a higher precision thus using microarcseconds gives us enough space
-  * Absolute allowed parallax values need to be less or equal thán 180 degrees
+  * with a higher precision thus using microarcseconds gives us enough space.
+  * Parallax values need to be in the interval [0°, 180°].
+  * 180° is a theoretical limit - actual astronomical values will be very small.
   */
-sealed abstract case class Parallax protected (μas: Quantity[Long, MicroArcSecond]) {
+sealed abstract case class Parallax protected (
+  μas: Quantity[Parallax.LongParallaxμas, MicroArcSecond]
```
cquiroz, line 27: This will break `lucuma-catalog` but that's fine. We should cap the negative numbers there anyway

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/Parallax.scala
toddburnside: I'm going to modify the decoder in Explore to coerce the negative numbers we receive there.

PR #246, opened 2020-12-11
microns

swalker2m:
Adds micron conversion support in `Wavelength`.  Most instruments (GMOS notwithstanding) reference wavelengths in microns so I think we need support for it.


cquiroz (approved):
LGTM

PR #249, opened 2020-12-14
Create types Angle180µas and Angle360µas

toddburnside:
Based on a comment by Rob in PR #244 about creating types to eliminate some Magic Numbers when creating refined types.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
@@ -12,15 +12,15 @@ import lucuma.core.optics._
 import monocle.{ Iso, Prism }
 
 /**
-  * Exact angles represented as integral microarcseconds. These values form a commutative group over
-  * addition, where the inverse is reflection around the 0-180° axis. The subgroup of angles where
-  * integral microarcseconds correspond with clock microseconds (i.e., where they are evenly
-  * divisible by 15 microarcseconds) is represented by the HourAgle subtype.
-  *
-  * Lawful conversion to and from other types/scales is provided by optics defined on the companion
-  * object. Floating-point conversions are provided directly
-  * @param toMicroarcseconds This angle in microarcseconds. Exact.
-  */
+ * Exact angles represented as integral microarcseconds. These values form a commutative group over
+ * addition, where the inverse is reflection around the 0-180° axis. The subgroup of angles where
+ * integral microarcseconds correspond with clock microseconds (i.e., where they are evenly
+ * divisible by 15 microarcseconds) is represented by the HourAgle subtype.
+ *
+ * Lawful conversion to and from other types/scales is provided by optics defined on the companion
+ * object. Floating-point conversions are provided directly
+ * @param toMicroarcseconds This angle in microarcseconds. Exact.
+ */
 sealed class Angle protected (val toMicroarcseconds: Long) {
 
   // Sanity checks … should be correct via the companion constructor.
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
```diff
@@ -153,13 +151,22 @@ sealed class Angle protected (val toMicroarcseconds: Long) {
...
-   * angle difference or separation between two angles in the range [0 .. π]
-   */
+    * angle difference or separation between two angles in the range [0 .. π]
+    */
   def difference(a: Angle): Angle =
     Angle.difference(this, a)
 }
 
 object Angle extends AngleOptics {
+  type Angle180µas = 648000000000L  // 180 * µasPerDegree
```
cquiroz, line 161: Why a type?

toddburnside: It is used in `type Parallaxμas = Interval.Closed[0, Angle.Angle180µas]` in Parallax, where it has to be a type.

cquiroz: Oh neat

rpiaggio:
Just a note about formatting. There's an agreement to use Java style ScalaDoc blocks, with asterisks aligned on the first of the two asterisk columns. This can be achieved with `docstrings.style = Asterisk` in scalafmt.

rpiaggio (approved):
LGTM

cquiroz (approved):

toddburnside:
> Just a note about formatting. There's an agreement to use Java style ScalaDoc blocks, with asterisks aligned on the first of the two asterisk columns. This can be achieved with `docstrings.style = Asterisk` in scalafmt.

I didn't know what the agreement was. I just assumed it would get formatted the agreed upon way automagically. But, now I see that the `.scalafmt.conf` file isn't checked in. I'm wondering know how I got the one I have... Anyway, I'll make that change and reformat. Should we have a definitive `.scalafmt.conf` in git?

rpiaggio:
> I didn't know what the agreement was. I just assumed it would get formatted the agreed upon way automagically. But, now I see that the `.scalafmt.conf` file isn't checked in. I'm wondering know how I got the one I have... Anyway, I'll make that change and reformat. Should we have a definitive `.scalafmt.conf` in git?

Well, there isn't an agreement on using autoformatting. Some of us use it, some don't. That's why it's `.gitignore`d here. Personally, I copy over the one we have in Explore.

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
tpolecat: Thinking about it a bit more, it might make more sense to say `final val Angle180µas = 648000000000L` and then use `Interval.Closed[0, Angle.Angle180µas.type]` so we have both the type and term available.

toddburnside: We do have both available, because I create `µasPer180` using `valueOf[Angle180µas]` a few lines down. But, it would be terser to use `.type` where needed. I can switch to that. Is the `final` needed?

tpolecat: Your solution with `valueOf` is fine, I don't feel strongly. 

But yes, `final` is required to get a singleton type.

```scala
scala> object Foo { val x = 123 }
object Foo

scala> val y: Foo.x.type = 42
                           ^
       error: type mismatch;
        found   : Int(42)
        required: Foo.x.type

scala> val y: Foo.x.type = 123
                           ^
       error: type mismatch;
        found   : Int(123)
        required: Foo.x.type

scala> object Foo { final val x = 123 } 
object Foo

scala> val y: Foo.x.type = 42
                           ^
       error: type mismatch;
        found   : Int(42)
        required: 123

scala> val y: Foo.x.type = 123 
val y: 123 = 123
```

toddburnside: Interesting. I can do `Interval.Closed[0, Angle.µasPer180.type]` without µasPer180 being final and it seems to work fine. 🤔 I was a bit surprised that I could also use `val µasPer180 = 180L * µasPerDegree` and that worked.

tpolecat: I think you won't be able to construct a value of that type with that bound unless it's final. Kind of error-prone, I think I like the `valueOf` strategy better now.

toddburnside: OK. I will leave it as it is.

toddburnside: Just one last bit that is interesting (to me). With your first version of Foo with the non-final val, I can't do `val xx: Foo.x.type = 123`, but I CAN do `val xx: Int = valueOf[Foo.x.type]`.

PR #254, opened 2020-12-23
Fix ArbGid capitalization

swalker2m:
A minuscule change, but it something that I've wasted time on twice now.  `ArbGid` containing `ArbGid` makes it so that 

```
import lucuma.core.util.arb.ArbGid
```

followed by

```
import ArbGid._
```

prevents the `implicit def ArbGid` definition from being seen by the implicit lookup.


toddburnside (approved):
LGTM

PR #263, opened 2021-01-15
SetterOps

swalker2m:
Brings the `:=` `State` syntax to Monocle's `Setter`.  (Also moves the existing optics test cases to the corresponding package `lucuma.core.optics` instead of `lucuma.core.math.optics`.)


cquiroz (approved):
LGTM and thanks. I can use this in a few places

PR #272, opened 2021-02-08
Add Enums for ConstraintSet.

toddburnside:
[ch-695]


swalker2m (commented):
I see now what you mean by putting this in "the model".  I was only thinking you were referring only to the throwaway model in the tmp API server.  I think this is great, but there was considerable debate about putting even the `Target` model in `core` at one point. I assume the case for/against constraints going here is similar though.  Particularly at the very least all the enums and basic types seem like they belong to me.  What do other people think?

toddburnside:
> but there was considerable debate about putting even the `Target` model in `core` at one point

Hmmm. Now that you mention it, I seem to remember that debate. It being due to the fact that the UI may not request all the fields for an entity, and will need to create it's own types, anyway. 

I could move the ConstraintSet to lucuma-tmp-api if people think that is the best approach. Not sure about ElevationRange, AirmassRange and HourAngleRange.

toddburnside:
> Not sure about ElevationRange, AirmassRange and HourAngleRange.

Actually, `AirmassRange` and `HourAngleRange` have `setMin` and `setMax` methods that are mostly appropriate for the UI. In the API, optics could be used instead. In fact, in the API, maybe the `apply` methods should return an None if min > max instead of swapping the order. This might be an argument for moving the "Ranges" to lucuma-tmp-api and creating what is needed in the UI when it comes time. If so, should we keep the refined units in lucuma-core to keep them consistent?

modules/core/shared/src/main/scala/lucuma/core/enum/ImageQuality.scala
```diff
@@ -0,0 +1,38 @@
+// Copyright (c) 2016-2020 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enum
+
+import lucuma.core.util.{ Display, Enumerated }
+
+sealed abstract class ImageQuality(val toDeciArcSeconds: Int) extends Product with Serializable {
+  def toArcSeconds: Double = toDeciArcSeconds / 10.0
```
swalker2m, line 9: If we need it at some point we could add a `toAngle: Angle` option.

swalker2m (approved):
This is all looks great to me.  I raised the question about the model and called the initial more complete version into question, but it is a bit sad to see the rest go.  I think it is the consensus to make the `core` pretty light but I can't help but think this will inevitably lead to duplication.  At any rate, :+1: for this as it stands.

toddburnside:
> This is all looks great to me. I raised the question about the model and called the initial more complete version into question, but it is a bit sad to see the rest go. I think it is the consensus to make the `core` pretty light but I can't help but think this will inevitably lead to duplication. At any rate, 👍 for this as it stands.

The rest of it is teed up to go into the api once I release this. The one piece of duplication I'd like to avoid is the refined types for AirmassRange and HourAngleRange. As far as I can see, we don't really have a good place to put those. One option would be to put them in the ConstraintSet object that defines the ConstraintSet.Id. I'll probably revisit that before I move on to the UI side of things.

PR #285, opened 2021-02-19
Update to Monocle 3

cquiroz:
An experimental release of Monocle 3 was released today and I thought to take it for a drive.

We don't often shun experimental release so maybe we can start using it. It needs the whole infrastructure to be ported tough

The state module is deprecated so I internalized the code


modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
```diff
@@ -12,125 +12,125 @@ import lucuma.core.optics._
...
+  * integral microarcseconds correspond with clock microseconds (i.e., where they are evenly
+  * divisible by 15 microarcseconds) is represented by the HourAgle subtype.
+  *
  * Lawful conversion to and from other types/scales is provided by optics defined on the companion
- * object. Floating-point conversions are provided directly
- * @param toMicroarcseconds This angle in microarcseconds. Exact.
- */
+  * object. Floating-point conversions are provided directly
+  * @param toMicroarcseconds This angle in microarcseconds. Exact.
+  */
```
swalker2m, line 23: Is this a scalafmt thing?

modules/core/shared/src/main/scala/lucuma/core/model/EphemerisCoordinates.scala
```diff
@@ -17,17 +17,17 @@ import monocle.macros.GenLens
...
   /** Interpolates the position and rate of change at a point between this
     * coordinate and the given coordinate.
     */
   def interpolate(that: EphemerisCoordinates, f: Double): EphemerisCoordinates = {
     def interpolateAngle(a: Angle, b: Angle): Angle =
       Angle.fromMicroarcseconds(
-        (Angle.signedMicroarcseconds.get(a).toDouble * (1 - f) + Angle.signedMicroarcseconds.get(b) * f).round
+        (Angle.signedMicroarcseconds.get(a).toDouble * (1 - f) + Angle.signedMicroarcseconds.get(
+          b
+        ) * f).round
```
swalker2m, line 30: scalafmt?

swalker2m (approved):
This looks fine to me.  The change to `andThen` seems like an improvement.  I never liked the Monocle compose symbols.

Let's see if @tpolecat, @rpiaggio, and/or @toddburnside have any input, especially since it isn't officially released yet.

Do you know why the state stuff was deprecated?  I've found it useful.

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
cquiroz: Yes it is

cquiroz:
`State` was deprecated mostly to reduce the amount of things to maintain. Julien mentioned he has never heard anybody using it until I mentioned we use it

cquiroz:
I'll try to get a similar PR for `scalajs-react` as it would be a showstopper if we want to take this into use

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
rpiaggio: I think this should revert if you use the `scalafmt` we have commited in Explore and other projects.

rpiaggio (approved):
LGTM. In general I like the new syntax. I'll miss the `@Lenses` macro though.

cquiroz:
I miss it to, there maybe an alternative upcoming that will generate lenses for all case class fields, but there is no ETA

Also note that `GenIso`, `GenPrism` are still present but on dotty `Focus` can do those too

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
cquiroz: The conf we have in `lucuma-core` is identical to `explore`

toddburnside (approved):
Seems good to me. How much of this was automated by the scalafix? Could it handle replacing the @Lens annotation?

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
rpiaggio: Ok. We don't have `scalafmt.conf` committed here in `lucuma-core` (it's actually in `.gitignore`) so each of us has their own version. So, just checking since yours might have been different.

cquiroz:
The scalafix isn't ready so not included in this PR

cquiroz:
https://github.com/japgolly/scalajs-react/pull/862

tpolecat:
@cquiroz would it be possible to make a new branch and cherry-pick the non-formatting/header changes into it? It would be nice for this PR to be focused specifically on the Monocle changes, since it seems likely we'll want to look back and see what changed if we run into trouble.

cquiroz:
Right I could do that and squash it as we have code for 2 monocle versions

cquiroz:
See #290

PR #291, opened 2021-03-05
Allow pattern-match parsing for Ids.

rpiaggio:
When interfacing with JS libraries, in particular the drag'n'drop library we use, ids are passed back and forth as strings and there are situations where we have to figure out the type of objects based on their stringified `Id`.

The `unapply` method comes very handy for this, allowing us to do things like:

``` scala
string match {
  case Target.Id(targetId) => ... /* targetId is a Target.Id here */
  case Asterism.Id(asterismId) => ... /* asterismId is an Asterism.Id here */
  ...
}
```


toddburnside (approved):
Nice

PR #302, opened 2021-03-29
remove cats-effect

tpolecat:
This removes core's dependency on Cats-Effect, which frees up downstream projects to upgrade to CE3 at their leisure (rather than forcing the upgrade). It was only used for a few convenience methods that needed to look at the clock, and these are easy enough to do inline when needed.


swalker2m (approved):
Makes sense to me.  :+1:

cquiroz (approved):
This is great, somehow I always thought `core` shouldn't depend on CE as it was mostly about models and math

PR #323, opened 2021-04-30
WithId syntax. Enum shortNames " for arcsec.

rpiaggio:
* Generally use `"` for arcsecs in enum's `shortName`.
* `WithId` is now a class which takes the `idTag` as parameter, for easier definition.
* Add `Atom.Id` and `Step.Id`.
* Avoid warnings in `build.sbt`.


cquiroz (approved):
Looks Good

PR #335, opened 2021-05-19
Add filter type property

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.m.quiroz@gmail.com>

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enum/FilterType.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enum/FilterType.scala
@@ -1,0 +1,48 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma
+package core
+package enum
+
+import cats.syntax.eq._
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for filter types.
+ * @group Enumerations (Generated)
+ */
+sealed abstract class FilterType(
+  val tag: String
+) extends Product with Serializable
+
+object FilterType {
+
+  /** @group Constructors */ case object BroadBand extends FilterType("BroadBand")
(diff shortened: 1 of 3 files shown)


rpiaggio (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enum/FilterType.scala
```diff
@@ -0,0 +1,48 @@
...
+  val tag: String
+) extends Product with Serializable
+
+object FilterType {
+
+  /** @group Constructors */ case object BroadBand extends FilterType("BroadBand")
+  /** @group Constructors */ case object NarrowBand extends FilterType("NarrowBand")
+  /** @group Constructors */ case object Combination extends FilterType("Combination")
+  /** @group Constructors */ case object Spectroscopic extends FilterType("Spectroscopic")
+  /** @group Constructors */ case object Engineering extends FilterType("Engineering")
```
swalker2m, line 25: How will these be used?

I wonder if the combination filters should provide a way to get the component values.  If they are a combination of other filters perhaps we should provide a set of `FilterType` or if they are always the same use the underlying type?

swalker2m: From the point of view of the ICTD / Resource, I'm guessing you can't have a combination without having the constituent filters and if you have the constituent filters then you have the combination available.

swalker2m (approved):
Looks good.  :+1:

modules/core/shared/src/main/scala/lucuma/core/enum/FilterType.scala
cquiroz: So this comes essentially because on the UI we want to group filters, on three main categories broadband/narrowband and combinations. I wish I could classify them based on eg the wavelength but there isn't a direct correlation thus it gets classified based on data table

We could create a combination filter that contains two of them but not every combination is allowed,
Nor you can deduce the properties out of the properties of the combined filters

PR #348, opened 2021-06-22
Add inactive obs statuses

swalker2m:
From the inception document:

> There is a pull-down menu that shows the current status, and with appropriate permissions, allows changing the status.  Each observation may also be separately marked as either "Active'' or "Inactive'' (the vertical toggle switch on the right side of the badge) which allows PIs to prevent or halt execution of "Ready'' or "Ongoing'' observations while retaining their status information.

The description seems to make the active vs. inactive state apply only to `Ready` and `Ongoing` observations status options.  If that's the case, we should add two new observation statuses to reflect whether the observation is active when otherwise in `Ready` or `Ongoing`.  That's what this PR would accomplish.

That said, I'm not sure if this is the intent.  Should it be possible to mark an observation `Inactive` regardless of its observation status?  For example, a `ForReview` but `Inactive` observation or an `Observed` and yet `Active` observation?

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enum/ObsStatus.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enum/ObsStatus.scala
@@ -11,18 +11,32 @@ import lucuma.core.util.Display
 sealed abstract class ObsStatus(val label: String) extends Product with Serializable
 
 object ObsStatus {
-  case object New       extends ObsStatus("New")
-  case object Included  extends ObsStatus("Included")
-  case object Proposed  extends ObsStatus("Proposed")
-  case object Approved  extends ObsStatus("Approved")
-  case object ForReview extends ObsStatus("For Review")
-  case object Ready     extends ObsStatus("Ready")
-  case object Ongoing   extends ObsStatus("Ongoing")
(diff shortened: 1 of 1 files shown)


cquiroz:
Should the transitions be restricted, only allowing some changes e.g. via a state machine?

rpiaggio:
It would definitely be desirable to:
* Populate the dropdown selector from all the enum values (which wouldn't be the case here).
* Make it impossible to model illegal states (which is the case here).

If not all transitions are allowed, I'm wondering if a dropdown selector is the right UI element for this. For example: can an `InactiveOngoing` observation be set to `Proposed`? Or a `New` directly to `Ready`? A FSM makes a lot of sense to prevent this, but how would we model it in the API?

andrewwstephens:
Setting an observation to "Inactive" might also set it's planned time to zero.  This would allow PIs to control the planned time of their program by deactivating observations (instead of having to delete them).  In this scenario it might be useful if the active flag could be set for all status values.

swalker2m:
>  In this scenario it might be useful if the active flag could be set for all status values.

I'll close this then in favor of a separate status flag.

PR #349, opened 2021-06-23
ObsActiveStatus

swalker2m:
A second attempt to satisfy the model requirements of this part of the inception document:

> There is a pull-down menu that shows the current status, and with appropriate permissions, allows changing the status. Each observation may also be separately marked as either "Active'' or "Inactive'' (the vertical toggle switch on the right side of the badge) which allows PIs to prevent or halt execution of "Ready'' or "Ongoing'' observations while retaining their status information.

Adds an `ObsActiveStatus`, for want of a better name.


cquiroz (approved):
LGTM

PR #350, opened 2021-06-23
Add Scorpio/Zorro/Alopeke

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.m.quiroz@gmail.com>


modules/core/shared/src/main/scala/lucuma/core/enum/Instrument.scala
```diff
@@ -56,7 +59,10 @@ object Instrument {
       Niri,
       Phoenix,
       Trecs,
-      Visitor
+      Visitor,
+      Scorpio,
+      Alopeke,
+      Zorro
```
swalker2m, line 65: It isn't clear in my head when to add a visitor instrument to the list and when to just let it be `Visitor`.  Seems like we should add all of them as they come up and remove `Visitor` or else remove the specific visitor instances?

swalker2m (approved):
Looks good to me.

modules/core/shared/src/main/scala/lucuma/core/enum/Instrument.scala
cquiroz: It seems zorro and alopeke are kind of permanent? They show up on the PIT but not in the OT nor the seqexec right?

PR #365, opened 2021-07-27
Remove ConstraintSet.Id

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/model/ids.scala
```diff
@@ -8,7 +8,6 @@ import eu.timepit.refined.auto._
 object Asterism       extends WithId('a')
 object Atom           extends WithId('m')
 object Configuration  extends WithId('x')
-object ConstraintSet  extends WithId('c')
```
toddburnside, line 11: I guess we could give `c` to Configuration if we wanted.

swalker2m: Bonus!

swalker2m (approved):
Looks good to me. :+1:

PR #368, opened 2021-08-02
Adopt Spire's Interval and IntervalSeq

rpiaggio:
Remove our `Interval` and `Schedule` classes in favor of `spire`'s `Interval[Instant]` and `IntervalSeq[Instant]`.

Actually, `Bounded[Instant]` is used mostly everywhere. `Bounded <: Interval` and represents an `Interval` with defined bounds on both ends (as opposed to going to infinity in either end).

Convenience methods that we use frequently were kept as extensions as well as some optics.


build.sbt
```diff
@@ -50,6 +50,7 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
       "com.manyangled"    %%% "coulomb-refined"            % coulombVersion,
       "com.manyangled"    %%% "coulomb-physical-constants" % coulombVersion,
       "org.typelevel"     %%% "spire"                      % spireVersion,
+      "org.typelevel"     %%% "spire-extras"               % spireVersion,
```
cquiroz, line 53: Oh more stuff for the port!

modules/core/shared/src/main/scala/lucuma/core/instances/BoundedIntervalInstances.scala
```diff
@@ -0,0 +1,18 @@
...
+package lucuma.core.instances
+
+import cats.Eq
+import spire.math.Bounded
+import spire.math.Interval
+
+trait BoundedIntervalInstances {
+  implicit def BoundedIntervalEq[A: Eq]: Eq[Bounded[A]] =
+    new Eq[Bounded[A]] {
+      def eqv(x: Bounded[A], y: Bounded[A]): Boolean = Interval.eq[A].eqv(x, y)
```
cquiroz, line 13: Odd this wasn't defined in `spire` itself

modules/core/shared/src/main/scala/lucuma/core/syntax/BoundedInterval.scala
```diff
@@ -0,0 +1,47 @@
...
+package lucuma.core.syntax
+
+import cats.Order
+import cats.syntax.all._
+import spire.math.Bounded
+import spire.math.Interval
+final class BoundedIntervalOps[A](val self: Bounded[A]) extends AnyVal {
+
+  /** True if this and the other interval abut each other. */
+  def abuts(other: Bounded[A])(implicit order: Order[A]): Boolean =
```
cquiroz, line 15: `abuts`?

cquiroz (approved):
Looks great, I didn't look at the tests in detail though

modules/core/shared/src/main/scala/lucuma/core/syntax/BoundedInterval.scala
rpiaggio: This was present in our original `Interval`. I think I learned the meaning of "to abut" there.

modules/core/shared/src/main/scala/lucuma/core/instances/BoundedIntervalInstances.scala
rpiaggio: There's `Eq[Interval]` but not for its subclasses (and `Eq`'s type parameter is invariant).

modules/core/shared/src/main/scala/lucuma/core/math/skycalc/solver/Solver.scala
```diff
@@ -19,15 +20,15 @@ object SolverStrategy {
 }
 
 /**
- * Typeclass defininig an algorithm that finds a [[Schedule]] within a given [[Interval]] for which a given
- * function <code>Instant => Boolean</code> is true.
+ * Typeclass defininig an algorithm that finds an `IntervalSeq[Instant]` within a
+ * given `Bounded[Interval]` for which a given function <code>Instant => Boolean</code> is true.
```
toddburnside, line 24: Bounded[Instant]?

rpiaggio: Yup, thank you!

PR #372, opened 2021-08-03
Iterator for Samples. Fix Scaladoc.

rpiaggio:


swalker2m (approved):
Seems reasonable to me. :+1:

PR #379, opened 2021-08-16
Update SpatialProfile model

cquiroz:
The current `SpatialProfileType` enum is not enough we need a full model to support an angle for `GaussianSource`


swalker2m (approved):
Looks great. :+1:

PR #384, opened 2021-08-31
Add extra magnitude system and units

cquiroz:
ITC uses a couple of extra magnitude systems. We may want to let users set magnitudes with them in the UI too

I learned mag systems may have units and I thought to include them in the encoding


swalker2m:
> We may want to let users set magnitudes with them in the UI too

Seems like you'll have to at least display them since they now can appear in the model.

swalker2m (approved):
I'm not sure what we'll do with this but it seems good to have. :+1:

cquiroz:
Yes they will appear as extra options here:
image(3)

andrewwstephens:
:+1: Definitely useful!

PR #386, opened 2021-09-01
Make the width contain units and add start/end functions

cquiroz:
A Band's width has no units. It should be nanometers

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enum/MagnitudeBand.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enum/MagnitudeBand.scala
@@ -7,8 +7,16 @@ package enum
 
 import cats.Order
 import cats.syntax.eq._
+import coulomb._
+import coulomb.refined._
+import eu.timepit.refined.auto._
+import eu.timepit.refined.numeric.Positive
+import eu.timepit.refined.types.numeric.PosInt
 import lucuma.core.math.Wavelength
+import lucuma.core.math.units.Nanometer
 import lucuma.core.util.Enumerated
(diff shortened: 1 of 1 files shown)


rpiaggio (approved):
LGTM

swalker2m (approved):
Seems good to me.  :+1:  Can't believe we had a loose `Int` in our model like that!

cquiroz:
Yeah. I was a bit surprised too

modules/core/shared/src/main/scala/lucuma/core/enum/MagnitudeBand.scala
```diff
@@ -19,32 +27,40 @@ sealed abstract class MagnitudeBand(
...
   val longName: String,
   val center: Wavelength,
-  val width: Int,
+  val width: Quantity[PosInt, Nanometer],
   val magnitudeSystem: MagnitudeSystem
-) extends Product with Serializable
+) extends Product with Serializable {
+  require(center.toPicometers.value >= width.value / 2)
+
+  def end: Wavelength = Wavelength(center.toPicometers + (width.value.value / 2).withRefinedUnit[Positive, Nanometer])
```
toddburnside, line 35: Not a big deal here, but I _think_ you can avoid stripping the units off width for the division and then adding them back on by `center.toPicometers + width / 2.withRefinedUnit[Positive, Unitless]`.

cquiroz: yes you can see
https://github.com/erikerlandson/coulomb/issues/211

PR #388, opened 2021-09-02
Use units for gmos detector fields and fix some values

cquiroz:
I got a bit confused with some values here as they don't have units and now I marked them as pixels. This is not required yet but it serves as documentation and would help on unit analysis eventually
Also the values were a bit off in some cases


modules/core/shared/src/main/scala/lucuma/core/enum/GmosDetector.scala
```diff
@@ -16,18 +23,35 @@ sealed abstract class GmosDetector(
...
-  val xSize: Int,
-  val ySize: Int,
-  val maxRois: Int
-) extends Product with Serializable
+  val shuffleOffset: Quantity[PosInt, Pixels],
+  val maxRois: PosInt
+) extends Product with Serializable {
+  def xSize(site: Site): Quantity[PosInt, Pixels]
+  def ySize(site: Site): Quantity[PosInt, Pixels]
+  def pixelSize(site: Site): Angle
```
rpiaggio, line 31: Would it make sense to group these 3 values into another structure? (Maybe not, I don't know the model that well)

cquiroz: Maybe it would be simpler to have three instances `G[N|S]Hammamtsu` @swalker2m WDYT?

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -19,6 +19,9 @@ import eu.timepit.refined.types.numeric.PosInt
 import spire.math._
 
 trait units {
+  trait Pixels
```
cquiroz, line 22: This maybe odd as a unit but I think it is correct. At least it would let us clearly mark what values are in pixels

cquiroz:
I reworked this to have 4 instances of the enum with a detector per site instrument rather than functions that depend on the site

rpiaggio:
I was going to suggest this at some point, but shouldn't we change the ODB API accordingly?

cquiroz:
I don't think this is exposed on the API yet

modules/core/shared/src/main/scala/lucuma/core/enum/GmosDetector.scala
```diff
@@ -16,22 +23,60 @@ sealed abstract class GmosDetector(
...
+                               "GmosSE2V",
+                               "E2V",
+                               "E2V",
+                               Angle.fromMicroarcseconds(73000),
+                               1536.withRefinedUnit[Positive, Pixels],
+                               6144.withRefinedUnit[Positive, Pixels],
+                               4608.withRefinedUnit[Positive, Pixels],
+                               4
+                             )
+  /** @group Constructors */ case object GmosNHammamtsu extends GmosDetector(
```
swalker2m, line 56: Typo here, I think.  "Hamamatsu" should be instead of "Hammamtsu"?

modules/core/shared/src/main/scala/lucuma/core/enum/GmosDetector.scala
```diff
@@ -16,22 +23,60 @@ sealed abstract class GmosDetector(
...
+                               "GmosNHAMAMATSU",
```
swalker2m, line 57: I think this will turn into "GMOSNHAMAMATSU" in the GraphQL enum.  GMOS is a given maybe we should make it [NS]HAMAMATSU and [NS]E2V?  I don't know.

swalker2m (approved):
I think this is a good improvement.  Hmm, don't the values get used in the drawing code?

The only potential drawback that occurs to me is that it becomes possible to mix a `GmosNHamamatsu` in a GMOS South observation.  We split up filters, FPUs, etc. maybe we need two separate enums altogether like we did for those.

cquiroz:
I hope we use the values for drawing. There was a discrepancy on the values in the current code and the real ones that should reflect on the drawing.
I'll split this into GN and GS detectors

swalker2m (approved):
This seems like the right approach.  Looks good. :+1:

rpiaggio:
> I don't think this is exposed on the API yet

https://github.com/gemini-hlsw/explore/blob/master/common-graphql/src/main/resources/graphql/schemas/ObservationDB.graphql#L1139

But it seems it can be easily migrated to `Gmos{North,South}Detector`, which I agree is the way to go.

cquiroz:
I guess we don't use the API yet

PR #391, opened 2021-09-07
Rename Ifu enums to avoid confusions

cquiroz:
For ages the `Ifu1` enum has been associated with `IFU-2` which is IMHO confusing.
I'm proposing we rename the `Ifu` enums to something more appropriate.

I fear this change my cause the telescope to implode but it seems we should do it.

Worth checking its effect in observe/seqexec @jluhrs

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthFpu.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthFpu.scala
@@ -35,13 +35,13 @@ object GmosNorthFpu {
   /** @group Constructors */ case object LongSlit_1_50 extends GmosNorthFpu("LongSlit_1_50", "1.5\"", "Longslit 1.50 arcsec", Some(Angle.fromDoubleArcseconds(1.50)), Angle.fromDoubleArcseconds(0.000))
   /** @group Constructors */ case object LongSlit_2_00 extends GmosNorthFpu("LongSlit_2_00", "2.0\"", "Longslit 2.00 arcsec", Some(Angle.fromDoubleArcseconds(2.00)), Angle.fromDoubleArcseconds(0.000))
   /** @group Constructors */ case object LongSlit_5_00 extends GmosNorthFpu("LongSlit_5_00", "5.0\"", "Longslit 5.00 arcsec", Some(Angle.fromDoubleArcseconds(5.00)), Angle.fromDoubleArcseconds(0.000))
-  /** @group Constructors */ case object Ifu1 extends GmosNorthFpu("Ifu1", "IFU-2", "IFU 2 Slits", Option.empty[Angle], Angle.fromDoubleArcseconds(33.500))
-  /** @group Constructors */ case object Ifu2 extends GmosNorthFpu("Ifu2", "IFU-B", "IFU Left Slit (blue)", Option.empty[Angle], Angle.fromDoubleArcseconds(31.750))
-  /** @group Constructors */ case object Ifu3 extends GmosNorthFpu("Ifu3", "IFU-R", "IFU Right Slit (red)", Option.empty[Angle], Angle.fromDoubleArcseconds(35.250))
+  /** @group Constructors */ case object Ifu2 extends GmosNorthFpu("Ifu2", "IFU-2", "IFU 2 Slits", Option.empty[Angle], Angle.fromDoubleArcseconds(33.500))
+  /** @group Constructors */ case object IfuBlue extends GmosNorthFpu("IfuBlue", "IFU-B", "IFU Left Slit (blue)", Option.empty[Angle], Angle.fromDoubleArcseconds(31.750))
+  /** @group Constructors */ case object IfuRed extends GmosNorthFpu("IfuRed", "IFU-R", "IFU Right Slit (red)", Option.empty[Angle], Angle.fromDoubleArcseconds(35.250))
 
   /** All members of GmosNorthFpu, in canonical order. */
   val all: List[GmosNorthFpu] =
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthFpu.scala
```diff
@@ -35,13 +35,13 @@ object GmosNorthFpu {
...
   /** @group Constructors */ case object LongSlit_2_00 extends GmosNorthFpu("LongSlit_2_00", "2.0\"", "Longslit 2.00 arcsec", Some(Angle.fromDoubleArcseconds(2.00)), Angle.fromDoubleArcseconds(0.000))
   /** @group Constructors */ case object LongSlit_5_00 extends GmosNorthFpu("LongSlit_5_00", "5.0\"", "Longslit 5.00 arcsec", Some(Angle.fromDoubleArcseconds(5.00)), Angle.fromDoubleArcseconds(0.000))
-  /** @group Constructors */ case object Ifu1 extends GmosNorthFpu("Ifu1", "IFU-2", "IFU 2 Slits", Option.empty[Angle], Angle.fromDoubleArcseconds(33.500))
-  /** @group Constructors */ case object Ifu2 extends GmosNorthFpu("Ifu2", "IFU-B", "IFU Left Slit (blue)", Option.empty[Angle], Angle.fromDoubleArcseconds(31.750))
-  /** @group Constructors */ case object Ifu3 extends GmosNorthFpu("Ifu3", "IFU-R", "IFU Right Slit (red)", Option.empty[Angle], Angle.fromDoubleArcseconds(35.250))
+  /** @group Constructors */ case object Ifu2 extends GmosNorthFpu("Ifu2", "IFU-2", "IFU 2 Slits", Option.empty[Angle], Angle.fromDoubleArcseconds(33.500))
```
swalker2m, line 38: We might go with `Ifu2Slits` as Rob suggested because `Ifu2` was the old name for `IfuBlue`?

swalker2m (commented):
This is an obvious improvement in readability.  I'm a bit curious about XML migration though.

https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/InstGmosCommon.java#L1523

When importing a pre-2022A program if the name is `"Ifu2"` then I'm worried it won't be converted to `IfuBlue`, etc.  Usually in these cases we have to add a migration where we manipulate the actual XML.  For example, see

https://github.com/gemini-hlsw/ocs/blob/494f3b60f5aefca0947424bd7df606cb6cea1b11/bundle/edu.gemini.spModel.io/src/main/scala/edu/gemini/spModel/io/impl/migration/to2018A/To2018A.scala#L15

Is it not necessary here for some reason?

modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthFpu.scala
cquiroz: Sounds good, `Ifu2Sllits` seems good

cquiroz:
Thanks for mentioning the xmls. My main worry about this change is exactly the impacts downstream. e.g. I know the conf matrix uses the old Ifu names and there maybe many other cases.
Also in terms of the seqexec/observe. things like ifu2 are probably even encoded in epics channel.

My hope was that this change would only apply to GPP but didn't consider we may want to migrate from the old db. Is that going to be even possible, the models are quite different

swalker2m:
> My hope was that this change would only apply to GPP but didn't consider we may want to migrate from the old db. Is that going to be even possible, the models are quite different

My expectation is that we will migrate programs from `ocs` to `lucuma`.  Each will be as if explicitly configured instead of relying on sequence generation.

The main worry with migration wrt this change though is between semesters.  When we switch to a new semester in December we'll read the exported XMLs from 2021B and there will be programs that write the old values of these FPU options in to the program XML that we'll then read with the 2022A version.

swalker2m:
> The main worry with migration wrt this change though is between semesters.

That's not to say the rename is a bad idea or that it shouldn't be done.  I think though that we need to write a migration of the XML like we've had to do in the past.

cquiroz:
ocs doesn't depend on lucuma-core so I guess a migration would be for the one time we move from ocs to gpp right?

swalker2m (approved):
Gah, I was confused and somehow thinking you were updating the `ocs` codebase.  This should be fine.  By all means :+1:

cquiroz:
Still it is a valid point if we were going to migrate old programs to GPP. Is that in the plans?

swalker2m:
> Still it is a valid point if we were going to migrate old programs to GPP. Is that in the plans?

That was a question in the inception review and I believe it is highly desired.  In the prototype `gem` project we included at least partial conversion.  It does require translation from the old names to the new.  At any rate, I don't want to be limited to whatever names we chose years ago in `ocs` if they don't make sense. Pending addressing any concerns from @jluhrs  I think we should go ahead with this.

jluhrs:
The web Seqexec translates them to EPICS values like this:
For GN:
```
        case FPU.IFU_1       => ("IFU-2", Beam.InBeam).some
        case FPU.IFU_2       => ("IFU-B", Beam.InBeam).some
        case FPU.IFU_3       => ("IFU-R", Beam.InBeam).some
```
For GS:
```
        case FPU.IFU_1       => ("IFU-2", Beam.InBeam).some
        case FPU.IFU_2       => ("IFU-B", Beam.InBeam).some
        case FPU.IFU_3       => ("IFU-R", Beam.InBeam).some
        case FPU.IFU_N       => ("IFU-NS-2", Beam.InBeam).some
        case FPU.IFU_N_B     => ("IFU-NS-B", Beam.InBeam).some
        case FPU.IFU_N_R     => ("IFU-NS-R", Beam.InBeam).some
```
I just have to make sure Observe has the right translations.

jluhrs (approved):
Looks OK. It is a good thing you didn't kept any of the old names. That will prevent any confusion.

PR #392, opened 2021-09-09
Sanitize MagnitudeSystem tags

cquiroz:
Turns out the tags with the funny units don't play nice with the json encoder/decoders. I'm giving them a simple tag and provide a `Display` for use on the UI


swalker2m (approved):
Looks good. :+1:

rpiaggio (approved):
LGTM

PR #394, opened 2021-09-12
Improved sky brighness calculation.

rpiaggio:
- New sky brightness calculation according to http://jira.gemini.edu:8080/browse/REL-3978.
- Don't returon airmass of 0.0 for altitudes less than 1 radian.
- Expose getSiderealTime in ImprovedSkyCalc.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/skycalc/ImprovedSkyCalcMethods.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/skycalc/ImprovedSkyCalcMethods.scala
@@ -64,10 +64,9 @@ trait ImprovedSkyCalcMethods {
    */
   def getAirmass(alt: Double): Double = {
     val secz = secant_z(alt)
-    if (secz >= 0.0)
-      if (secz < 12.0) return true_airmass(secz)
-      else return secz
-    0.0
+    if (secz < 0.0) 0.0
+    else if (secz < 12.0) true_airmass(secz)
+    else secz
   }
 
   /**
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/skycalc/ImprovedSkyCalcMethods.scala
```diff
@@ -60,24 +60,24 @@ trait ImprovedSkyCalcMethods {
...
-    * Return the airmass for the given altitude in degrees.
-    */
+   * Return the airmass for the given altitude in degrees.
+   */
   def getAirmass(alt: Double): Double = {
     val secz = secant_z(alt)
     if (secz >= 0.0)
       if (secz < 12.0) return true_airmass(secz)
-      else if (secz <= 99.0) return secz
+      else return secz
```
swalker2m, line 69: Are the `return`s a hold-over from the Java code?  That is, should it be something like:

```scala
if (secz < 0.0) 0.0
else if (secz < 12.0) true_airmass(secz) else secz
```

swalker2m (approved):
It's hard (for me) to judge the algorithm but the update looks good.  The task was requested for the December `ocs` release, but of course this is for GPP consumption only?  At any rate, looks good to me!  :+1:

modules/core/shared/src/main/scala/lucuma/core/math/skycalc/ImprovedSkyCalcMethods.scala
rpiaggio: Yes, this is remnant from auto conversion from Java code. I'm removing these at least.

rpiaggio:
>This is for GPP consumption only? 

Yes

swalker2m (approved):
Looks good. :+1:

PR #395, opened 2021-09-22
Make the target orders complete

swalker2m:
I noticed the existing name and tracking target orders ignore magnitudes.  Magnitudes should be moved into the ITC source definition at some point, but until then is there a reason not to include them in the `Order` definitions?

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -37,7 +37,7 @@ object Target extends WithId('t') with TargetOptics {
    * Not implicit.
    */
   val TargetTrackOrder: Order[Target] =
-    Order.by(t => (t.track, t.name))
+    Order.by(t => (t.track, t.name, t.magnitudes.toList))
 
   /**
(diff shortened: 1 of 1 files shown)


cquiroz:
What is the natural way to sort a `Magnitude`? via the band?

swalker2m:
> What is the natural way to sort a `Magnitude`? via the band?

I suppose there is no one true way. Similarly, there is also no natural reason to sort non-sidereal targets before sidereal ones or to sort RA before dec, etc.

This would sort by band and then brightness in that band as a way to disambiguate the order of two targets that are the same in all other respects.

cquiroz:
Do we need it to sort the results of a query?

swalker2m:
> Do we need it to sort the results of a query?

I had thought to group target environments by sorted set of `Target` but perhaps it should be grouped by hash sets of `Target`.

Regardless, it seems like the order should be defined for any two targets.  I do get a bit lost in the difference between `Order`, `Ordered`, and `Ordering`.  Perhaps these should be `Ordering`s instead?

cquiroz (approved):
LGTM

PR #398, opened 2021-10-05
Remove Asterism. Add TargetEnvironment.Id test.

rpiaggio:


toddburnside:
There is also an `AsterismType` in `lucuma.core.enum`. Is that still needed? It is for Ghost dual target vs single target, so maybe it will be used in the future?

cquiroz:
I don't recall a `CoordinateSuite` error before

tpolecat:
The floating point calculations sometimes don't come in within the expected bounds, it's probably not a big deal. Restart and cross fingers.

rpiaggio:
> There is also an `AsterismType` in `lucuma.core.enum`. Is that still needed? It is for Ghost dual target vs single target, so maybe it will be used in the future?

I hadn't noticed that one, seems to be ported from our previous version of the library with the rest of the enums. I'd keep it for the time being.

PR #401, opened 2021-10-06
Target as an ADT.

rpiaggio:


cquiroz:
I'm quite happy with this. I never really liked the either in the targets as it complicates quite a bit the lenses but we may have had, I recall some discussions about it earlier, maybe here:
https://github.com/gemini-hlsw/lucuma-core/pull/173

swalker2m:
We basically do it this way in [the `ocs` codebase](https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.spModel.core/src/main/scala/edu/gemini/spModel/core/Target.scala).  I don't have a strong preference but this does give us a way to add other kinds of targets like `ToO`.

@tpolecat I can't remember why we didn't follow the same model for `gem` when we were prototyping.

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -13,63 +13,230 @@ import lucuma.core.math._
...
-import monocle.std.either._
+import monocle.macros.GenPrism
 
 import scala.collection.immutable.SortedMap
 
 /** A target of observation. */
-final case class Target(
+sealed trait Target extends Product with Serializable {
+  def name: NonEmptyString
+  def magnitudes: SortedMap[MagnitudeBand, Magnitude]
```
swalker2m, line 25: This will eventually move to the ITC source definition, I assume.

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -13,63 +13,230 @@ import lucuma.core.math._
...
+      case (a @ SiderealTarget(_, _, _), b @ SiderealTarget(_, _, _))       =>
+        SiderealTarget.NameOrder.compare(a, b)
+      case (a @ NonsiderealTarget(_, _, _), b @ NonsiderealTarget(_, _, _)) =>
+        NonsiderealTarget.NameOrder.compare(a, b)
+      case (NonsiderealTarget(_, _, _), _)                                  => -1
+      case _                                                                => 1
+    }
+}
+
+trait SiderealTargetOptics {
```
swalker2m, line 123: Should we add something like `this: SiderealTarget.type =>` so that it cannot be mixed in elsewhere?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -13,63 +13,230 @@ import lucuma.core.math._
...
+
+  /** @group Optics */
+  lazy val name: Lens[SiderealTarget, NonEmptyString] =
+    Focus[SiderealTarget](_.name)
```
swalker2m, line 127: Just out of curiosity, why do we make these `lazy`?  Are there initialization issues with these?

PR #402, opened 2021-10-12
Spectral Distribution

cquiroz:
Spectral distribution model and enums


cquiroz:
This may change in the future but for now I want to use this to share the enums across explore and itc

PR #403, opened 2021-10-18
Added some enums to be used with the ObservationDB schema.

jluhrs:
These are enums that will be used by Observe


cquiroz (approved):
LGTM

PR #412, opened 2021-11-12
Make proper motion nullable in SiderealTarget

swalker2m:
When updating the fake GPP ODB to use the latest `lucuma-core`, I took advantage of all the new (?) target optics instead of defining them locally as I had before.  I hit one snag with proper motion though.  In particular, we need to support `SiderealTarget`s with and without proper motion (just as we do for radial velocity).  Since it is optional in `SiderealTracking` it should be nullable in the GraphQL API as well.

This PR changes `Target.properMotion` to `Optional[Target, Option[ProperMotion]]` accordingly.  Does this seem reasonable?

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -273,14 +273,14 @@ trait TargetOptics { this: Target.type =>
     sidereal.andThen(SiderealTarget.epoch)
 
   /** @group Optics */
-  val properMotion: Optional[Target, ProperMotion] =
-    siderealTracking.andThen(SiderealTracking.properMotion.some)
+  val properMotion: Optional[Target, Option[ProperMotion]] =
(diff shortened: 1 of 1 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -273,14 +273,14 @@ trait TargetOptics { this: Target.type =>
     sidereal.andThen(SiderealTarget.epoch)
 
   /** @group Optics */
-  val properMotion: Optional[Target, ProperMotion] =
-    siderealTracking.andThen(SiderealTracking.properMotion.some)
+  val properMotion: Optional[Target, Option[ProperMotion]] =
```
cquiroz, line 276: We discussed turning this into `Lens[Target, Option[ProperMotion]` but the structure won't allow it

cquiroz (approved):
LGTM

PR #415, opened 2021-11-19
Add AngularSize to targets.

rpiaggio:
The idea is to pull angular size information from catalogs when available so that we can present a full image of the target in the visualizer by setting the right FOV.

We can store this information in the ODB or not. We probably want to. Regardless of that, in explore we store the FOV in the separate user preferences DB if the user changes it. The target's `AngularSize` will be the default.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/AngularSize.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/AngularSize.scala
@@ -1,0 +1,29 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.model
+
+import cats.Eq
+import cats.Show
+import lucuma.core.math.Angle
+import monocle.Focus
+import monocle.Lens
+
+/**
+ * Angular size of an object in the sky. The major axis is the longest one, but we don't enforce
+ * this.
+ */
+final case class AngularSize(majorAxis: Angle, minorAxis: Angle)
+
+object AngularSize {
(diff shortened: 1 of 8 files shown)


swalker2m:
Is this ultimately part of the source definition,  or is it something else?

Screen Shot 2021-11-19 at 10 52 05

cquiroz:
It maybe related but this is more of a visual thing to have a proper fov on the aladin display

modules/core/shared/src/main/scala/lucuma/core/model/AngularSize.scala
```diff
@@ -0,0 +1,29 @@
...
+import cats.Show
+import lucuma.core.math.Angle
+import monocle.Focus
+import monocle.Lens
+
+/**
+ * Angular size of an object in the sky. The major axis is the longest one, but we don't enforce
+ * this.
+ */
+final case class AngularSize(majorAxis: Angle, minorAxis: Angle)
```
cquiroz, line 16: There is this `https://github.com/gemini-hlsw/react-aladin/blob/master/demo/src/main/scala/react/aladin/package.scala#L22` though it is more basic

cquiroz: Never mind, it should be the other way around. Make `react-aladin` depend on `AngularSize`

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/AngularSize.scala
rpiaggio: I'm not sure both classes have the same definition. In `Fov` the angles are aligned to the `x` and `y` axes, while in `AngleSize` they are not: it describes the ellipse that inscribes the target.

PR #416, opened 2021-11-19
Change SiderealTarget.properMotion from Optional to Lens[..., Option[...]]

rpiaggio:
`Target.properMotion.getOption` was erroneously returning `Some(None)` instead of `None` when there was no PM defined.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -174,16 +174,16 @@ trait SiderealTargetOptics { this: SiderealTarget.type =>
     tracking.andThen(SiderealTracking.epoch)
 
   /** @group Optics */
-  val properMotion: Optional[SiderealTarget, ProperMotion] =
-    tracking.andThen(SiderealTracking.properMotion.some)
+  val properMotion: Lens[SiderealTarget, Option[ProperMotion]] =
+    tracking.andThen(SiderealTracking.properMotion)
 
   /** @group Optics */
   val properMotionRA: Optional[SiderealTarget, ProperMotion.RA] =
-    properMotion.andThen(ProperMotion.ra)
+    properMotion.some.andThen(ProperMotion.ra)
 
   /** @group Optics */
   val properMotionDec: Optional[SiderealTarget, ProperMotion.Dec] =
-    properMotion.andThen(ProperMotion.dec)
+    properMotion.some.andThen(ProperMotion.dec)
 }
 
 trait NonsiderealTargetOptics { this: NonsiderealTarget.type =>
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -276,14 +276,14 @@ trait TargetOptics { this: Target.type =>
     sidereal.andThen(SiderealTarget.epoch)
 
   /** @group Optics */
-  val properMotion: Optional[Target, Option[ProperMotion]] =
-    siderealTracking.andThen(SiderealTracking.properMotion)
+  val properMotion: Optional[Target, ProperMotion] =
```
cquiroz, line 279: @swalker2m I think we had some discussion about this?

cquiroz:
Should there be a test for this?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
rpiaggio: > Should there be a test for this?

I'm not sure what we would test. We were testing `Optional` laws already, which passed.

I don't remember a discussion about this. Having `properMotion: Optional[Target, ProperMotion]` seems consistent with `properMotionRA: Optional[Target, ProperMotion.RA]` and `properMotionDec: Optional[Target, ProperMotion.Dec]`.

swalker2m: https://github.com/gemini-hlsw/lucuma-core/pull/412

swalker2m: See also: `val radialVelocity: Optional[Target, Option[RadialVelocity]]`

rpiaggio: Oh ok, I didn't remember that PR. I thought this was a copy/paste error on my part when I redid these optics a while ago.

It seems to me then that if we want to set `sideralTarget.properMotion` to `None` via optics we should have `SiderealTarget.properMotion` be a `Lens[SiderealTarget, Option[ProperMotion]` and `Target.properMotion = sidereal.andThen(SiderealTarget.properMotion)`.

Unless there's an objection to this, I'll make that change then.

cquiroz: I think the problem is that Target is an ADT and non sideral cannot take a proper motion, this laws fail

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -174,16 +174,16 @@ trait SiderealTargetOptics { this: SiderealTarget.type =>
     tracking.andThen(SiderealTracking.epoch)
 
   /** @group Optics */
-  val properMotion: Optional[SiderealTarget, ProperMotion] =
-    tracking.andThen(SiderealTracking.properMotion.some)
+  val properMotion: Lens[SiderealTarget, Option[ProperMotion]] =
+    tracking.andThen(SiderealTracking.properMotion)
```
swalker2m, line 178: This seems right to me.

swalker2m (approved):
:+1:

cquiroz (approved):
LGTM

PR #419, opened 2021-11-30
Switch to cats-time 0.5.0

cquiroz:
Signed-off-by: Carlos Quiroz <carlos.quiroz@noirlab.edu>

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Epoch.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Epoch.scala
@@ -6,6 +6,7 @@ package lucuma.core.math
 import cats.Order
 import cats.Show
 import cats.syntax.all._
+import eu.timepit.refined._
 import eu.timepit.refined.api.Refined
 import eu.timepit.refined.auto._
 import eu.timepit.refined.cats._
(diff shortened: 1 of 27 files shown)


toddburnside (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/Epoch.scala
```diff
@@ -77,13 +78,13 @@ object Epoch extends EpochOptics {
     * Standard epoch.
     * @group Constructors
     */
-  val J2000: Epoch = Julian.fromIntegralYears(2000)
+  val J2000: Epoch = Julian.fromIntegralYears(refineMV[Year](2000))
```
rpiaggio, line 81: When using custom refined types (like `IntYear`), we can define:

`object IntYear extends RefinedTypeOps.Numeric[IntYear, Int]`

which allows us to define instances at compile time as `IntYear(2000)`. At runtime we can use `IntYear.unsafeFrom(xxx): FTP`,  `IntYear.from(t: T): Either[String, FTP]` and `IntYear.unapply(t: T): Option[FTP]`, and a bunch of methods that return `Validated` when we `import eu.timepit.refined.cats.syntax._`.

If the underlying type is not numeric, we can just do `extends RefinedTypeOps[<Refined>, <Underlying>]`.

I don't think this changes anything besides code style, which is of course a matter of preference.

rpiaggio (approved):
LGTM

PR #422, opened 2021-12-06
Runtime units + TargetBrightness rework.

rpiaggio:
This PR is an attempt to bridge the gap between `coulomb`'s units, which are types, and our need to model units in runtime.

This is done mainly by introducing a `UnitOfMeasure[U]` type class for unit `U`. Instances are derived by leveraging the fact that `coulomb` provides runtime information for unit types defined through `BaseUnit` and `DerivedUnit`.

The type `Qty` emulates `coulomb.Quantity` but with the runtime unit type.

Aside from this, a way to group units is introduced. Grouping is just a type tag over units and can be hierarchical. Types are defined to enforce the usage of units that belong to a certain group. This is done only at the runtime level. We *could* add logic to define unit groups at the type level and in fact I have version with that, but decided not to include unless needed, in order to simplify things.

- `Brightness` (ex `Magnitude`) has been renamed to the more accurate `TargetBrightness`.
- Brightness units can now be defined as any other unit using `coulomb`.
- Brightness units can now be grouped into "integrated" and "surface" units.
- Regular `Enumerated` instances can be defined from their runtime instances. Tags have changed and are now the units abbreviation (eg: `erg/s/cm²/Hz`).

Finally, some files were changed because of organizing imports.

[skip-sc]


modules/core/shared/src/main/scala/lucuma/core/math/dimensional/quantity.scala
```diff
@@ -0,0 +1,41 @@
...
+package lucuma.core.math.dimensional
+
+import coulomb._
+import monocle.Focus
+import monocle.Lens
+
+/**
+ * A magnitude of type `N` and a runtime representation of a physical unit.
+ */
+trait Qty[N] {
```
cquiroz, line 13: Isn't this essentially the same as `Quantity`? Why do we need it?

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
```diff
@@ -0,0 +1,82 @@
...
+  }
+}
+
+/**
+ * Type-parametrized runtime representation of physical unit `U`.
+ *
+ * Can be automatically derived if there's an implicit `BaseUnit[U]` or `DerivedUnit[U, _]` in
+ * scope.
+ */
+trait UnitOfMeasure[U] extends UnitType {
```
cquiroz, line 34: Is this something that should be in `coulomb`?

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -78,23 +80,76 @@ trait units {
...
+
+  type ErgsFrequencyBrightness
+  implicit val defineErgsFrequencyBrightness =
+    DerivedUnit[ErgsFrequencyBrightness, Erg %/ (Second %* (Centimeter %^ 2) %* Hertz)](
+      abbv = "erg/s/cm²/Hz"
+    )
+
+  // Surface Brightness units
+
+  // Derive a surfrace unit from an integrated unit
```
cquiroz, line 123: typo

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/quantity.scala
rpiaggio: `Quantity[N, U]` is a value class wrapping an `N`. The unit `U` is erased at runtime. Also, if we want to define a value that can have different units we are in trouble since `U` is part of the type definition of a `Quantity`.

`Qty` hides the units from the type definition and moves it to a runtime value.

Better name suggestions are very welcome, please!

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
rpiaggio: If `coulomb` had a common ancestor of `BaseUnit[U]` and `DerivedUnit[U, D]` that was typed in `U`, we could use it and wouldn't need this.

PR #428, opened 2021-12-16
Brightness adt

rpiaggio:
* `TargetBrightness` renamed to `BandBrightness`.
* `Target` ADT members now within `Target` object (eg: `SiderealTarget` is now `Target.Sidereal`)
* A `Target` now has a `SourceProfile`, which can be `SourceProfile.Point`, `SourceProfile.Uniform` or `SourceProfile.Gaussian`.
* A `SourceProfile` has a `SpectralDefinition`, which can either be:
  * `BandNormalized`, which combines an unnormalized SED + band brightnesses. (Unnormalized SEDs are all SEDs except for emission lines.)
  * `EmissionLines`, which combines emission lines for different wavelength + one flux density continuum value.

Unless further requiremens show up, I expect this to resemble the final model.


cquiroz (approved):
LGTM

PR #429, opened 2021-12-17
Renames, aliases, deep optics.

rpiaggio:


swalker2m (approved):
Looks good to me.

modules/core/shared/src/main/scala/lucuma/core/model/SpectralDefinition.scala
```diff
@@ -25,20 +26,18 @@ sealed trait SpectralDefinition[T]
 object SpectralDefinition {
 
   final case class BandNormalized[T](
-    sed:          UnnormalizedSpectralEnergyDistribution,
```
swalker2m, line 28: My impression, probably wrong, is that SED is a pretty common (for astronomy) abbreviation but `USED` might seem odd.  If I were correct, we might consider `UnnormalizedSed` to shorten it, albeit not by as much?  At any rate, when it appears in the GraphQL API we might want to check whether `UnnormalizedSed` would be more recognizable to an astronomer.

rpiaggio: Sounds good, I'll just rename `UnnormalizedSpectralEnergyDistribution` to `UnnormalizedSED` then.

rpiaggio: Sounds good, I'll just rename `UnnormalizedSpectralEnergyDistribution` to `UnnormalizedSED` then.

PR #430, opened 2021-12-21
Add obsolete field

cquiroz:
Mark BHROS as obsolete


swalker2m (commented):
I thought this would come from Resource in GPP and be expanded with time intervals for validity etc.  Perhaps I missed a conversation about this.

cquiroz:
Yes, but there is no ETA for that yet, this is now mocked on the ITC and I thought maybe better to push it to `core`

swalker2m (approved):
> Yes, but there is no ETA for that yet

That certainly a fair point and I'll go ahead and approve.  I kind of fear we'll end up just keeping the boolean flags though.

cquiroz:
I can certainly live without this change and maybe it is a bad signal, let's close it.

PR #431, opened 2021-12-21
Add objectType to CatalogInfo. Move it to Target.

rpiaggio:
- `CatalogId` renamed to `CatalogInfo`.
- Moved `CatalogInfo` to `Target`.
- Added `objectType`, so that we can store and show the object type retrieved from the catalog. Users may find this helpful when determining the profile and SED of targets.

Also, organize imports in tests.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/BandBrightness.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/BandBrightness.scala
@@ -9,6 +9,7 @@ import lucuma.core.enum.Band
 import lucuma.core.math.BrightnessUnits._
 import lucuma.core.math.BrightnessValue
 import lucuma.core.math.dimensional._
+import lucuma.core.util.Enumerated
 import monocle.Focus
 import monocle.Lens
 
@@ -60,6 +61,26 @@ object BandBrightness {
     error:    BrightnessValue
   ): BandBrightness[T] =
     new BandBrightness(quantity, band, error.some)
+
+  def apply[T, U](value: BrightnessValue, band: Band, error: Option[BrightnessValue])(implicit
+    enumerated:          Enumerated[GroupedUnitType[Brightness[T]]],
+    unit:                UnitOfMeasure[U]
+  ): Option[BandBrightness[T]] =
+    enumerated.all
+      .find(_.ungrouped === unit)
+      .map(u => new BandBrightness(u.withValue(value), band, error))
+
+  def apply[T, U](value: BrightnessValue, band: Band, error: BrightnessValue)(implicit
+    enumerated:          Enumerated[GroupedUnitType[Brightness[T]]],
+    unit:                UnitOfMeasure[U]
+  ): Option[BandBrightness[T]] =
+    apply[T, U](value, band, error.some)
+
+  def apply[T, U](value: BrightnessValue, band: Band)(implicit
+    enumerated:          Enumerated[GroupedUnitType[Brightness[T]]],
+    unit:                UnitOfMeasure[U]
+  ): Option[BandBrightness[T]] =
+    apply[T, U](value, band, none)
 
   /** Secondary constructor using default units for the band. */
   def apply[T] = new GroupApplied[T]
--- a/modules/core/shared/src/main/scala/lucuma/core/model/CatalogInfo.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/CatalogInfo.scala
@@ -9,6 +9,8 @@ import eu.timepit.refined.cats._
 import eu.timepit.refined.collection.NonEmpty
 import eu.timepit.refined.types.string.NonEmptyString
 import lucuma.core.enum.CatalogName
+import monocle.Focus
+import monocle.Lens
 
 final case class CatalogInfo(catalog: CatalogName, id: NonEmptyString, objectType: String)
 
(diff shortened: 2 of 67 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/CatalogInfo.scala
```diff
@@ -0,0 +1,36 @@
...
+import cats._
+import eu.timepit.refined._
+import eu.timepit.refined.cats._
+import eu.timepit.refined.collection.NonEmpty
+import eu.timepit.refined.types.string.NonEmptyString
+import lucuma.core.enum.CatalogName
+import monocle.Focus
+import monocle.Lens
+
+final case class CatalogInfo(catalog: CatalogName, id: NonEmptyString, objectType: String)
```
swalker2m, line 15: I would have expected `objectType` to be, maybe, an `Option[NonEmptyString]`.  Do we always get an `objectType` from all the catalogs?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -62,12 +64,13 @@ object Target extends WithId('t') with TargetOptics {
     name:          NonEmptyString,
     ephemerisKey:  EphemerisKey,
     sourceProfile: SourceProfile,
+    catalogInfo:   Option[CatalogInfo],
```
swalker2m, line 67: We have asteriod vs. comet vs. minor/major body in the nonsidereal side, which appears in the horizons key.  The object type overlaps a bit with that.  Would we fill in an `objectType` from the `EphemerisKey`?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -35,12 +36,13 @@ object Target extends WithId('t') with TargetOptics {
     name:          NonEmptyString,
     tracking:      SiderealTracking,
     sourceProfile: SourceProfile,
+    catalogInfo:   Option[CatalogInfo],
```
swalker2m, line 39: I suppose there is no standard for the values we might see in order to enumerate them.  I wonder if we could parse these values when possible into something we could use.  In the mockups we show, for example, galaxies with a unique icon in target lists.

swalker2m: The object types of sidereal objects would not overlap with those of nonsidereal objects, but if it is just a `String` that we display then it seems fine to put this on `Target`.  If we could dig out what the types actually are though, then it might make more sense to separate the sidereal types from the nonsidereal ones and have `SiderealObjectType` and `NonSiderealObjectType`.

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
swalker2m: In fact the `EphemerisKey` is kind of `CatalogInfo` from horizons.  I wonder if we need `SiderealCatalogInfo` and `NonSiderealCatalogInfo`.

swalker2m (approved):
This raises some questions in my mind but seems like an improvement.

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
rpiaggio: In the case of Simbad, the requirement here is to present to the user the object type, morphological type and spectral type, so that they can use this to fill out the SED fields. Something along these lines in the mockup:



So for Simbad, we would concatenate those 3 fields into a single string and put it as `objectType`, just to be able to present it to the user. For the moment, we are not extracting semantic information out of this. The UI could actually fetch this info from Simbad when needed, since we are storing the catalog id. So, this actually acts as a cache to avoid that extra query to Simbad.

I moved `CatalogInfo` to `Target` since I'm expecting some kind of object type string to come from Horizons too, but this seems to add redundance with the ephemeris key. We should probably move `CatalogInfo` into `Target.Sidereal` then. I'll make that change. I'll also change the `objectType` to an `Option[NonEmptyString]`. 

I agree it would be nice to have a nicely modeled object type on the target and show this as an icon, but that's not what this PR attempts to address. When we do have such field, the parser can make a best-effort attempt to infer it based on the catalog's object type info. Also, the plan is for the parser to eventually attempt to infer the SED from the catalog object type.

rpiaggio:
Changes:
- `CatalogInfo` only in `Target.Sidereal`.
- `CatalogInfo.objectType` is now an `Option[NonEmptyString]`.
- Merged `integratedUnnormalizedSED` and `surfaceUnnormalizedSED` into a single `unnormalizedSED` optic whenever possible.
- Added convenience constructors to `BandBrightness` and added construction tests.

modules/core/shared/src/main/scala/lucuma/core/model/CatalogInfo.scala
```diff
@@ -0,0 +1,53 @@
...
+import eu.timepit.refined.collection.NonEmpty
+import eu.timepit.refined.types.string.NonEmptyString
+import lucuma.core.enum.CatalogName
+import monocle.Focus
+import monocle.Lens
+
+final case class CatalogInfo(
+  catalog:    CatalogName,
+  id:         NonEmptyString,
+  objectType: Option[NonEmptyString]
```
swalker2m, line 19: I think this is good (the `Option`).  The `id` was added I believe to track guide star catalog ids and the `objectType`, I see now, was mostly inspired by Simbad results.  I'm not sure what `objectType` information we can expect from guide star catalogs so it's good that it is optional.

modules/core/shared/src/main/scala/lucuma/core/model/BandBrightness.scala
```diff
@@ -61,6 +62,26 @@ object BandBrightness {
...
+    enumerated:          Enumerated[GroupedUnitType[Brightness[T]]],
+    unit:                UnitOfMeasure[U]
+  ): Option[BandBrightness[T]] =
+    apply[T, U](value, band, error.some)
+
+  def apply[T, U](value: BrightnessValue, band: Band)(implicit
+    enumerated:          Enumerated[GroupedUnitType[Brightness[T]]],
+    unit:                UnitOfMeasure[U]
+  ): Option[BandBrightness[T]] =
+    apply[T, U](value, band, none)
```
swalker2m, line 83: Thanks, this should help.

swalker2m (approved):
Looks great!

PR #435, opened 2021-12-22
Removable catalog info optic

swalker2m:
Since `catalogInfo` is optional, it should be removable via the optic (like proper motion, radial velocity, and parallax).


modules/tests/shared/src/test/scala/lucuma/core/model/TargetSuite.scala
```diff
@@ -374,7 +374,6 @@ final class TargetSuite extends DisciplineSuite {
   checkAll("Target.baseCoordinates", OptionalTests(Target.baseCoordinates))
   checkAll("Target.baseRA", OptionalTests(Target.baseRA))
   checkAll("Target.baseDec", OptionalTests(Target.baseDec))
-  checkAll("Target.catalogInfo", OptionalTests(Target.catalogInfo))
```
swalker2m, line 377: This was duplicated below.

cquiroz (approved):
LGTM

PR #436, opened 2021-12-22
Derive runtime compound units. Minimally redefine custom units. Normalize custom unit names.

rpiaggio:
- Runtime units (`UnitOfMeasure[U]`) were previously derived from implicit `BaseUnit` or `DerivedUnit` instances. This left out compound units (eg: `Meters %/ Second`), requiring us to define a `DerivedUnit` whenever we needed a runtime representation of a compound unit.

This PR lifts that restriction by leveraging `coulomb`s machinery to derive `UnitString` instances, which provides an implicit instance of a base, derived or compound unit.

- A convenience method is added for us to override derived `UnitString` instances, which sometimes don't look nice.

- Custom units are now derived as compound types whenever possible, instead of using `DerivedUnit`.

- Aliases for compound units are now named using a unified criterion (eg: `WattsPerMeter2MicrometerArcsec2` instead of `WattsBrightnessPerArcsec2`).

- Unrelated: Remove `ocs2tag` from `StellarLibrarySpectrum`.


modules/core/shared/src/main/scala/lucuma/core/enum/StellarLibrarySpectrum.scala
```diff
@@ -7,63 +7,62 @@ import lucuma.core.util.Enumerated
 
 sealed abstract class StellarLibrarySpectrum(
   val tag:         String,
-  val sedSpectrum: String,
-  val ocs2Tag:     String
```
cquiroz, line 11: 👍

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
```diff
@@ -18,26 +21,32 @@ import coulomb.define._
  */
 trait UnitType { self =>
   type Type
-  def definition: UnitDefinition
```
cquiroz, line 21: I was going to do matching on `definition` should I use `name` instead?

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
```diff
@@ -104,10 +104,11 @@ trait GroupedUnitOfMeasure[UG, U] extends UnitOfMeasure[U] with GroupedUnitType[
...
     new GroupedUnitOfMeasure[UG, U] {
-      val definition = unit.definition
+      val name = unit.name
+      val abbv = unit.abbv
     }
 
   // `UnitDefiniton`s are expected to be singletons.
   implicit def eqGroupedUnitOfMeasure[UG, U]: Eq[GroupedUnitOfMeasure[UG, U]] =
-    Eq.instance((a, b) => a.definition == b.definition)
+    Eq.fromUniversalEquals
```
cquiroz, line 113: This will use the `equals` on `UnitOfMeasure` right?

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
rpiaggio: Can you use `===` ?

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
rpiaggio: Yes. Which ends up matching on both `name` and `abbv`.

toddburnside (approved):
LGTM

PR #437, opened 2021-12-23
Tagged unit types

rpiaggio:
I was not happy with the concepts of `GroupedUnitType` and `GroupedUnitQty`. They were cumbersome and misleading since they didn't group anything and just type-tagged units and quantities. So, it seems that a simpler approach is to use, well, tagged types. This is what this PR does, by using tags on `UnitType` and `Qty` instead.

I think this generally makes code, eg `BandBrightness` constructors, a bit more intuitive. Also, associations between units and tags now exist at the type-level (via the `IsTaggedUnit` type class).

Notes:
- `shapeless` was already brought in transitively. I just added it explicitly to `build.sbt` now.
- Tagged types don't seem to be part of `shapeless` 3. They can be easily implemented though using intersection types, as shown here: https://contributors.scala-lang.org/t/poor-or-rich-mans-refinement-types-in-scala-3-x/4647/12.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,6 +15,7 @@ lazy val refinedVersion        = "0.9.28"
 lazy val catsTimeVersion       = "0.5.0"
 lazy val circeVersion          = "0.14.1"
 lazy val catsScalacheckVersion = "0.3.1"
+// lazy val shapelessVersion      = ""
 
 inThisBuild(
   Seq(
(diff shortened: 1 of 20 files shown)


build.sbt
```diff
@@ -15,6 +15,7 @@ lazy val refinedVersion        = "0.9.28"
 lazy val catsTimeVersion       = "0.5.0"
 lazy val circeVersion          = "0.14.1"
 lazy val catsScalacheckVersion = "0.3.1"
+lazy val shapelessVersion      = "2.3.3"
```
cquiroz, line 18: cough, scala 3

cquiroz (approved):
I'm not sure I understand the details but 👍

PR #441, opened 2021-12-28
CI JVM Adoptium 8 & 17 + more heap for sbt.

rpiaggio:


.github/workflows/scala.yml
```diff
@@ -15,6 +15,6 @@ jobs:
       - name: Set up JDK
         uses: olafurpg/setup-scala@v13
         with:
-          java-version: adoptium@17
+          java-version: adoptium@11
```
cquiroz, line 18: Sorry to do this, but maybe we should try to keep 17 and use 8 for releases

rpiaggio: Ok with releasing for 8, but I think CI tests should include the same JVM we use for release.

I'll look into using a version matrix that includes 8 and 17.

.github/workflows/release.yml
```diff
@@ -11,9 +15,11 @@ jobs:
         with:
           fetch-depth: 0
       - uses: olafurpg/setup-scala@v13
+        with:
+          java-version: adoptium@11
```
cquiroz, line 19: Will we still realease on 11?

rpiaggio: > Will we still realease on 11?

Ouch, missed that. Fixed to 8 now.

rpiaggio: > Maybe it makes more sense to build on the same jvm as test but put the `-target` option to go to 1.8?

We are testing on both 8 and 17 now. BTW, they are run in parallel, so it doesn't add time to check the PR. (It does increase the probability of flaky tests failing though).

PR #442, opened 2021-12-28
Conversions between SourceProfile types.

rpiaggio:
Also, renaming of some methods and optics `*T` -> `*Tagged`.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
@@ -20,20 +20,58 @@ import shapeless.tag.@@
 
 import scala.collection.immutable.SortedMap
 
-sealed trait SourceProfile extends Product with Serializable
+sealed trait SourceProfile extends Product with Serializable {
+
+  /**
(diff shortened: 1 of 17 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
```diff
@@ -20,20 +20,58 @@ import shapeless.tag.@@
...
+sealed trait SourceProfile extends Product with Serializable {
+
+  /**
+   * Convert to a `Point` source profile. Units are converted to their `Integrated` versions, if
+   * necessary.
+   */
+  def toPoint: SourceProfile.Point
+
+  /**
+   * Convert to a `Uniform` source profile. Units are converted to their `Surface` versions, if
```
cquiroz, line 32: I guess they are converted without changing the number right? only the units

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
rpiaggio: Yes, only the units.

PR #443, opened 2021-12-28
More idiomatic names for tagged units.

rpiaggio:
- Create a type alias `Of` for `@@`. Besides allowing more idiomatic type names (see below), this removes the need for code to import directly from `shapeless`.
- `Qty` -> `Measure`.
- `UnitType` -> `Units`.

Before: `UnitType @@ Brightness[Integrated]`
After: `Units Of Brightness[Integrated]`

Before: `Qty[PosBigDecimal] @@ LineFlux[Surface]`
After: `Measure[PosBigDecimal] Of LineFlux[Surface]`

Other changes:
- `IsTagged` -> `Tag`
- `IsTaggedUnit` -> `TaggedUnit`


modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
```diff
@@ -0,0 +1,50 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math.dimensional
+
+import _root_.cats.kernel.Eq
```
cquiroz, line 6: Why not plain `cats.Eq`?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
rpiaggio: Good catch, must have been an automatic import. It was probably placed below `coulomb._`, so the `_root_` was needed to avoid going into `coulomb.cats`. `OrganizeImports` then moved it to the top.

PR #444, opened 2021-12-29
Remove GaussianSource.

rpiaggio:
Following @cquiroz 's suggestion, we move the `full width at half maximum of the seeing disc` `Angle` directly to `SourceProfile.Gaussian`. This simplifies the model a bit.

The only reason why I had kept `GaussianSource` was because of the `arcsec` optic, but I don't think it will be sorely missed. If such functionality is needed, we can add an optic to `Angle`.


modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
```diff
@@ -42,14 +43,20 @@ sealed trait SourceProfile extends Product with Serializable {
...
+   */
   final case class Point(spectralDefinition: SpectralDefinition[Integrated]) extends SourceProfile {
 
     override def toPoint: Point = this
 
     override def toUniform: Uniform =
       Uniform(spectralDefinition.to[Surface])
 
-    override def toGaussian: Gaussian = Gaussian(GaussianSource.Zero, spectralDefinition)
+    override def toGaussian: Gaussian = Gaussian(Angle.Angle0, spectralDefinition)
```
cquiroz, line 59: I didn't know you could convert them, I guess it makes sense that a point is gaussian with 0 width

modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
```diff
@@ -61,14 +68,20 @@ object SourceProfile {
...
+   */
   final case class Uniform(spectralDefinition: SpectralDefinition[Surface]) extends SourceProfile {
 
     override def toPoint: Point = Point(spectralDefinition.to[Integrated])
 
     override def toUniform: Uniform = this
 
     override def toGaussian: Gaussian =
-      Gaussian(GaussianSource.Zero, spectralDefinition.to[Integrated])
+      Gaussian(Angle.Angle0, spectralDefinition.to[Integrated])
```
cquiroz, line 84: This seems a bit odd, Would uniform have a width 0?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
rpiaggio: The use case I imagine for this is in the UI having a drop-down for `Source Profile` with options `Point`, `Uniform` and `Gaussian`. If changing from `Point`/`Uniform` -> `Gaussian`, the angle input will appear, with a default value. I set it to `Angle0` but maybe it makes more sense to have another default value here.

This conversion between source profile types can also be used in the API.

PR #446, opened 2021-12-31
Restrict what cool star models can be build

cquiroz:
ITC doesn't support arbitrary temperatures for Cool Stars, This PR restricts what can be built to what's available on the backend


rpiaggio (approved):
LGTM

PR #447, opened 2022-01-03
switch BrightnessValue.fromBigDecimal to SplitEpi

swalker2m:
Shouldn't `BrightnessValue.fromBigDecimal` be a `SplitEpi`?  Ignoring `BigDecimal` values that are too big for an `Int` (which we were doing before as well) it seems like you always get some `BrightnessValue` from a `BigDecimal`.  It's never optional.


rpiaggio (approved):
LGTM

PR #448, opened 2022-01-03
Serialized form for units. BandBrightness constructors.

rpiaggio:
This PR addresses 2 issues:
- Keep in one place serialized forms of runtime units.
  - We can't use units abbreviations as GraphQL `enum` members due to the constrained charset.
  - Therefore we add a `serialized` property to `Units`.
  - Convenience constructors are added to create `coulomb` type instances together with the `serialized` string.
  - `Enumerated` instances of brightness tagged units now use `serialized` as `tag` instead of `abbv`.
- Constructors of `BandBrightness` where the units were specified as a type still required another type parameter for the brightness type (`Integrated`, `Surface`) even when it could be inferred.
  - Now, it's inferred:
    - Before: `BandBrightness[Integrated, ABMagnitude](BrightnessValue.fromDouble(10.0), Band.R)`
    - After: `BandBrightness[ABMagnitude](BrightnessValue.fromDouble(10.0), Band.R)`
  - Previous `BandBrightness[T]` constructor for brightness type `T` was used to infer default units for the band. It's now renamed explicitly to `BandBrightness.withDefaultUnits[T]`:
    - Before: `BandBrightness[Integrated](BrightnessValue.fromDouble(10.0), Band.R)`
    - After: `BandBrightness.withDefaultUnits[Integrated](BrightnessValue.fromDouble(10.0), Band.R)`
    - 
Also, more heap and stack is given to CI `sbt` runs. This is an attempt to solve compilation issues in CI.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/dimensional/coulomb.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/dimensional/coulomb.scala
@@ -1,0 +1,60 @@
+// Copyright (c) 2016-2021 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math.dimensional
+
+import coulomb.UnitTypeName
+import coulomb.define.BaseUnit
+import coulomb.define.DerivedUnit
+import coulomb.unitops.UnitString
+import lucuma.core.syntax.string._
+import lucuma.core.util.TypeString
+import spire.math.Rational
+
+object BaseUnitDef {
+
+  /** Convenience method to create a coulomb `BaseUnit` and `TypeString` in one go. */
+  def apply[U](name: String = "", abbv: String = "", serialized: String = "")(implicit
+    ut:              UnitTypeName[U]
+  ): BaseUnit[U] with TypeString[U] = {
+    val base = BaseUnit[U](name, abbv)
+
+    val _serialized = if (serialized != "") serialized else base.name.toScreamingSnakeCase
+
+    new BaseUnit[U](base.name, base.abbv) with TypeString[U] {
+      override val serialized: String = _serialized
+    }
+  }
+}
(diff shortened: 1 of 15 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/dimensional/coulomb.scala
```diff
@@ -0,0 +1,60 @@
...
+}
+
+object DerivedUnitDef {
+
+  /** Convenience method to create a coulomb `DerivedUnit` and `TypeString` in one go. */
+  def apply[U, D](
+    coef:        Rational = Rational(1),
+    name:        String = "",
+    abbv:        String = "",
+    serialized:  String
```
cquiroz, line 37: I thought non default members needed to be declared first

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
```diff
@@ -38,7 +49,7 @@ trait Units { self =>
     case _           => false
   }
 
-  override def hashCode: Int = Objects.hash(name, abbv)
+  override def hashCode: Int = Objects.hash(name, abbv, serialized)
```
cquiroz, line 52: Do you need to ask override `equals`?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/coulomb.scala
rpiaggio: Actually, that should have a default value.

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala
rpiaggio: You are right, `serialized` should take part in both `hashCode` and `equals` or none.

PR #449, opened 2022-01-04
disable parallel compilation

tpolecat:
This seems to stabilize the build and eliminate the crashes documented at https://github.com/tpolecat/lucuma-core-crashes, at the cost of increased build time. See also https://github.com/scala/bug/issues/10682 for discussion.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,5 +1,7 @@
 import sbtcrossproject.CrossType
 
+Global / concurrentRestrictions += Tags.limit(Tags.Compile, 1)
+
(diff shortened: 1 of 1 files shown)


cquiroz:
Could we enable it only on CI?

tpolecat:
I suggest we try it this way for a while and see how bad it is. If people find that they're commenting that line out locally then maybe we could make it CI-only. It was very very unstable on my local machine.

rpiaggio (approved):
LGTM

rpiaggio:
Thanks for taking care of this

PR #450, opened 2022-01-04
Use 'normal' tag

swalker2m:
Might we use the "normal" tag for `CoolStarTemperature`?  GraphQL enums cannot start with a number so we'd need to define them with a leading `T`.  That's fine, but then the `Decoder` defined in `Enumerated`, which looks for the tag, doesn't work for reading inputs.  It sees, for example, `T400_K` but that doesn't match the (screaming snake) tag `400_K`.  We can overcome that with a custom `Decoder` but maybe the "normal" tag is okay?

On the other hand, perhaps relying upon the `Enumerated` decoder is too fragile though?  Discuss.


cquiroz (approved):
LGTM

PR #451, opened 2022-01-05
Dimensional basic documentation.

rpiaggio:
Also, a couple of fixes that came up while documenting.


swalker2m (approved):
I'm going to read this more carefully tomorrow but I can see it's going to be helpful.  Thanks for doing this! :+1:

DIMENSIONAL.md
```diff
@@ -0,0 +1,51 @@
+# The `dimensional` package
+
+- We use `coulomb` for unit analysis.
+- `coulomb` uses static types to represent units and this information doesn't make it to runtime.
+  - Eg:
```
swalker2m, line 5: Did you mean to add an example here?

rpiaggio: Yes, which later made it into the next section. I will just remove this line.

DIMENSIONAL.md
```diff
@@ -0,0 +1,51 @@
...
+`Units` can be tagged too. We can access the tagged `Units` of a tagged `Measure` by using the `unitsTagged` lens:
+
+```scala
+scala> val unitsTagged: Units Of Velocity = Measure.unitsTagged.get(velocityMTagged)
+val unitsTagged: lucuma.core.math.dimensional.Of[lucuma.core.math.dimensional.Units,Velocity] = m/s
+```
+
+### Final note
+
+- No numeric operations are supported on `Measure`, but they could be implemented. (A `Measure` can be converted to a `Quantity` and back).
```
cquiroz, line 51: Do you mean operations at the type level or e.g. adding `Measure` objects like you can do with `Quantity`?

rpiaggio: I meant the latter, I'll clarify.

PR #452, opened 2022-01-05
Remove BandBrightness; Wavelength from EmissionLine.

rpiaggio:
The model allowed construction of inconsistent brightness maps in `SpectralDefinition.BandNormalized` and `SpectralDefinition.EmissionLines`: the `Band` of `Wavelength` used as key in each map entry may not coincide with the one inside the `BandBrightness` or `EmissionLine`.

To solve this, `Wavelength` was removed from `EmissionLine`.

The `BandBrightness` class was removed altogether, in favor of adding an `error` value to `Measure`.


modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
```diff
@@ -12,18 +13,32 @@ import shapeless.tag
...
    */
-  def toQuantity: Quantity[N, units.Type] = value.withUnit[units.Type]
+  def toExactQuantity: Quantity[N, units.Type] = value.withUnit[units.Type]
+
+  /**
+   * Convert error to `coulomb.Quantity`.
+   */
+  def valueQuantity: Option[Quantity[N, units.Type]] = error.map(_.withUnit[units.Type])
+
+  // def withError(error: N): self.type = copy(error = error.some).asInstanceOf[self.type]
```
cquiroz, line 28: Commented code?

modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
```diff
@@ -10,16 +10,14 @@ import coulomb.cats.implicits._
...
 import eu.timepit.refined.types.numeric.PosBigDecimal
 import lucuma.core.math.BrightnessUnits._
-import lucuma.core.math.Wavelength
 import lucuma.core.math.dimensional._
 import lucuma.core.math.units._
 import monocle.Focus
 import monocle.Lens
 
 final case class EmissionLine[T](
-  wavelength: Wavelength,
```
cquiroz, line 20: So this is gone and instead it is in the map's index?

modules/tests/shared/src/test/scala/lucuma/core/model/SpectralDefinitionSuite.scala
```diff
@@ -5,84 +5,76 @@ package lucuma.core.model
...
-
-  val sd1Surface: SpectralDefinition[Surface] = {
-    val b: BandBrightness[Surface] =
-      BandBrightness.withDefaultUnits[Surface](BrightnessValue.fromDouble(10.0), Band.R)
 
+  val sd1Surface: SpectralDefinition[Surface] =
     SpectralDefinition.BandNormalized(
       UnnormalizedSED.StellarLibrary(StellarLibrarySpectrum.A0I),
-      SortedMap(b.band -> b)
+      SortedMap(Band.R -> Band.R.brightnessMeasure[Surface](BrightnessValue.fromDouble(10.0)))
```
swalker2m, line 50: Is the `Band.R.brightnessMeasure` just used to pick up the default units for `R` bands?  You could just as well do `Band.R` -> `Band.J.brightnessMeasure[Surface](BrightnessValue.fromDouble(10.0))` in terms of what gets created in the end?

cquiroz (approved):
Some small comments but LGTM

modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
rpiaggio: Exactly

modules/tests/shared/src/test/scala/lucuma/core/model/SpectralDefinitionSuite.scala
rpiaggio: Yes, it's just used to pick up the default units, and yes, nothing stops us from using the default units from another band.

Maybe we should the remove the `brightnessMeasure` convenience method?

We could still create the brightness with `Band.R.defaultUnits[Surface].withValueTagged(BrightnessValue.fromDouble(10.0))`, where it is more explicit that we are just using the band's default units.

swalker2m: Yes, that sounds good.  I was confused at first when I saw it, assuming that `Band.R` was somehow being associated with the measure being created there.

PR #455, opened 2022-01-07
Remove UC band

cquiroz:
After some discussions with science it was suggested we could drop UC.
ITC cannot handle UC and it is only relevant to the UCAC catalog while we'll be using GAIA in GPP


swalker2m (approved):
Always good to remove something, however small. :+1:

PR #457, opened 2022-01-11
Measure + Units ~show~/display, test suit for Units, minor renamings

rpiaggio:
- Add `Display[Measure]`, `Show[Measure]` and `Show[Units]`.
- Add `UnitTest`.
- Minor renamings that were pending (`Units` was previously called `UnitType`).

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
@@ -9,6 +9,8 @@ import coulomb._
 import monocle.Focus
 import monocle.Lens
 import shapeless.tag
+import _root_.cats.Show
+import lucuma.core.util.Display
 
 /**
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
```diff
@@ -9,6 +9,8 @@ import coulomb._
 import monocle.Focus
 import monocle.Lens
 import shapeless.tag
+import _root_.cats.Show
```
cquiroz, line 12: I thought we weren't really using `Show` anymore

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/measure.scala
rpiaggio: I'll remove it. That `_root_` crept in anyway.

PR #458, opened 2022-01-12
Display[Measure] derived from value Display.

rpiaggio:
Make `Display[Measure[N]]` invoke `Display[N]` instead of `N.toString`.


cquiroz (approved):
LGTM

PR #459, opened 2022-01-14
Display tagged Units and Measure.

rpiaggio:
- `Display[Measure[N]]`:
  - Derive instance from `Display[N]`
- `Display[Units Of T]` and `Display[Measure[N] Of T]`:
  - Support these tagged variants.


cquiroz (approved):
LGTM

PR #462, opened 2022-01-18
baseRa and baseDec optics in SiderealTracking

swalker2m:
Very minor update to `SiderealTracking` to add `baseRa` and `baseDec`.  When working directly with a `SiderealTracking` these are useful and otherwise have to be typed in.


cquiroz (approved):
LGTM

PR #465, opened 2022-01-28
Update to sbt-lucuma 0.6.0

armanbilge:
(Opening for CI).


.scalafmt.conf
```diff
@@ -0,0 +1,3 @@
+version = "3.3.3"
+runner.dialect = scala213source3
+project.includePaths = [] # disables formatting
```
swalker2m, line 3: :+1:

armanbilge:
Looks like you may have broken bincompat against 0.24.x. Is the next release going to be 0.25?
```
[error] lucuma-core: Failed binary compatibility check against edu.gemini:lucuma-core_2.13:0.24.0 (e:info.versionScheme=early-semver)! Found 4 potential problems
[error]  * abstract synthetic method lucuma$core$model$SiderealTrackingOptics$_setter_$baseRa_=(monocle.PLens)Unit in interface lucuma.core.model.SiderealTrackingOptics is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("lucuma.core.model.SiderealTrackingOptics.lucuma$core$model$SiderealTrackingOptics$_setter_$baseRa_=")
[error]  * abstract synthetic method lucuma$core$model$SiderealTrackingOptics$_setter_$baseDec_=(monocle.PLens)Unit in interface lucuma.core.model.SiderealTrackingOptics is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("lucuma.core.model.SiderealTrackingOptics.lucuma$core$model$SiderealTrackingOptics$_setter_$baseDec_=")
[error]  * abstract method baseRa()monocle.PLens in interface lucuma.core.model.SiderealTrackingOptics is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("lucuma.core.model.SiderealTrackingOptics.baseRa")
[error]  * abstract method baseDec()monocle.PLens in interface lucuma.core.model.SiderealTrackingOptics is present only in current version
[error]    filter with: ProblemFilters.exclude[ReversedMissingMethodProblem]("lucuma.core.model.SiderealTrackingOptics.baseDec")
```

swalker2m:
> Looks like you may have broken bincompat against 0.24.x. Is the next release going to be 0.25?

Confusion.  I added a couple of optics (but didn't remove anything) in #462.  The error seems to be related.

This check is new.  What should be done now when we make a change?  Do we need to publish a new release immediately?

tpolecat:
What you do is set `tlBaseVersion = "0.25"` and then your next tag will need to be `0.25.0`.

.scalafmt.conf
cquiroz: 😢

tpolecat:
So we should start getting warnings (errors?) from sbt now, if we pull in incompatible versions of core from two directions.

build.sbt
```diff
@@ -107,10 +97,10 @@ val MUnitFlakyOK   = sys.env.get("MUNIT_FLAKY_OK") match {
 lazy val tests = crossProject(JVMPlatform, JSPlatform)
   .crossType(CrossType.Full)
   .in(file("modules/tests"))
+  .enablePlugins(NoPublishPlugin)
```
cquiroz, line 100: Is this typelevel?

tpolecat: Yes.

armanbilge:
Yep, they will be errors.

build.sbt
armanbilge: Yes.
https://github.com/typelevel/sbt-typelevel/blob/main/no-publish/src/main/scala/org/typelevel/sbt/NoPublishPlugin.scala

armanbilge:
> @armanbilge the build now sets the version policy for the artfact to "early-semver", correct?

Yes that's right. version policy is early-semver and its enforced by MiMa using git tags to determine previous releases.

armanbilge:
Ok this PR is ready. Let me know if you have questions.

tpolecat:
We probably don't want to depend on the plugin snapshot long-term, so assuming this works and everything gets published ok let's go ahead and tag a new version of `sbt-lucuma` and update this again. Does that seem sensible?

armanbilge:
Yup 👍 sounds good. I'll work on your other builds (I'm working on observe rn) and continue upstreaming stuff into sbt-lucuma. Btw it would be good to get the formatting PR in as well before tagging.

PR #468, opened 2022-02-06
Spire coulomb

cquiroz:
Updates both spire and coulomb and removes the use of `scalatic` which I think came transitively


tpolecat (approved):
Seems fine.

PR #470, opened 2022-02-08
HashWithOrder

swalker2m:
This is probably a "don't do that" case (as opposed to "use `traverse`").  Nevertheless, I'm opening a PR to get the "don't do that" feedback.

I'd like to, ultimately, hash ITC inputs for the purpose of caching them.  Should we be defining `cats.Hash` instances at least for the things that will be used in this way?  If so, things for which there is also an `Order` instance should be problematic since both `Hash` and `Order` extend `Eq`.  Hence, `HashWithOrder` :-/

So, I shouldn't do this, but any advice on what I should do instead?

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
@@ -3,22 +3,21 @@
 
 package lucuma.core.math
 
-import cats.Order
 import cats.Show
 import coulomb._
-import coulomb.cats.implicits._
 import eu.timepit.refined._
 import eu.timepit.refined.auto._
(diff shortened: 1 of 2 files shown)


swalker2m:
Looking at cats itself, it seems there are individual `Order` classes for all the standard things.  Like

```
class IntOrder extends Order[Int] with Hash[Int] with IntBounded with IntEnumerable { self =>
```

I suppose I could just make a `class WavelengthOrder extends Order[Int] with Hash[Int]` along the same lines and an implicit value

```
implicit val orderWavelength: Order[Wavelength] with Hash[Wavelength] = new WavelengthOrder
```

 instead of adding a dubious `HashWithOrder`?

PR #471, opened 2022-02-09
Add name to user defined SED

cquiroz:
TIL user defined SEDs should have a name


swalker2m (approved):
Code looks good.  👍 this will have to bubble through everything of course

Why do they need a name?  What will we do with it?  Are they expecting to reuse them across targets?

cquiroz:
All good questions, I guess the idea is to upload a file name and give it a name so you could refer it in other places

cquiroz:
This is not urgent. Perhaps we should review it when we really implement iTC support for user data

swalker2m:
> Perhaps we should review it when we really implement iTC support for user data

Yeah it seems like the model may have to change more if we want to reference a shared set of flux densities.

cquiroz:
Yeah, let's close this

PR #479, opened 2022-02-22
Delete unused enum

swalker2m:
I noticed that we aren't using this and suspect it corresponds to a previous incarnation of the target definition world.  Is it okay to delete?


rpiaggio (approved):
LGTM

PR #480, opened 2022-02-23
Black body temp as Int, remove unused class, rename

rpiaggio:
Some model fine tuning:
- Black body temperature as `PosInt` instead of `PosBigDecimal`.
- Remove unused `GaussianSource` class.
- Rename `UnnormalizedSpectralEnergyDistributionSuite` -> `UnnormalizedSEDSuite`


swalker2m (approved):
Is good. :+1:

PR #487, opened 2022-03-11
Add gaia bands

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/enum/Band.scala
```diff
@@ -276,9 +276,39 @@ object Band {
...
+      )
+
+  /** @group Constructors */
+  case object GaiaGrp
+      extends BandWithDefaultUnits[VegaMagnitude, VegaMagnitudePerArcsec2](
+        "GaiaGrp",
+        "grp",
+        "Grp",
+        Wavelength(845000),
+        410.withRefinedUnit[Positive, Nanometer] // 640-1050
```
swalker2m, line 306: Where does `410` come from?

modules/core/shared/src/main/scala/lucuma/core/enum/GuideProbe.scala
```diff
@@ -0,0 +1,49 @@
...
+package enum
+
+import cats.syntax.eq._
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for Guiding probes
+ * @group Enumerations (Generated)
+ */
+sealed abstract class GuideProbe(val tag: String) extends Product with Serializable
```
swalker2m, line 15: There are more like categories of guide probes rather than particular guide probes. Maybe `GuideProbeCategory` or something?  I can't remember why we need this.  What do we do with it?

swalker2m (approved):
Looks good.  :+1:

modules/core/shared/src/main/scala/lucuma/core/enum/GuideProbe.scala
swalker2m: This is orthogonal to the GAIA bands, correct?

modules/core/shared/src/main/scala/lucuma/core/enum/Band.scala
cquiroz: 1050-640, that's what I understand as the width

swalker2m: :+1:

modules/core/shared/src/main/scala/lucuma/core/enum/GuideProbe.scala
cquiroz: AGS indexes the search params by guide probe type. I'm making a few non-trivial changes to ags on how the searches are conducted, maybe I could keep this enum in AGS until I'm sure we are going to need it

PR #489, opened 2022-03-14
Rename Gaia bands.

rpiaggio:


swalker2m (approved):
Was this prompted by a request from science?  At any rate, it seems good to me.  :+1:

cquiroz (approved):
LGTM

rpiaggio:
> Was this prompted by a request from science?

It all started because `GaiaG`'s `shortName` clashed with one of the Sloan bands. This is what's displayed to the user when selecting a band, so the idea is to bring it as close as possible to the names used by science, which actually have `BP` and `RP` as a subscript. Maybe at some point we can make the UI display as a subscript whatever's followed by the `_`. For the API/tag, `G` already seems to stand for Gaia, so `GaiaG` seemed duplicate.

It'd be a good idea to have this validated by science, I'll ask them.

andrewwstephens:
:+1:

PR #490, opened 2022-03-15
Add ConstraintSet, ElevationRange.

rpiaggio:


.scalafmt.conf
```diff
@@ -1,2 +1 @@
-include ".scalafmt-common.conf"
-project.includePaths = [] # disables formatting
+include ".scalafmt-local.conf"
```
rpiaggio, line 1: Whoever whants to enable scalafmt can just create a local `.scalafmt-local.conf` file with a single line containing `include ".scalafmt-common.conf"`.

modules/core/shared/src/main/scala/lucuma/core/model/ConstraintSet.scala
```diff
@@ -0,0 +1,78 @@
...
+import monocle.Lens
+import monocle.Optional
+
+/** Constraints for an observation. */
+final case class ConstraintSet(
+  imageQuality:    ImageQuality,
+  cloudExtinction: CloudExtinction,
+  skyBackground:   SkyBackground,
+  waterVapor:      WaterVapor,
+  elevationRange:  ElevationRange
```
cquiroz, line 25: Do we ever let not having an elevation range? Or do we have the equivalent to `Any` range

modules/core/shared/src/main/scala/lucuma/core/model/ElevationRange.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  final case class AirMass private (
+    min: AirMass.DecimalValue,
+    max: AirMass.DecimalValue
+  ) extends ElevationRange
+
+  final case class HourAngle private (
+    minHours: HourAngle.DecimalHour,
+    maxHours: HourAngle.DecimalHour
+  ) extends ElevationRange
```
swalker2m, line 27: We don't need the `sealed abstract case class` trick to prevent using the `copy` constructor anymore correct?

swalker2m (approved):
Looks good to me. :+1:

modules/core/shared/src/main/scala/lucuma/core/model/ConstraintSet.scala
swalker2m: Hmm good question. `Airmass.Default` (1-2) seems fine to me.

modules/core/shared/src/main/scala/lucuma/core/model/ElevationRange.scala
rpiaggio: Correct. And we can still use the `copy` method within the companion object, eg from lenses.

PR #493, opened 2022-03-30
Add Point shape

cquiroz:
This is the first of one of a bunch of PRs towards AGS.
I found it useful to be able to define a single point geometry, specifically to designate the location of a base target

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/GmosScienceAreaGeometry.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/GmosScienceAreaGeometry.scala
@@ -16,6 +16,13 @@ import lucuma.core.math.syntax.int._
   */
 object GmosScienceAreaGeometry {
 
+  // base target
+  def base =
+    ShapeExpression.point(Offset.Zero)
+
+  def baseAt(posAngle: Angle, offsetPos: Offset) =
+    base ↗ offsetPos ⟲ posAngle
+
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/lucuma/core/geom/GmosScienceAreaGeometry.scala
```diff
@@ -16,6 +16,13 @@ import lucuma.core.math.syntax.int._
   */
 object GmosScienceAreaGeometry {
 
+  // base target
+  def base =
+    ShapeExpression.point(Offset.Zero)
```
swalker2m, line 21: It's trivial but maybe add the `: ShapeExpression` type annotation to these?  I thought that was checked somewhere.

modules/core/shared/src/main/scala/lucuma/core/geom/GmosScienceAreaGeometry.scala
```diff
@@ -16,6 +16,13 @@ import lucuma.core.math.syntax.int._
...
+
+  def baseAt(posAngle: Angle, offsetPos: Offset) =
```
swalker2m, line 23: By definition the base is always at `(0, 0)` so maybe `def pointAt` would be an appropriate name?

swalker2m (approved):
Seems good to me. :+1:

PR #494, opened 2022-03-31
Add trigonometric functions to angle

cquiroz:
The rationale for this change is to remove a step where you need to look exactly what units needs the param when you call sine/cos/tan


swalker2m (approved):
Seems like a good idea to me.  :+1:

PR #495, opened 2022-03-31
Add a diff method for coordinates

cquiroz:
The current offset functions on `Coordinates` are for a planar geometry but we often need them on a spherical one.
This PR removes the old ones and adds one taken from the OT


modules/core/shared/src/main/scala/lucuma/core/math/Coordinates.scala
```diff
@@ -11,39 +11,50 @@ import lucuma.core.syntax.all._
...
+      val acosValue = acos(cosd)
+      if (acosValue.isNaN) 0.0 else acosValue
     }
 
-  /**
-   * Offset these `Coordinates` by the given deltas. If the declination crossed a pole the right
-   * ascension will have been flipped 180°.
-   * @group Operations
-   */
-  def offset(dRA: HourAngle, dDec: Angle): Coordinates =
```
swalker2m, line 38: I think `offset` and `offsetWithCarry` are correct.  It adds an offset in `HourAngle` so it's correct at any declination, though the actual angular separation will differ according to declination.  At any rate, maybe we don't need this method at the moment but I would keep it and `offsetWithCarry`.

swalker2m (commented):
Looks great, thanks for fixing this!  I would just vote to keep the `offset` methods as well.

modules/core/shared/src/main/scala/lucuma/core/math/Coordinates.scala
swalker2m: I guess I could be swayed that this calculation, though correct, is never going to be useful, if anyone disagrees.

cquiroz: I thought about keeping but renaming it as `planarOffset` and add another `offsetBy(posAngle, offset): Coordinates` method

swalker2m (approved):
Looks good. :+1:

PR #498, opened 2022-04-01
Migrate Uid from lucuma-tmp-api

rpiaggio:
Migrates `Uid` from `lucuma-tmp-api` and defines the model `WithUid` instances.

I had to remove the `random` methods in order not to introduce a dependency on `cats-effect`.

Also: organize imports.


swalker2m (approved):
:+1:

PR #500, opened 2022-04-05
BrightnessValue -> BigDecimal

rpiaggio:
For some brightness units (specifically, Watts and Ergs based units), the brightness value is usually expressed with very small values (eg `e-19`). The `BrightnessValue` class seemed designed to express only magnitude values. This PR deletes its and replaces its use for `BigDecimal`.

PD: Sorry @cquiroz for the work you put into https://github.com/gemini-hlsw/lucuma-core/pull/492, I didn't remember this pending requirement at the time of reviewing.


modules/core/shared/src/main/scala/lucuma/core/enum/GpiObservingMode.scala
```diff
@@ -23,8 +22,8 @@ sealed abstract class GpiObservingMode(
   val apodizer:             Option[GpiApodizer],
   val fpm:                  Option[GpiFPM],
   val lyot:                 Option[GpiLyot],
-  val brightLimitPrism:     Option[BrightnessValue],
-  val brightLimitWollaston: Option[BrightnessValue],
+  val brightLimitPrism:     Option[BigDecimal],
+  val brightLimitWollaston: Option[BigDecimal],
```
rpiaggio, line 26: We could rename `BrightnessValue` to `MagnitudeValue` and keep it only for this case.

cquiroz: I'm not sure I follow

cquiroz (approved):
This probably makes sense. LGTM

PR #502, opened 2022-04-07
Bounding box

cquiroz:
Adds a method to get a box of offsets for a given shape

image


cquiroz:
I'll rework this a bit

PR #504, opened 2022-04-08
Bounding box

cquiroz:
This is a rework of #502 where we can create a `BoundingBox` as a `ShapeExpression`
The box is now defined by two corner offsets and a function to calculate the widest side (as an angle) is included

Also a few tests are added.

The image below shows the box on the gmos shapes

image

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/BoundingOffsets.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/BoundingOffsets.scala
@@ -31,7 +31,7 @@ final case class BoundingOffsets(topLeft: Offset, bottomRight: Offset) {
    * |---|
    * |   |
    * |---|
-   * p3 p4
+   * p4 p3
    * }}}
    */
   def vertices: (Offset, Offset, Offset, Offset) =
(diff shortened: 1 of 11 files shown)


modules/core/shared/src/main/scala/lucuma/core/geom/BoundingOffsets.scala
```diff
@@ -0,0 +1,48 @@
...
+   * {{{
+   * p1 p2
+   * |---|
+   * |   |
+   * |---|
+   * p3 p4
+   * }}}
+   */
+  def vertices: (Offset, Offset, Offset, Offset) =
+    (topLeft, Offset(bottomRight.p, topLeft.q), bottomRight, Offset(topLeft.p, bottomRight.q))
```
swalker2m, line 38: I think maybe the order here doesn't match the comment.  It looks like `(p1, p2, p4, p3)`?

swalker2m (commented):
This looks good.  For context, what do we plan to do with this?

modules/core/shared/src/main/scala/lucuma/core/geom/BoundingOffsets.scala
cquiroz: Thanks, the comment is wrong. I wanted to keep the vertices on clockwise order

cquiroz:
I'm going to use this to create adql queries with a given geometry. e.g. I can calculate the radius of a shape and do a query there. Eventually more complex geometries would be possible

PR #505, opened 2022-04-14
Add geometry for the ags field in gmos and group all the gmos geometries

cquiroz:
This PR adds a geometry that includes the area we should read candidate targets for AGS

image

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosPatrolArea.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosPatrolArea.scala
@@ -1,0 +1,45 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.geom.gmos
+
+import lucuma.core.geom.ShapeExpression
+import lucuma.core.geom.syntax.all._
+import lucuma.core.math.Angle
+import lucuma.core.math.Offset
+import lucuma.core.math.syntax.int._
+
+/**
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosPatrolArea.scala
```diff
@@ -0,0 +1,45 @@
...
+import lucuma.core.geom.syntax.all._
+import lucuma.core.math.Angle
+import lucuma.core.math.Offset
+import lucuma.core.math.syntax.int._
+
+/**
+  * GMOS area that could be reachable by the patrol arm
+  * https://www.gemini.edu/instrumentation/gmos/capability#Guiding
+  */
+trait GmosPatrolArea {
```
swalker2m, line 16: This is tied to the OIWFS probe arm, maybe it should be an attribute of that like `patrolField` is?

Also I worry that "patrol area" sounds too much like "patrol field" and this is really a different thing.  Maybe call it `patrolArm.queryArea` or something?  Or maybe `patrolArm.candidateArea`?  Does that make sense?

swalker2m (commented):
Is this idea to turn this `Area` into an ADQL query?

cquiroz:
Yes we'd query the area for the candidates

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosPatrolArea.scala
cquiroz: I struggled with the names. I originally called it `agsField` though that maybe missleading too. `candidateArea` sounds good

swalker2m (approved):
Looks good to me. :+1:

PR #508, opened 2022-04-21
Ghost apply enum

cquiroz:
Note this is based on version 0.10.1 to make it compatible with the seqexec


cquiroz:
This can be closed after #771

PR #509, opened 2022-04-22
Rename GMOS Disperser => Grating

swalker2m:
A simple renaming, via suggestion from Andy.  We could get away with just renaming it in the API but it seems better to be consistent.


cquiroz (approved):
LGTM

PR #513, opened 2022-04-27
Partial sequence model.

rpiaggio:
Model for manual and generated sequences.

Past `Visit`s are not modeled yet.


modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+package lucuma.core.model.sequence
+
+import cats.Eq
+import cats.syntax.all._
+import lucuma.core.enum._
+import monocle.Focus
+import monocle.Lens
+import monocle.Prism
+import monocle.macros.GenPrism
+import org.typelevel.cats.time._
```
cquiroz, line 13: Is it for the `Eq` instances?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+
+import java.time.Duration
+
+sealed trait InstrumentConfig
+object InstrumentConfig {
+  final case class GmosNorth(
+    exposure: Duration,
```
cquiroz, line 20: Are we in general leaning to `FiniteDuration` for exposure times? @swalker2m

modules/tests/shared/src/test/scala/lucuma/core/model/IdsSuite.scala
```diff
@@ -12,12 +12,12 @@ final class IdsSuite extends DisciplineSuite {
   import ArbGid._
   import ArbUid._
 
-  checkAll("Atom.Id", UidTests[Atom.Id].uid)
+  // checkAll("Atom.Id", UidTests[Atom.Id].uid)
```
toddburnside, line 15: Commented code here and below.

toddburnside (approved):
Looks good to the best of my knowledge. Of course, that knowledge is pretty limited on what this stuff means...

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
```diff
@@ -0,0 +1,116 @@
...
+import monocle.Focus
+import monocle.Lens
+import monocle.Prism
+import monocle.macros.GenPrism
+import org.typelevel.cats.time._
+
+import java.time.Duration
+
+sealed trait InstrumentConfig
+object InstrumentConfig {
```
swalker2m, line 18: The static config is also instrument config, just unchanging from step to step.  I guess that's why I called this the dynamic config as opposed to the static config.  I'm not opposed to this name, but I would like to know your thoughts on the change.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/ExecutionSequence.scala
```diff
@@ -0,0 +1,62 @@
...
+import cats.syntax.all._
+import monocle.Focus
+import monocle.Lens
+import monocle.Prism
+import monocle.macros.GenPrism
+
+sealed trait ExecutionSequence {
+  val nextAtom: Atom
+  val possibleFuture: List[Atom]
+}
```
swalker2m, line 16: This pattern seems cleaner that what I did, where everything is parameterized by static and dynamic type all the way to the top.  I think what stopped me from going this way was that given an `ExecutionSequence` as defined above you have to match on the instances to do anything with it.  Otherwise the `Atom`s could in theory be from a mix of instruments.  This does look a lot cleaner though, so I'm very much okay with it if that doesn't seem like a problem to you.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Sequence.scala
```diff
@@ -0,0 +1,57 @@
...
+import monocle.Lens
+import monocle.Prism
+import monocle.macros.GenPrism
+
+sealed trait Sequence {
+  val atoms: List[Atom]
+  val time: StepTime
+}
+object Sequence       {
+  final case class GmosNorth(atoms: List[Atom.GmosNorth], time: StepTime) extends Sequence
```
swalker2m, line 18: `StepTime` will be derived by summing the atom step times.  I think we should keep it separate in the model?  I combined them in the API query results because I assumed the caller would just want to consume it.  Maybe that was wrong.

swalker2m (commented):
This looks very good, very clean to me.  I have a few comments I'm hoping to get your feedback on, but overall it looks great!  Thanks for doing this.

By the way I agree with putting this in core but that wasn't a universal opinion at one point.  I'm not sure if there is complete consensus on how complete the core model should be, but I don't have a good alternative.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
rpiaggio: Yes, it brings in `Eq[Duration]`.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
rpiaggio: `FiniteDuration` extends `scala.concurrent.duration.Duration` class which states that `This class is not meant as a general purpose representation of time, it is optimized for the needs of scala.concurrent.`.

So `java.time.Duration` seems approriate for modelling things.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
```diff
@@ -0,0 +1,69 @@
...
+import lucuma.core.util.WithUid
+import monocle.Focus
+import monocle.Lens
+import monocle.Prism
+import monocle.macros.GenPrism
+
+sealed trait Atom {
+  val id: Atom.Id
+  val steps: List[Step]
+  val time: StepTime
```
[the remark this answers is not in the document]
rpiaggio: Is this something that will be read from the API and displayed in the UIs? If so, we need a placeholder for it, which was the intent of these classes. Maybe we should model it some other way?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
rpiaggio: I'm still trying to grasp what everything means TBH. In this case I took the name from the property where it's used in the API. If I understand correctly, it would be better suited to call it `DynamicConfig` ?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Sequence.scala
rpiaggio: Ok, let's not have redundant information in the model. We could substitute it here for a method that computes it locally.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/ExecutionSequence.scala
rpiaggio: I struggled with this. I started by type-parametrizing, but that's unsafe to check at runtime, so I went with pattern matching, and then we can decide what we are dealing with at any point in the hierarchy.

I agree that the trait doesn't prevent mixed instances in the properties, but the types are fixed in the concrete classes, so mixed instances are not buildable. If it makes things clearer though, I think we could remove the general properties from the trait and just leave them in the concrete classes.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
swalker2m: I think when this exact point came up before the answer was along the lines of "well, everyone else uses it this way".  I'm fine with switching to `Duration` though.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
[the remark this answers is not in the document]
swalker2m: I'm not sure.  I probably was more focused with what a playground user typing things in would see and want in the query than in what the UI needs.  What would happen in the UI if you reorder the steps or, say, change a filter.  That could cause step times to change along with their sums all the way to the top.  Would you just reread the sequence from the model? Or would you compute it locally and ignore the values from the query result?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
swalker2m: In the API it is down in the step. Every step has a dynamic instrument config and I think I chose `instrumentConfig` to distinguish it from the non-instrument step configuration (gcal setup, offset position, etc.). Maybe I should rename it `dynamicConfig` there too.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/ExecutionSequence.scala
swalker2m: > I went with pattern matching, and then we can decide what we are dealing with at any point in the hierarchy

I like it. Let's go with it.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Sequence.scala
rpiaggio: In light of this, I wonder if we should completely remove `Sequence` and replace `Sequence.<Instrument>` with `List[Atom.<Instrument>]` everywhere. WDYT?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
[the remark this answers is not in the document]
rpiaggio: If we can compute it locally from the steps, then we should remove it from here. I'll assume that's the case and go ahead and remove it.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentConfig.scala
rpiaggio: Ok, for the moment I left `instrumentConfig: DynamicConfig`. LMK if you plan to change it so we name it accordingly here.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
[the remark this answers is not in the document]
swalker2m: It'll definitely be computed from the steps.  In the case of GMOS though, the readout time for example is a table with a key formed from the amp count, amp gain, read mode, ROI, and x and y binning.  In other words, if you look at all the combinations it's kind of big.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Sequence.scala
swalker2m: That's probably fine.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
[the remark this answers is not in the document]
rpiaggio: I guess we can make `StepTime` a `Monoid`, but I'll do that in another PR.

PR #517, opened 2022-05-02
Add DatasetQaState

swalker2m:
I'll need a QA state in order to mark datasets good or bad and then adjust the sequence generation accordingly.


rpiaggio (approved):
LGTM

PR #484, opened 2022-03-04
WIP Scala 3

armanbilge:
Currently only build hacking and file shuffling.

PR #520, opened 2022-05-10
Scala 3

armanbilge:
Tackling https://github.com/gemini-hlsw/lucuma-core/pull/484 from another angle.


modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
```diff
@@ -16,33 +18,33 @@ final class UnitsSuite extends FunSuite {
 
   test("Sanity tests") {
     assert(Rational.one.withUnit[Angstrom] === Rational(1, SafeLong(10).pow(10)).withUnit[Meter])
```
armanbilge, line 20: @erikerlandson currently getting a compile error for this:
```
[error] -- Error: /workspace/lucuma-core/modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:20:96 
[error] 20 |    assert(Rational.one.withUnit[Angstrom] === Rational(1, SafeLong(10).pow(10)).withUnit[Meter])
[error]    |                                                                                                ^
[error]    |                             unit type Meter not convertable to Angstrom
[error]    | This location contains code that was inlined from UnitsSuite.scala:20
[error]    | This location contains code that was inlined from unit.scala:27
[error]    | This location contains code that was inlined from policy.scala:23
[error] one error found
```

`Angstrom` is defined here:
https://github.com/gemini-hlsw/lucuma-core/blob/f6da50b2326b45448a8cd7c7a966333b8e68f09f/modules/core/shared/src/main/scala/lucuma/core/math/units.scala#L31-L32

I assume it's a simple and stupid mistake 😅 thanks!

cquiroz:
@armanbilge How are we doing with refined's macros? we use `auto` on quite a few places

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
erikerlandson: @armanbilge   my guess is this:
`import coulomb.units.si._`

should be `import coulomb.units.si.{*, given}`

In scala 3 givens aren't automatically imported via `*` or I assume `_`

armanbilge: Right, I've been trying to fix those. But is that import needed at the definition-site or the use-site?

armanbilge:
@cquiroz I haven't been able to test that yet. I think we will need to get the units stuff compiling first, because it is entangled with that.

Note that I make use of `auto` in one of my Scala 3 projects without problem.
https://github.com/armanbilge/cheshire/blob/1aba13367bc55aae8b47f53afe0374e9413d64c4/likelihood-laws/src/main/scala/cheshire/likelihood/laws/testkit.scala#L24

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
erikerlandson: it should not be needed for the definition

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
```diff
@@ -3,10 +3,12 @@
...
-import coulomb._
-import coulomb.mks._
-import coulomb.si.Meter
-import lucuma.core.math.units._
+import coulomb.{*, given}
+import coulomb.policy.standard.given
+import coulomb.policy.spire.standard.given
+import coulomb.units.mks.{*, given}
+import coulomb.units.si.{*, given}
+import lucuma.core.math.units.{*, given}
```
armanbilge, line 11: @erikerlandson so then these should all be fine, right?

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
erikerlandson: but it is  needed here:
```
assert(Rational.one.withUnit[Angstrom] === Rational(1, SafeLong(10).pow(10)).withUnit[Meter])
```

because `===` needs to instantiate the conversion, and that requires the `given`s

erikerlandson: You need to import `given` here too:

https://github.com/gemini-hlsw/lucuma-core/blob/f6da50b2326b45448a8cd7c7a966333b8e68f09f/modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala#L11

armanbilge: I think you're looking at an old commit, see https://github.com/gemini-hlsw/lucuma-core/pull/520#discussion_r869613369.

erikerlandson: also, is this supposed to be `trait` ?
https://github.com/gemini-hlsw/lucuma-core/blob/f6da50b2326b45448a8cd7c7a966333b8e68f09f/modules/core/shared/src/main/scala/lucuma/core/math/units.scala#L22

armanbilge: Yes. Then it is inherited by an `object`.

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
erikerlandson: yes, those look right, EXCEPT that if you import `coulomb.policy.spire.standard.given` you should not also import `coulomb.policy.standard.given` - these are one case where they do not "union"

Also, it should be harmless but you only need to `import coulomb.*`

armanbilge: Thanks, I just pushed a commit with those changes but I still get the same error 😕

erikerlandson: :sob: 

at the moment I don't see what is wrong, although I have never tried using type aliases like `PicoMeter` in a definition - I would have assumed it should work but it might be worth trying `Pico * Meter` directly in the definition

erikerlandson: if that is a problem I might need to make more use of `tr.dealias`

erikerlandson: ugh.  I'll experiment with more `dealias` to see if that helps

build.sbt
```diff
@@ -1,18 +1,22 @@
...
+
 lazy val attoVersion           = "0.9.5"
 lazy val catsVersion           = "2.7.0"
 lazy val kindProjectorVersion  = "0.13.2"
 lazy val monocleVersion        = "3.1.0"
 lazy val scalaJavaTimeVersion  = "2.3.0"
 lazy val geminiLocalesVersion  = "0.7.0"
 lazy val jtsVersion            = "0.2.2"
-lazy val coulombVersion        = "0.5.8"
+lazy val coulombVersion        = "0.6-3cc9f7f-SNAPSHOT"
```
armanbilge, line 19:
```suggestion
lazy val coulombVersion        = "0.6-7c69de6-SNAPSHOT"
```

armanbilge: @erikerlandson thanks for the quick fix, it's working now! :)

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
```diff
@@ -17,33 +17,33 @@ final class UnitsSuite extends FunSuite {
...
-  //   assertEquals(1.withUnit[WattsPerMeter2].show, "1 W/m²")
-  //   assertEquals(1.withUnit[ErgsPerSecondCentimeter2].show, "1 erg/s/cm²")
-  //   assertEquals(1.withUnit[WattsPerMeter2Arcsec2].show, "1 W/m²/arcsec²")
-  //   assertEquals(1.withUnit[ErgsPerSecondCentimeter2Arcsec2].show, "1 erg/s/cm²/arcsec²")
-  // }
+  test("Check brightness unit names") {
+    assertEquals(1.withUnit[VegaMagnitude].show, "1 Vega mag")
+    assertEquals(1.withUnit[ABMagnitude].show, "1 AB mag")
+    assertEquals(1.withUnit[Jansky].show, "1 Jy")
+    assertEquals(1.withUnit[WattsPerMeter2Micrometer].show, "1 W/m²/µm")
```
armanbilge, line 31: @erikerlandson ok sorry, next dumb question. Any thoughts on this?
```
=> Diff (- obtained, + expected)
-1 W/(m^2 ((Micro * Meter)))
+1 W/m²/µm
```
@cquiroz how particular are you about the formatting of these?

cquiroz: We use them in the UI in some places

cquiroz: But we could find perhaps other ways to do it

build.sbt
erikerlandson: :tada:   thanks for exercising the code!

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
erikerlandson: that result looks kind of weird.  In general it is probably dangerous to assume too much about how `show` renders but I'd expect it to not be printing `Micro` and `Meter` out there.

erikerlandson: is micrometer also a type alias?

rpiaggio: The abbreviation is made explicit when defining the units. See https://github.com/gemini-hlsw/lucuma-core/blob/master/modules/core/shared/src/main/scala/lucuma/core/math/units.scala#L78

Maybe `WattsPerMeter2Micrometer` can be an opaque type in Scala 3?

erikerlandson: aha, `UnitStringDef` is no longer a thing in scala3 coulomb, however when defined using the new variation it should still be explicit.  And yes, `Micrometer` is a type alias.  I see, `WattsPerMeter2Micrometer` is not a `DerivedUnit`.

I think I know what's going on, it's something similar to canonical sig, but over in the `show` metaprogramming.

armanbilge: > however when defined using the new variation it should still be explicit

@erikerlandson sorry, wdym by this? :)

erikerlandson: @armanbilge  I was assuming `WattsPerMeter2Micrometer` was a `DerivedUnit`, where you can tell it what to use for an abbv.  But it is a `type = ` alias.

armanbilge: I see, so that would be the way to define it?

erikerlandson: I think I can provide decent support for type aliases.  If you want *total* control over how you define the abbreviations, then declaring it as a `DerivedUnit` is probably the best practice.

armanbilge: Thanks, that's really helpful :) @cquiroz I guess a `DerivedUnit` is probably the way to go here.

rpiaggio: @erikerlandson I'm kind of worried `UnitStringDef` doesn't exist (at least implicitly) in Scala 3. We rely on it to have a runtime representation of units. (see https://github.com/gemini-hlsw/lucuma-core/blob/master/modules/core/shared/src/main/scala/lucuma/core/math/dimensional/unit.scala).

Actually, it would help us a lot if coulomb supported runtime representation of units out of the box. We have cases in which users can specify units for a quantity and we want to store the quantity in the specified units + the actual units.

erikerlandson: @rpiaggio  I can support it, if people want one.   I was excited about being able to instantiate the strings directly in-line, without needing the `UnitString` as an intermediary, but any method that creates things in-line can also populate a corresponding implicit.

rpiaggio: Our runtime `Units` representation leverages the fact that `UnitString` is a "de facto" runtime representation. We don't need `UnitString` per se, we wouldn't need it if coulomb supported runtime units by design. Do you think that's a useful feature for coulomb?

erikerlandson:
There's something I'm curious about:  the point of supporting prefix-units was so this would not be necessary:
```
  type DeciArcSecond  = Deci %* ArcSecond
  type MilliArcSecond = Milli %* ArcSecond
  type MicroArcSecond = Micro %* ArcSecond
  ```
The idea was that one would not need (or want) to define dozens of extra unit types to express prefixes. Just write `Deci *ArcSecond`, `Milli * ArcSecond`, etc.

However I see you are doing a lot of it, so it makes me think `Deci * ArcSecond` is a problem

File: modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala (hunk 4 above)
Comment (erikerlandson) [reply modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:31]: @rpiaggio it is certainly possible to do.  My entire philosophy around coulomb is unit analysis as a compile-time type checking system, so I haven't given a lot of thought to it.  However it is in many ways easier to do it at run-time.

I think the tricky part would be to somehow leverage static unit definitions in a run-time system.  The only alternative would be two totally separate unit definition systems, which seems inelegant and wasteful.

Comment (erikerlandson) [reply modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:31]: An example of runtime units which work mostly like coulomb units do, but at runtime, is the python 'pint' package:
https://pint.readthedocs.io/en/0.6/

Comment (armanbilge) [reply modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:31]: > I think the tricky part would be to somehow leverage static unit definitions in a run-time system.  The only alternative would be two totally separate unit definition systems, which seems inelegant and wasteful.

@erikerlandson relevant reading I think: https://hirrolot.github.io/posts/why-static-languages-suffer-from-complexity

Comment (erikerlandson) [reply modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:31]: @armanbilge  that is a nice articulation of the problem, and gives me a sad  :joy: 

So in practical terms,
1. I can re-implement `UnitString`, and allow whatever your current scheme is to work as it was
2. A parallel runtime (pint-like?) system seems possible, but is obviously a longer-term project. 
3. Deriving a pint-style 'unit registry' from static unit types seems fruitful.

Comment (rpiaggio) [reply modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:31]: `UnitString` nicely materializes runtime representations of units defined in the type-level, or at least a starting point.

If you wish, you can integrate our `Units` and `Measure` classes into a module in coulomb, which would give it the ability to handle runtime units. (`Measure` is a version of `Quantity` with runtime units).

File: modules/core/shared/src/main/scala/lucuma/core/math/Constants.scala [hunk 5]
```diff
@@ -60,8 +55,7 @@ object Constants {
    */
 
   /** Speed of light in meters per second. Exact. */
-  val SpeedOfLight: Quantity[Int, MetersPerSecond] =
-    speedOfLightInVacuum[Int]
+  val SpeedOfLight: Quantity[Int, MetersPerSecond] = ???
```
armanbilge, line 58: @erikerlandson dumb question of the day :) I couldn't figure out what to put on the rhs here, or at least the necessary `import`s to make it happen. In fact, I couldn't even summon `constant[coulomb.rational.Rational, SpeedOfLight]` without it complaining about a lack of conversion from `Rational` to itself.

erikerlandson: the `constant` method assumes a non-truncating value conversion, so probably you will need to do something like this:
```scala
import coulomb.units.constants.{*, given}
val SpeedOfLight = constant[Double, SpeedOfLight].tToValue[Int]
```

because I made truncating conversions explicit-only via `tToValue`

erikerlandson: I sort of assumed anyone who was in the business of using the kind of values defined in `coulomb.units.constants` would be wanting fractional value types

armanbilge: D'oh, that's was a dumb one. Thanks, works nicely with spire `Rational` instead of `Double.`

Yeah, not sure how this particular API is used. I'll keep an eye out for it.

modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala
erikerlandson: Latest `scala3` branch adds a `ShowUnit` such that `summon[ShowUnit[U]].value` yields the unit string.
You can also define the unit strings for type aliases using `ShowUnitAlias`:

https://github.com/erikerlandson/coulomb/blob/scala3/core/src/test/scala/coulomb/quantity.scala#L558

armanbilge: @erikerlandson nice, thank you! So is that the new "blessed" way, instead of using derived units like I did in 963ee5927707d65bab2ecdf4b2edfa755531a7b9?

erikerlandson: The mental image I had in mind when designing coulomb was that units having a standard name, like `Angstrom` or `Jansky` or `Erg`, would be defined using `DerivedUnit`

Something like  `ErgsPerSecondCentimeter2Angstrom`  would just be unit expressions: `Erg / (Second * ((Centi * Meter) ^ 2) * Angstrom)` - I don't (personally) understand the point of defining a type name for that.

I guess I consider `DerivedUnit` to be "more blessed" than `type X = `, but ultimately it's a matter of preferred style.  I don't mind supporting `type X = ...`

modules/core/shared/src/main/scala/lucuma/core/math/ApparentRadialVelocity.scala
```diff
@@ -22,10 +22,10 @@ final case class ApparentRadialVelocity(cz: Quantity[BigDecimal, MetersPerSecond
   /**
     * Converts the apparent radial velocity to a Redshift
     */
-  def toRedshift: Redshift = Redshift((cz / SpeedOfLight).value)
+  def toRedshift: Redshift = Redshift((cz / constant[BigDecimal, SpeedOfLight]).value)
 
   override def toString =
-    s"ApparentRadialVelocity(${cz.to[Double, KilometersPerSecond].show})"
+    s"ApparentRadialVelocity(${cz.toUnit[KilometersPerSecond].toValue[Double].show})"
```
armanbilge, line 28: @erikerlandson I'm getting a compile error here, that I can't really understand.
```
Found:    (ApparentRadialVelocity.this.cz : 
  coulomb.Quantity[BigDecimal, lucuma.core.math.units.MetersPerSecond]
)
Required: qopaque$_this.Quantity²[BigDecimal, coulomb.units².accepted.Meter / 
  coulomb.units².accepted.Second
]

where:    Quantity  is a type in package object quantity$package which is an alias of [V, U] =>> coulomb.qopaque.Quantity[V, U]
          Quantity² is a type in object qopaque with bounds <: [V, U] =>> Any
          units     is a object in package lucuma.core.math
          units²    is a package in package coulomb
```

armanbilge:
@erikerlandson I wonder if there is a regression in the latest coulomb snapshot. In e8b98994ff89b271e3521c7f09e1f5c3f9dd5342 which uses https://github.com/erikerlandson/coulomb/commit/7c69de63b9060ad75d614c856b6d0b99e2a419f5 the lucuma-core tests compile and pass.

However, bumping to https://github.com/erikerlandson/coulomb/commit/a58b5c75dd33e67b617dd210ef0ac60d5927bd00 introduces compile errors. Even worse, those were my changes that added the Cats instances, but I can't see at all how they could cause this to happen 😓 

```
[error] -- Error: /workspace/lucuma-core/modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:19:96 
[error] 19 |    assert(Rational.one.withUnit[Angstrom] === Rational(1, SafeLong(10).pow(10)).withUnit[Meter])
[error]    |                                                                                                ^
[error]    |Cannot prove that coulomb.qopaque.Quantity[V, coulomb.units.mks.Meter]
[error]    |
[error]    |where:    V is a type variable with constraint >: spire.math.Rational
[error]    | =:= coulomb.qopaque.Quantity[spire.math.Rational, lucuma.core.math².units.Angstrom]
[error]    |
[error]    |where:    math  is a package in package spire
[error]    |          math² is a package in package lucuma.core
[error]    |.
[error] -- Error: /workspace/lucuma-core/modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:22:45 
[error] 22 |        .withUnit[Watt / (Meter ^ 2) * Hertz]
[error]    |                                             ^
[error]    |Cannot prove that coulomb.qopaque.Quantity[V, coulomb.units.mks.Watt / (coulomb.units.mks.Meter ^ 
[error]    |  (2 : Int)
[error]    |) * coulomb.units.mks.Hertz]
[error]    |
[error]    |where:    V is a type variable with constraint >: spire.math.Rational
[error]    | =:= coulomb.qopaque.Quantity[spire.math.Rational, lucuma.core.math².units.Jansky]
[error]    |
[error]    |where:    math  is a package in package spire
[error]    |          math² is a package in package lucuma.core
[error]    |.
[error] -- Error: /workspace/lucuma-core/modules/tests/shared/src/test/scala/lucuma/core/math/UnitsSuite.scala:24:90 
[error] 24 |    assert(Rational.one.withUnit[Erg] === Rational(1, SafeLong(10).pow(7)).withUnit[Joule])
[error]    |                                                                                          ^
[error]    |Cannot prove that coulomb.qopaque.Quantity[V, coulomb.units.mks.Joule]
[error]    |
[error]    |where:    V is a type variable with constraint >: spire.math.Rational
[error]    | =:= coulomb.qopaque.Quantity[spire.math.Rational, lucuma.core.math².units.Erg]
[error]    |
[error]    |where:    math  is a package in package spire
[error]    |          math² is a package in package lucuma.core
[error]    |.
[error] three errors found
[error] (testsJVM / Test / compileIncremental) Compilation failed
```

erikerlandson:
> However, bumping to [erikerlandson/coulomb@a58b5c7](https://github.com/erikerlandson/coulomb/commit/a58b5c75dd33e67b617dd210ef0ac60d5927bd00) introduces compile errors. Even worse, those were my changes that added the Cats instances, but I can't see at all how they could cause this to happen sweat

Hmm.  All the `coulomb` CI is passing, including with your cats PR.  Circumstantially it seems like that could have caused the problem but I also don't see why, just looking at it.

modules/core/shared/src/main/scala/lucuma/core/math/ApparentRadialVelocity.scala
armanbilge: @erikerlandson I also have a guess about this one. It looks like a repeat of the `export` issue, `coulomb.Quantity` is the exported version of `qopaque$_this.Quantity`.

erikerlandson: @armanbilge  It looks like it ought to work, but I'm wondering if it is another problem relating somehow to type aliases.  

The other thing that might be suspect is doing all that in `toString` - in my experience it often doesn't work well with implicits, although usually when that is happening it will complain about a bad method signature or missing `using` arguments.

erikerlandson: One other possibility is that you are not importing all the unit defs you need for scala to resolve the types all the way down to base-units.

armanbilge: @erikerlandson I was able to fix it by switching all `coulomb.*` imports to `import coulomb.qopaque.{Quantity, withUnit}` instead. I really think `export` just doesn't work.

erikerlandson: I still consider it an option to go back to `AnyVal`.  I liked the idea of an opaque type because it seemed like a better conceptual fit, but if it is going to cause these hard-to-predict and hard-to-diagnose failures then maybe it isn't a good trade.

erikerlandson: > What exactly was broken?

I no longer remember exactly but trying to put it at the top level of the package was failing to compile.  Possibly there is a way to make it work that I failed to see?

Since then I ran across another reason:  when putting the extensions at the top level, the compiler was failing to distinguish between something like `.show` in `Quantity` versus `DeltaQuantity`. Putting the defs and extensions under `qopaque` and `dqopaque` fixed that.

Except it's still breaking...

erikerlandson: it would be nice to minimize or eliminate boxing, but OTOH it's already boxing in a lot of places, and I am becoming concerned with the current opaque-type representation being a minefield of compilation problems.

erikerlandson: xref on opaque boxing: https://github.com/lampepfl/dotty/discussions/12011

erikerlandson: aside: my dream for `coulomb` is that - at the byte code level - the actual class `Quantity` never appears, or even the various type-classes - it all eventually compiles down to things like `x + c y`, where `c` is some conversion constant the compilation process derived and threw everything else away after it was finished with it.

My hope was that `opaque type` would move me closer to that dream, but the state of the jvm and scala compiler aren't there yet.

erikerlandson: If I want to make real progress on `coulomb` I simply don't have the spare time to be fighting obscure battles with the compiler on fundamental types like `Quantity` and its operations.

erikerlandson: Possibly I could just stop supporting the `export qopaque.Quantity` and friends, and you just have to `import coulomb.quantity.*` or something similar.  If that works it would be a reasonable compromise that keeps the opaque type.

armanbilge: Right, that's effectively what we are doing now anyway.

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -3,186 +3,170 @@
...
-  implicit val convAP: UnitConverter[PosInt, Angstrom, PosInt, Picometer] =
-    unsafePosIntConverter[Angstrom, Picometer]
-
-  implicit val convMP: UnitConverter[PosInt, Micrometer, PosInt, Picometer] =
-    unsafePosIntConverter[Micrometer, Picometer]
+  given rationalPosIntConverter: ValueConversion[PosInt, Rational] = Rational(_)
+
+  inline given [UF, UT]: TruncatingUnitConversion[PosInt, UF, UT] = v =>
+    refineV[Positive](coulomb.conversion.standard.unit.ctx_TUC_Int(v))
+      .getOrElse(refineMV[Int, Positive](1))
```
armanbilge, line 140: @erikerlandson My own poor attempt at trying to add some support for refined ... I imagine you can do some compile-time magic here :)

modules/core/shared/src/main/scala/lucuma/core/math/ApparentRadialVelocity.scala
erikerlandson: I'm reading this:
https://dotty.epfl.ch/docs/reference/contextual/extension-methods.html#translation-of-calls-to-extension-methods

And my interpretation is that I should be able to define my extension inside companion `object Quantity`,
and those methods should be visible if I `import coulomb.Quantity`, however it doesn't seem to be working.

Wondering if it is yet another situation where `opaque` has bugs

armanbilge: > And my interpretation is that I should be able to define my extension inside companion `object Quantity`,
> and those methods should be visible if I `import coulomb.Quantity`, however it doesn't seem to be working.

Huh, it _should_ work like that. Here's an example from one of my projects:
https://github.com/armanbilge/schrodinger/blob/398e21ceadd862a61252f69559b9b074f6e2582d/math/src/main/scala/schrodinger/math/LogDouble.scala#L23

erikerlandson: I'm getting problems around `def value`, where it seems to be confusing `V` with `Quantity[V, _]`

armanbilge: That's the same as the `log` method in mine:
https://github.com/armanbilge/schrodinger/blob/398e21ceadd862a61252f69559b9b074f6e2582d/math/src/main/scala/schrodinger/math/LogDouble.scala#L57

erikerlandson:
```
  |value value is not a member of coulomb.quantity$package.Quantity[Double, String]
```

getting errors like this.  I notice you put your LogDouble in a file called LogDouble.scala, and mine is in `quantity.scala` - I am wondering if I am "supposed" to use Quantity.scala

armanbilge: @erikerlandson can you put up a branch for me to look at?

erikerlandson: I feel like I'm randomly thrashing around.  something is wrong but I can't quite center on it

erikerlandson: > can you put up a branch for me to look at?
Yes, let me start over clean, and put stuff in `Quantity` companion object, and I will push

build.sbt
```diff
@@ -18,7 +18,7 @@ lazy val monocleVersion        = "3.1.0"
 lazy val scalaJavaTimeVersion  = "2.3.0"
 lazy val geminiLocalesVersion  = "0.7.0"
 lazy val jtsVersion            = "0.2.2"
-lazy val coulombVersion        = "0.6-7e4eeeb-SNAPSHOT"
+lazy val coulombVersion        = "0.6-0c31769-SNAPSHOT"
```
armanbilge, line 21: @erikerlandson this works great 💯 thank you so much!

erikerlandson: :tada:

erikerlandson: @armanbilge  I'm wondering if this refactoring might fix the disambiguation of `===`

armanbilge:
Finally CI is green 🥳 

@erikerlandson Couldn't have done it without you ❤️ thank you for your tremendous effort on Coulomb over the last months. It's truly working beautifully here. What would you think about publishing an early milestone of Coulomb's scala3 branch, so we don't have to depend on a snapshot?

erikerlandson:
@armanbilge  sure, a milestone seems reasonable.  Barring any other issues that come up, I think the code for 0.6.0 is complete, it's down to writing proper docs.

erikerlandson:
@armanbilge @cquiroz   coulomb-0.6.0-M1 (and coulomb-units, coulomb-spire) should now exist.

cquiroz:
Thanks a lot @erikerlandson  , you've been extremely helpful  to get us this far. I really appreciate it

build.sbt
```diff
@@ -1,23 +1,29 @@
...
 lazy val attoVersion           = "0.9.5"
-lazy val catsVersion           = "2.7.0"
+lazy val catsVersion           = "2.8-3cb7639-SNAPSHOT"
 lazy val kindProjectorVersion  = "0.13.2"
 lazy val monocleVersion        = "3.1.0"
 lazy val scalaJavaTimeVersion  = "2.3.0"
 lazy val geminiLocalesVersion  = "0.7.0"
 lazy val jtsVersion            = "0.2.2"
-lazy val coulombVersion        = "0.5.8"
+lazy val coulombVersion        = "0.6-5c6079f-SNAPSHOT"
```
armanbilge, line 21:
```suggestion
lazy val coulombVersion        = "0.6.0-M1"
```

modules/core/shared/src/main/scala/lucuma/core/enums/GmosNorthGrating.scala
```diff
@@ -6,12 +6,14 @@ package core
...
-import coulomb._
+import coulomb.*
+import coulomb.policy.spire.standard.given
+import coulomb.syntax.*
 import eu.timepit.refined.types.numeric.PosInt
 import lucuma.core.math.Angle
 import lucuma.core.math.Coverage
 import lucuma.core.math.Wavelength
-import lucuma.core.math.units._
+import lucuma.core.math.units.{_, given}
```
cquiroz, line 16: `{*, given}`

modules/core/shared/src/main/scala/lucuma/core/enums/GmosNorthGrating.scala
```diff
@@ -34,7 +36,7 @@ sealed abstract class GmosNorthGrating(
 
   /** Compute the coverage of this disperser, given a central wavelength. */
   def coverage(λ: Wavelength): Coverage =
-    Coverage.centered(λ, simultaneousCoverage)
+    Coverage.centered(λ, simultaneousCoverage.tToUnit[Picometer])
```
cquiroz, line 39: This is an example of using `tToUnit`. I'm not sure what does it mean i practice to truncate

modules/core/shared/src/main/scala/lucuma/core/enums/ImageQuality.scala
```diff
@@ -3,37 +3,37 @@
...
-  case object PointOne     extends ImageQuality(refineMV[Positive](1).withUnit[DeciArcSecond])
-  case object PointTwo     extends ImageQuality(refineMV[Positive](2).withUnit[DeciArcSecond])
-  case object PointThree   extends ImageQuality(refineMV[Positive](3).withUnit[DeciArcSecond])
-  case object PointFour    extends ImageQuality(refineMV[Positive](4).withUnit[DeciArcSecond])
-  case object PointSix     extends ImageQuality(refineMV[Positive](6).withUnit[DeciArcSecond])
-  case object PointEight   extends ImageQuality(refineMV[Positive](8).withUnit[DeciArcSecond])
-  case object OnePointZero extends ImageQuality(refineMV[Positive](10).withUnit[DeciArcSecond])
-  case object OnePointFive extends ImageQuality(refineMV[Positive](15).withUnit[DeciArcSecond])
-  case object TwoPointZero extends ImageQuality(refineMV[Positive](20).withUnit[DeciArcSecond])
+  case object PointOne     extends ImageQuality(refineMV[Int, Positive](1).withUnit[DeciArcSecond])
```
cquiroz, line 28: Could this be `1.refined[Positive].withUnit...`

modules/core/shared/src/main/scala/lucuma/core/enums/ProposalClass.scala
```diff
@@ -0,0 +1,34 @@
+// Copyright (c) 2016-2022 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+sealed abstract class ProposalClass(val label: String) extends Product with Serializable
```
cquiroz, line 8: How is this new? I thought it would be present on the scala 2 version

modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
```diff
@@ -79,7 +81,7 @@ object Wavelength {
     * Construct a wavelength from a positive int
     * @group constructor
     */
-  def apply(picometers: PosInt): Wavelength =
+  def apply(picometers: PosInt)(using DummyImplicit): Wavelength =
```
cquiroz, line 84: What's `DummyImpicit`?

modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
```diff
@@ -104,27 +106,28 @@ object Wavelength {
    * Try to build a Wavelength with a value in nm in the range (0 .. 2147]
    * @group constructor
    */
-  def fromMicrometers(µm: Int): Option[Wavelength] =
-    refineV[Positive](µm).toOption.flatMap(µm =>
-      Option.when(µm.value <= MaxMicrometer)(Wavelength(µm.withUnit[Micrometer]))
-    )
+  def fromMicrometers(µm: Int): Option[Wavelength] = for {
+    q <- refineQV[Positive](µm.withUnit[Micrometer].tToUnit[Picometer]).toOption
```
cquiroz, line 110: What is `refineQV`?

modules/core/shared/src/main/scala/lucuma/core/math/dimensional/package.scala
```diff
@@ -23,14 +25,14 @@ package object dimensional {
...
       tagged
     }
   }
 
   // Default `TypeString` for a unit type.
-  implicit def typeStringFromUnitString[U](implicit ev: UnitString[U]): TypeString[U] =
-    new TypeString[U] {
-      val serialized: String = ev.abbv
-    }
+  // implicit def typeStringFromUnitString[U](implicit ev: ShowUnit[U]): TypeString[U] =
```
cquiroz, line 34: Commented code. Do we need this anymore?

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -3,189 +3,189 @@
 
 package lucuma.core.math
 
-import coulomb._
-import coulomb.accepted._
+import _root_.spire.math._
+import coulomb./
+import coulomb.^
+import coulomb.*
```
cquiroz, line 9: Duplicated

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -3,189 +3,189 @@
...
-          sys.error(s"Cannot convert exactly with coef ${cu.coef}")
+  given ValueConversion[NonNegInt, Double] = _.value.toDouble
 
-    }
+  given ValueConversion[Parallax.LongParallaxμas, Rational] = Rational(_)
 
-  // Implicit conversions that can be exact as Pico/Nano/Angstrom are multiples of 10
-  implicit val convNP: UnitConverter[PosInt, Nanometer, PosInt, Picometer] =
-    unsafePosIntConverter[Nanometer, Picometer]
+  inline given [UF, UT]: TruncatingUnitConversion[PosInt, UF, UT] = v =>
```
cquiroz, line 148: is this resolved?

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -3,189 +3,189 @@
...
+      
+  inline def refineQV[R]: RefineQV[R] = RefineQV()
+
+  final class RefineQV[P] {
+    inline def apply[V, U](q: Quantity[V, U])(using Validate[V, P]): Either[String, Quantity[V Refined P, U]] = {
+      refineV(q.value).map(_.withUnit[U])
+    }
+  }
+
+  // // This can build a converter for units that use PosInt but they are exact only
```
cquiroz, line 163: Commented code

modules/core/shared/src/main/scala/lucuma/core/model/ElevationRange.scala
```diff
@@ -28,7 +29,9 @@ object ElevationRange {
 
   object AirMass {
     val MinValue = BigDecimal(1.0)
+    given WitnessAs[MinValue.type, BigDecimal] = WitnessAs(MinValue, MinValue)
```
cquiroz, line 32: Can you expand on what does this do?

cquiroz (commented):
I gave it a review and added some comments

build.sbt
```diff
@@ -1,23 +1,29 @@
...
+ThisBuild / tlBaseVersion                         := "0.44"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
-ThisBuild / scalacOptions += "-Xsource:3"
+
+ThisBuild / tlCiReleaseBranches += "topic/scala3"
 
 Global / concurrentRestrictions += Tags.limit(Tags.Compile, 1)
 
+ThisBuild / scalaVersion := "3.1.2"
```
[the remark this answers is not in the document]
armanbilge: The PR was stale, it was bumped on the published branch. Sorry :)

build.sbt
```diff
@@ -1,23 +1,29 @@
...
+ThisBuild / crossScalaVersions := Seq("3.1.2")
+ThisBuild / scalacOptions += "-language:implicitConversions" // TODO
```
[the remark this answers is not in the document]
armanbilge: It seems that the use of implicit conversions are becoming very discouraged, thus ideally we should remove this compiler flag in the long-term.

E.g. this is why I added the `.refined` syntax instead of auto-refinement.

build.sbt
```diff
@@ -1,23 +1,29 @@
...
 lazy val monocleVersion        = "3.1.0"
 lazy val scalaJavaTimeVersion  = "2.4.0"
 lazy val geminiLocalesVersion  = "0.7.0"
 lazy val jtsVersion            = "0.3.0"
-lazy val coulombVersion        = "0.5.8"
+lazy val coulombVersion        = "0.6.0-M2"
 lazy val spireVersion          = "0.18.0"
 lazy val singletonOpsVersion   = "0.5.2"
 lazy val refinedVersion        = "0.10.0"
+lazy val lucumaRefinedVersion  = "0.0-e39b79d-SNAPSHOT"
```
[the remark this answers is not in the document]
armanbilge: This was another stale bit, the published branch was on the stable.

build.sbt
```diff
@@ -38,23 +44,20 @@ lazy val core = crossProject(JVMPlatform, JSPlatform)
...
-      "com.manyangled" %%% "coulomb-cats"               % coulombVersion,
-      "com.manyangled" %%% "coulomb-refined"            % coulombVersion,
-      "com.manyangled" %%% "coulomb-physical-constants" % coulombVersion,
+      "com.manyangled" %%% "coulomb-core"               % coulombVersion,
+      "com.manyangled" %%% "coulomb-spire"              % coulombVersion,
+      "com.manyangled" %%% "coulomb-units"              % coulombVersion,
       "org.typelevel"  %%% "spire"                      % spireVersion,
       "org.typelevel"  %%% "spire-extras"               % spireVersion,
-      "eu.timepit"     %%% "singleton-ops"              % singletonOpsVersion,
+      "eu.timepit"     %%% "singleton-ops"              % singletonOpsVersion cross CrossVersion.for3Use2_13,
```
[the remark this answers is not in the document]
armanbilge: I think we can remove singleton-ops. We still need shapeless because of refined.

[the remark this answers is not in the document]
armanbilge: Huh. I removed singleton-ops and the compiler crashed ...

modules/core/shared/src/main/scala/lucuma/core/enums/GmosNorthGrating.scala
armanbilge: `simultaneousCoverage` is expressed in integer nanometers. We cannot change the unit to picometers without loss of precision i.e. truncation.

modules/core/shared/src/main/scala/lucuma/core/enums/ProposalClass.scala
armanbilge: Thanks. Seems it was deleted in https://github.com/gemini-hlsw/lucuma-core/pull/523. Must be a bad merge.

modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
armanbilge: Because `PosInt` and `Quantity[PosInt, Picometer]` both erase to `Int` in bytecode the two `apply` methods have the same signature and conflict. The `DummyImplicit` distinguishes them.

Probably we can use `@targetName` annotation here instead.

armanbilge: Eh, I forgot we already published 0.51 so now we have compatibility constraints :) a `DummyImplicit` is harmless.

modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
armanbilge: A way to refine the value of a `Quantity`.

https://github.com/gemini-hlsw/lucuma-core/blob/3ac15623099b1b854029edbfd3a2cc1780e9f99c/modules/core/shared/src/main/scala/lucuma/core/math/units.scala#L155-L161

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
armanbilge: What do you mean?

modules/core/shared/src/main/scala/lucuma/core/model/ElevationRange.scala
armanbilge: I added this code comment.
```scala
    // manually create witnesses, since refined macro not working in scala 3
    // these are necessary to implement RefinedTypeOps
```

modules/core/shared/src/main/scala/lucuma/core/enums/GmosNorthGrating.scala
armanbilge: Hmm, seems our organize imports rule enforces `_` instead of `*` ...

armanbilge: - https://github.com/liancheng/scalafix-organize-imports/issues/221

cquiroz: Uh haven't seen than

PR #521, opened 2022-05-18
Fix for proper motion calculation considering the value from catalogs is

cquiroz:
multiplied by cos(dec)

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
@@ -49,11 +49,11 @@ final case class SiderealTracking(
   parallax:        Option[Parallax]
 ) {
 
-  def at(i: Instant): Coordinates =
+  def at(i: Instant): Option[Coordinates] =
     plusYears(epoch.untilInstant(i))
 
   /** Coordinates `elapsedYears` fractional epoch-years after `epoch`. */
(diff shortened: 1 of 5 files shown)


toddburnside (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
```diff
@@ -150,7 +150,9 @@ object SiderealTracking extends SiderealTrackingOptics {
 
     // Break out our components
     val (ra, dec)   = baseCoordinates
-    val (dRa, dDec) = properMotion
+    val (dRaʹ, dDec) = properMotion
+    // See: https://app.shortcut.com/lucuma/story/1388/proper-motion-calculation
+    val dRa  =  dRaʹ / cos(dec)
```
cquiroz, line 155: I wasn't sure but dec can be `[90, -90]` inclusive and this becomes undefined. Should I make the return an `Option`? It will be a bit of a problem to handle it upstream but I guess better than an exception

And we all know somebody's going to try

swalker2m (approved):
It bugs me that we cannot have proper motion for an object at 90º, but if proper motion is reported that way relative to dec I don't know what else to suggest.  :+1:

cquiroz:
Maybe @andrewwstephens can comment on Dec 90

PR #523, opened 2022-05-19
Add Proposal and associated classes/tests

toddburnside:
There are some changes from the temporary model in explore. The biggest change is that `ProposalClass` is no longer a simple Enumerated. The instances of `ProposalClass` contain the `minimum percent time` information, and optionally the `minimum percent total time` and `total time`. This makes the API and the UI more complex than previously, but it seems more correct. The old, flatter `Proposal` contained all of those values, even though they weren't applicable to all proposal classes.


modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -183,6 +184,10 @@ trait units {
 
   implicit val convMP: UnitConverter[PosInt, Micrometer, PosInt, Picometer] =
     unsafePosIntConverter[Micrometer, Picometer]
+
+  // Integer Percents
+  type ZeroTo100  = numeric.Interval.Closed[0, 100]
+  type IntPercent = Int Refined ZeroTo100
```
toddburnside, line 190: This doesn't really have units. I started out with it having the coulomb `Percent` unit, but all it seemed to buy me was tears. I want to share it between the API and the UI, but maybe there is a better place than `math.units`?

swalker2m (commented):
Code looks great.  Just a couple of questions.  There is more to a proposal of course, but I figure that will be added as we get to it?

toddburnside:
> There is more to a proposal of course, but I figure that will be added as we get to it?

Right. This is just to get us up to where the UI is.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
```diff
@@ -0,0 +1,43 @@
...
+  val proposalClass = Focus[Proposal](_.proposalClass)
+  val category      = Focus[Proposal](_.category)
+  val toOActivation = Focus[Proposal](_.toOActivation)
+  val abstrakt      = Focus[Proposal](_.abstrakt)
+  val partnerSplits = Focus[Proposal](_.partnerSplits)
+
+  val Default: Proposal =
+    Proposal(None, ProposalClass.Queue(80), None, ToOActivation.None, None, SortedMap.empty)
+
+  implicit val eqProposal: Eq[Proposal] = Eq.instance {
```
cquiroz, line 39: Maybe `Eq.by`?

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
```diff
@@ -0,0 +1,43 @@
...
+final case class Proposal(
+  title:         Option[NonEmptyString],
+  proposalClass: ProposalClass,
+  category:      Option[TacCategory],
+  toOActivation: ToOActivation,
+  abstrakt:      Option[NonEmptyString],
+  partnerSplits: SortedMap[Partner, IntPercent]
+)
+
+object Proposal {
```
cquiroz, line 28: I wonder if we need to include the semester this proposal applies to

modules/core/shared/src/main/scala/lucuma/core/model/ProposalClass.scala
```diff
@@ -0,0 +1,192 @@
...
+import io.circe.generic.semiauto._
+import io.circe.refined._
+import lucuma.core.math.units.IntPercent
+import monocle.macros.GenPrism
+import monocle.{ Focus, Lens, Optional, Prism }
+import org.typelevel.cats.time.instances.duration._
+
+import java.time.Duration
+
+sealed trait ProposalClass { val minPercentTime: IntPercent }
```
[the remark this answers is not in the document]
toddburnside: My understanding is that it is the minimum amount of the requested time for the semester that the PI requires for this proposal. I guess if they can't get that, they won't be able to do science?

modules/testkit/src/main/scala/lucuma/core/model/arb/ArbProposalClass.scala
```diff
@@ -0,0 +1,157 @@
...
+        arbitrary[PoorWeather],
+        arbitrary[Queue],
+        arbitrary[SystemVerification],
+        arbitrary[LargeProgram],
+        arbitrary[Intensive]
+      )
+    )
+
+  implicit val cogProposalClass: Cogen[ProposalClass] =
+    Cogen[Either[
```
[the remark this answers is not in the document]
toddburnside: I'll try. I just took this approach from similar classes.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
toddburnside: I can switch to that. It is shorter. But, the thing I don't like about `Eq.by` is that it is easy to forget to add fields as the class changes (which I'm assuming will happen here). I tried this approach because it is impossible for that to happen, but it is definitely more verbose.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
cquiroz: And in fact we have glossed over the model changes happening from one semester to the next...

modules/core/shared/src/main/scala/lucuma/core/model/ProposalClass.scala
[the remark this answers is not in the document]
swalker2m: Ok, I see.  So we'll track the time request here in the proposal class?  Is this where the actual time asked for goes?  Or is that even an option?  Maybe we just add up the included observations (somehow accounting for OR groups) and do the request with that?

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
cquiroz: Do we ever expect decimals here? maybe `spire.Rational` could be an alternative if so

modules/testkit/src/main/scala/lucuma/core/model/arb/ArbProposalClass.scala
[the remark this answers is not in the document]
toddburnside: Good thinking! That's much cleaner.

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
toddburnside: Andy settled on integer percents. So, for 3 equal partners, the PI will need to decide who to ask 34% from.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
swalker2m: Semester doesn't really apply equally to all proposal classes.  Some are active for less than a full semester and others over multiple semesters.  We had been discussing storing the active period for a program as start and end times.  Maybe it does makes sense to track the semester in which we expect the program to be first made active though.

Model changes should be a different problem than the one we have today.  We don't have to support opening up XMLs from any past time and continuing to migrate it up to the current version.  Instead a model change can happen whenever and we have to include a database migration at that point, and change the API and UI accordingly.  Or am I missing the point?

swalker2m: Semester might be a thing to include in the `Queue` proposal class and others where it makes sense?  Though even there sometimes they'll "rollover" a queue proposal into the next semester to get it done.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
toddburnside: Do you mean `EqTests[Proposal].eqv`? I have that, but I think `Eq` based on a subset of fields is a valid instance wrt the tests. It just isn't what we would usually want. I'm looking forward to someday being able to derive Eq so that this isn't an issue.

modules/core/shared/src/main/scala/lucuma/core/model/ProposalClass.scala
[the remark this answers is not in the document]
toddburnside: Yeah. I think we somehow base the request time on the included observations. The exact calculation probably hasn't been determined... For this initial pass I was just going to use a value from the planned time summary - although I don't know if it should be PI time or Execution time. 

The discussion in Shortcut specified that we will apply these percentages to `band 1 & 2` and `band 3`. But first we need to have bands. For now, `band 3` will always be zero, but Andy wanted to leave it in the UI so people can see it will eventually exist.

swalker2m (approved):
@cquiroz is really the Phase 1 expert, but this all seems perfectly cromulent to me so far. :+1:

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
toddburnside: The semester thing will probably require some additional thought. Even for `LargeProgram` and `Intensive` classes, which are definitely multi-semester, the PI is supposed to be entering the `minPercentTime` that they need for this semester. They also specify the expected `totalTime` for the entire program and minimum percent that they need for that. 

So, it seems to me a proposal, at least, is only for a given semester, regardless of the lifetime of a program. 

I, for one, am not clear on how this works in Explore and the API. In Explore, do we default to the upcoming semester when creating a proposal, but give the user the option of selecting from several in the future? Do we support multiple proposals for a program? I don't recall seeing this in the mockups, but that doesn't mean it isn't there. My recollection is for a logged in user to have the ability to create a proposal if there isn't one. In the API, I have an `Option[Proposal]` on the `ProgramModel`. Maybe this needs to be a `Map[Semeter, Proposal]`?

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
cquiroz: There is an urban legend that derivation is working fairly well in scala 3

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
cquiroz: I guess we'll have q history of m proposal for n semesters for a give user. That maybe an issue with respect to schema evolution too

modules/core/shared/src/main/scala/lucuma/core/model/ProposalClass.scala
```diff
@@ -0,0 +1,192 @@
...
+
+object ProposalClass {
+
+  case class Classical(minPercentTime: IntPercent)          extends ProposalClass
```
rpiaggio, line 23: `final`?

modules/testkit/src/main/scala/lucuma/core/model/arb/ArbProposalClass.scala
```diff
@@ -0,0 +1,141 @@
...
+        Option[FastTurnaround],
+        Option[PoorWeather],
+        Option[Queue],
+        Option[SystemVerification],
+        Option[LargeProgram],
+        Option[Intensive]
+      )
+    ].contramap(p =>
+      (
+        ProposalClass.classical.getOption(p),
```
rpiaggio, line 127: This looks nicer than going with nested `Either`s.

I havent thought this fully through, bu I'm a bit worried that it may present a chicken and egg issue: is this `Cogen` used to test the `Prism`s that are being used to derive it?

rpiaggio: Also, would it be an issue if the prisms were ill formed such that two or more of them returned a value for the same `p`? (Just asking, I don't have an answer).

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
```diff
@@ -0,0 +1,43 @@
...
+
+import scala.collection.immutable.SortedMap
+
+final case class Proposal(
+  title:         Option[NonEmptyString],
+  proposalClass: ProposalClass,
+  category:      Option[TacCategory],
+  toOActivation: ToOActivation,
+  abstrakt:      Option[NonEmptyString],
+  partnerSplits: SortedMap[Partner, IntPercent]
```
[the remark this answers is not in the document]
andrewwstephens: FWIW, my goal was to allow the PI to specify the percentages so that they don't have to worry about adjusting hours per partner when the total time request invariably changes because they added targets or changed their constraints.

modules/core/shared/src/main/scala/lucuma/core/model/ProposalClass.scala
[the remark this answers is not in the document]
andrewwstephens: The "minimum time" is specified in hours in the PIT but here I also suggested using a percentage so that it doesn't change when the total changes.   Todd:  I think this is a percentage of the Execution Time  (PI + partner).

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
andrewwstephens: I would think that you would always be preparing a proposal for the currently open cycle, and the only choice is what type of proposal to submit.  For example, if it's an Fast-Turnaround proposal it would be for the current month, if it's a queue proposal its for next semester, if it's a director's or poor weather proposal it's for the current semester.  It might be difficult to support proposals for times in the more distant future because we won't know which instruments will be offered or their configurations.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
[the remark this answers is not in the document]
swalker2m: This makes sense, particularly considering that the time request itself will just be calculated from the included observations instead of typed in by the user.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
cquiroz: Would you want to keep a record of all proposals ever? So now in 2022 you want to see a proposal sent on 2010, do we allow that?

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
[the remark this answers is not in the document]
swalker2m: As the program changes during phase 2 and execution, we'll lose the initial time request (along with which observations were initially proposed and how long they were initially considered to last).  Is that an issue?  If so, I suppose we might need to store a snapshot of the information at proposal time.

modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
swalker2m: I hope we don't have to import old pre-GPP proposals, but going forward it seems like many programs will have associated proposal information and that will live forever along with the program.

modules/testkit/src/main/scala/lucuma/core/model/arb/ArbProposalClass.scala
toddburnside: Okay... I did a little research and testing. Cogen is used for seed generation, so I don't think nested Either vs tuple of Option makes a difference.

I removed the `Cogen[ProposalClass]` (and by necessity the `Cogen[Proposal]`. The only tests affected by the removals were `EqTests[ProposalClass]`, ˙LensTests(Proposal.proposalClass)' and `EqTests[Proposal]`. The first 2 were due to the lack of `Cogen[ProposalClass]`. The prism tests don't involve the `Cogen[ProposalClass]`, so it seems safe to use them here. 

Any objections?

swalker2m: > Any objections?

No objections here, regardless of how it goes.  It seemed like it might be possible to avoid the `.asRight` chains but in reality it's a minor issue.

PR #525, opened 2022-05-20
advanced config calculations

swalker2m:
The Advanced Config panel in the inception document includes a number of computed values.  This PR adds those computations to core.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthGrating.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthGrating.scala
@@ -5,10 +5,11 @@ package lucuma
 package core
 package `enum`
 
-import cats.syntax.eq._
-import coulomb._
+import cats.syntax.eq.*
+import coulomb.*
 import eu.timepit.refined.types.numeric.PosInt
-import lucuma.core.math.units._
+import lucuma.core.math.{Angle, Coverage, Wavelength}
+import lucuma.core.math.units.*
 import lucuma.core.util.Enumerated
 import spire.math.Rational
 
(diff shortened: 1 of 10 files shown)


modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthGrating.scala
```diff
@@ -18,28 +19,171 @@ import spire.math.Rational
...
+  val dispersion:           Quantity[Rational, NanometersPerPixel],
+  val simultaneousCoverage: Quantity[PosInt, Nanometer],
+  val blazeWavelength:      Wavelength,
+  val referenceResolution:  PosInt,
+  val obsolete:             Boolean
+) extends Product with Serializable {
+
+  /** Compute the coverage of this disperser, given a central wavelength. */
+  def coverage(λ: Wavelength): Coverage =
+    Coverage.centered(λ, simultaneousCoverage)
```
swalker2m, line 35: This is the "λ Coverage" value.

modules/core/shared/src/main/scala/lucuma/core/enum/GmosNorthGrating.scala
```diff
@@ -18,28 +19,171 @@ import spire.math.Rational
...
+  private def Δλ: Double =
+    blazeWavelength.nm.value.toDouble / referenceResolution.value.toDouble
+
+  /** Resolution at λ with the specified slit width. */
+  def resolution(λ: Wavelength, slitWidth: Angle): Int =
+    ((λ.nm.value.toDouble / Δλ) * (0.5 / Angle.signedDecimalArcseconds.get(slitWidth).toDouble)).toInt
+
+  /** Resolution at λ with the effective slit width of the given FPU. */
+  def resolution(λ: Wavelength, fpu: GmosNorthFpu): Int =
+    resolution(λ, fpu.effectiveSlitWidth)
```
swalker2m, line 50: This would be the "λ / Δλ" value.

modules/core/shared/src/main/scala/lucuma/core/math/Coverage.scala
```diff
@@ -0,0 +1,72 @@
...
+package lucuma.core.math
+
+import cats.Eq
+import cats.syntax.order._
+import coulomb._
+import eu.timepit.refined.types.numeric.{NonNegInt, PosInt}
+import lucuma.core.math.units._
+
+/** Wavelength coverage. */
+sealed trait Coverage {
```
swalker2m, line 13: Ported over from @tpolecat 's original in the old `gem` project.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/GmosCcdMode.scala
```diff
@@ -14,7 +17,20 @@ final case class GmosCcdMode(
   ampCount:    GmosAmpCount,
   ampGain:     GmosAmpGain,
   ampReadMode: GmosAmpReadMode
-)
+) {
+
+  def readNoise: Quantity[PosBigDecimal, Electron] = {
```
swalker2m, line 22: And this, of course, is the "Read Noise".

modules/core/shared/src/main/scala/lucuma/core/math/Coverage.scala
```diff
@@ -0,0 +1,72 @@
...
+      case (Empty, _) => Empty
+      case (_, Empty) => Empty
+      case (Range(a, b), Range(aʹ, bʹ)) => Coverage(a max aʹ, b min bʹ)
+    }
+
+  /** Coverage width; i.e., difference between max and min (or zero). */
+  def width: Quantity[NonNegInt, Picometer] =
+    this match {
+      case Empty       => NonNegInt.unsafeFrom(0).withUnit[Picometer]
+      case Range(a, b) => NonNegInt.unsafeFrom(b.toPicometers.value.value - a.toPicometers.value.value).withUnit[Picometer]
```
cquiroz, line 28: Given there is a substraction, this could go negative or is that protected by the `Range` constructor?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -21,6 +21,9 @@ import spire.math._
 
 trait units {
 
+  type Electron
+  implicit val defineUnitElectron = BaseUnit[Electron](abbv = "e-")
```
swalker2m, line 25: I'm not sure about this.  Need to check with Andy I think.

modules/core/shared/src/main/scala/lucuma/core/enum/GmosSouthGrating.scala
```diff
@@ -18,26 +19,146 @@ import spire.math.Rational
...
+    tag                  = "B480_G5327",
+    shortName            = "B480",
+    longName             = "B480_G5327",
+    rulingDensity        = 480,
+    dispersion           = pmToDispersion( 62),
+
+    // FAKE NEWS!  TODO: what are the real values?
+    simultaneousCoverage = nm( 1),
+    blazeWavelength      = blazeNm( 1),
+    referenceResolution  = resolution(1),
```
swalker2m, line 131: Need to ask about the `B480` values.  I don't know where to find them.

modules/core/shared/src/main/scala/lucuma/core/math/Coverage.scala
swalker2m: Right, `Range` can only be created if it max is strictly > min.

PR #528, opened 2022-05-25
Move NonNegDuration from tmp-api and update ProposalClasses

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/model/package.scala
```diff
@@ -0,0 +1,61 @@
...
+package lucuma.core
+
+import cats.Eq
+import cats.Monoid
+import cats.syntax.all._
+import eu.timepit.refined.api.{ Refined, RefinedTypeOps, Validate }
+import eu.timepit.refined.api.Validate.Plain
+import eu.timepit.refined.boolean.Not
+import eu.timepit.refined.numeric.{ GreaterEqual, Interval, Less, NonNegative }
+import shapeless.Nat._0
```
cquiroz, line 13: Can this be done without shapeless?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/package.scala
toddburnside: I really don't know. I just copied it from the api. 😄 I can look into it a little. Maybe @swalker2m knows more?

toddburnside: I'll try replacing it.

toddburnside: I'm going to merge, and then see if I can fix this. I'd like to get my api PR in before there are more merge conflicts.

swalker2m: > I'm going to merge, and then see if I can fix this. I'd like to get my api PR in before there are more merge conflicts.

:+1:

toddburnside: It turns out this probably isn't easily possible. It is relying on the fact that `Not` and `Less` are case classes with companion objects with the implicit validation functions. `NonNegative` is a type alias and so cannot have a companion object. I think in order to try to do the validation ourselves, we'd end up using the `shapeless.Nat._0`. Or we'd have to implement the `Nat`s.

cquiroz: Never mind, I was mostly asking in the context of the scala 3 port

PR #529, opened 2022-05-27
Add PosAngle constructors with wider type

swalker2m:
While affixing `: PosAngle` to everything to widen the type, it occurred to me these might be handy.  No hurry in getting them in, but wanted to leave it there for a future release.


swalker2m:
Also, maybe could we consider renaming it `PosAngleConstraint`?  Taken as a whole, it's actually a constraint on what the position angle may be and not a position angle per se.

cquiroz:
I'm fine with the renaming

cquiroz (approved):
LGTM

PR #530, opened 2022-05-30
Pos angle constraint

swalker2m:
This is mostly a renaming of `PosAngle` to `PosAngleConstraint`, to better reflect that it is a constraint on the eventual position angle and not a position angle per se.  Includes:

* Rename `PosAngle` to `PosAngleConstraint along with arbitraries and test suite
* Adds a `PosAngleConstraint.angle` `Optional` to extract specific position angles where possible
* Renames and groups the optics

I'm not sure about the importance of the last point but I noticed that we were including `Prism` in the names and that we don't do that elsewhere (and that not all the optics were in fact prisms anyway).  If anybody disagrees or has better suggestions let me know.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-ThisBuild / tlBaseVersion                         := "0.39"
+ThisBuild / tlBaseVersion                         := "0.40"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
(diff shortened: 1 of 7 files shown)


cquiroz:
Right, I'd say we could think how do we want the API to present the params, and follow accordingly?

cquiroz (approved):
LGTM

PR #538, opened 2022-06-08
Remove PosAngle unconstrained

cquiroz:
This is an optional constraint thus a `Some(Unconstrained)` sounds illogical


swalker2m (approved):
Looks good to me. :+1:  No constraint on the position angle implies no `PosAngleConstraint`.

PR #550, opened 2022-06-29
Add inner radius geometry

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
```diff
@@ -39,5 +40,36 @@ trait GmosCandidatesArea {
...
+  def innerRadius(probe: GuideProbe): ShapeExpression =
+    probe match {
+      case GuideProbe.GmosOIWFS =>
+        // 20 arcsec
+        ShapeExpression.centeredEllipse((20 * 2).arcsec, (20 * 2).arcsec)
+      case GuideProbe.PWFS1 =>
+        ShapeExpression.centeredEllipse((345 * 2).arcsec, (345 * 2).arcsec)
+      case GuideProbe.PWFS2 =>
+        ShapeExpression.centeredEllipse((315 * 2).arcsec, (315 * 2).arcsec)
+    }
```
swalker2m, line 56: eventually there would be F2, canopus, etc. guide probes here which i guess would need to all be handled via `case _ =>`

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
```diff
@@ -39,5 +40,36 @@ trait GmosCandidatesArea {
       a <- posAngles
       o <- offsetPositions
     } yield candidatesAreaAt(a, o)).fold(ShapeExpression.Empty)(_ ∪ _)
+
+  /**
+    * GMOS inner radius
+    */
+  def innerRadius(probe: GuideProbe): ShapeExpression =
```
swalker2m, line 47: This is about avoiding selection of the science target itself or avoiding vignetting?  It seems to be the former in the case of GMOS OI and the later for PWFS?

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
```diff
@@ -39,5 +40,36 @@ trait GmosCandidatesArea {
...
+    probe match {
+      case GuideProbe.GmosOIWFS =>
+        // 20 arcsec
+        ShapeExpression.centeredEllipse((20 * 2).arcsec, (20 * 2).arcsec)
+      case GuideProbe.PWFS1 =>
```
swalker2m, line 52: Thinking about this, the "candidates area" will be completely different for GMOS OI and the PWFSes in general.  I wonder if we need the `GuideProbe` enum.  Wouldn't there be a completely different `Pwfs1CandidatesArea` just as there is a `GmosCandidatesArea`?  That is, the `GmosCandidatesArea` would just be for GMOS OI.  Then presumably we could extract the common interface into a `CandidatesArea`.

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
swalker2m: One reason I ask is because I'm hoping to not have the vignetting avoidance inner radius at all for PWFS and instead define the geometries and vignetting as we do for GMOS OI.

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
cquiroz: Right, we'd be adding other instrument's probes and for OI too

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
cquiroz: Will that PWFS area be the same for any instrument?
I suspect we still need the enum to define the selected probe though it maybe not needed at this level

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
swalker2m: If it is just about avoiding the selection of the science target, why would it differ based on guide probe?

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
swalker2m: The PWFS guide probe should reach the same stars regardless of which instrument is in use.  In the old codebase we defined differing inner radius limits to avoid vignetting the science area of the various instruments.  In the new code, hopefully we can do it via defining the PWFS probe geometry and calculating vignetting like we do for GMOS OI.

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
cquiroz: I suppose not only the science target but those very near. This was specified as 20 arcsecs for GMOS OIWFS

modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosCandidatesArea.scala
cquiroz: In the case of not removing the center area, we'd calculate a vignetting value even for the science target. We'd expect that to have the highest vignetting but would still be an allowed selection.

swalker2m: Perhaps what we want (?) is to ask "does positioning the guide probe to this guide star cause the science target coordinate to be vignetted by the probe arm?" rather than "how close is this guide star to the science target" (i.e., is the science target coordinate in the probe arm shadow geometry). It's going to depend on how the probe arm moves and how much of a shadow it casts.

swalker2m: For that, we might need to calculate the effective size of the science target itself.

cquiroz: Right, which is probably hard to do
We could also be covering the science target if we pick a star very  close right?

swalker2m: Something like this:

https://github.com/gemini-hlsw/lucuma-tmp-api/blob/master/modules/core/src/main/scala/lucuma/odb/api/model/gmos/longslit/GmosLongslitMath.scala#L35-L39

but we need to review the `Uniform` profile case.  If we have the estimate for the size of the target and we have the probe arm geometry then we can take the intersection.

The exclusionary-zone-around-the-science-target idea seems a bit rough since it depends on how the probe arm moves and how big the shadow is.

cquiroz: I suppose it also depend on PosAngle?

cquiroz:
After some discussions we'll do this in a different way directly in AGS

PR #555, opened 2022-07-07
Validation framework.

rpiaggio:
This PR attempts to generalize logic we had elsewhere to define validators and formatters.

It defines 2 new optics:
- `ValidSplitEpi`: essentialy a `Format` but the getter returns an `Either[E, A]` instead of an `Option[A]`, allowing us to centrally define parsing errors.
- `ValidWedge`: similar to `ValidSplitEpi` but allows normalization to happen when formatting (`reverseGet`). Especially useful when we want to define truncated decimals.

Furthermore, several instances of these are predefined for common numeric and model types.


modules/core/shared/src/main/scala/lucuma/core/validation/InputValidWedge.scala
```diff
@@ -0,0 +1,107 @@
...
+      (iso.get _).andThen(_.asRight),
+      iso.reverseGet
+    )
+
+  /**
+   * Build a `InputValidWedge` for `BigDecimal` truncated to `Dec` decimals.
+   */
+  def truncatedBigDecimal[Dec <: XInt](implicit
+    @unused req: Require[&&[Dec >= 0, Dec < 10]],
+    vo:          ValueOf[Dec]
```
toddburnside, line 74: If we no longer have Truncated* types for BigDecimal, RA, Dec, etc., do we still need the decimals as a type parameter, or could we just pass in a NonNegInt as a regular paramter. My recollection is we needed it this way because the type of TruncatedBigDecimal depended on the number of decimals - which is not a thing anymore. The best way to deal with singleton ops for scala 3 would be to eliminate them...

The upper bound of 10 was just an arbitrary sanity thing. We could either leave that to programmer discretion, or have the parameter be a refined Interval.

rpiaggio: I think you are right, we could just make the decimals a regular parameter.

cquiroz:
There is quite a bit of forrmatting changes, Could they be in a separate commit?

modules/core/shared/src/main/scala/lucuma/core/optics/ValidWedge.scala
```diff
@@ -0,0 +1,167 @@
...
+import cats.syntax.all._
+import eu.timepit.refined.api.Refined
+import eu.timepit.refined.api.{Validate => RefinedValidate}
+import lucuma.core.optics._
+import monocle.Iso
+import monocle.Prism
+
+/**
+ * A `ValidWedge` is to a `Wedge` what `ValidSplitEpi` is to a `SplitEpi`.
+ */
```
cquiroz, line 17: Two comment blocks?

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidWedge.scala
```diff
@@ -0,0 +1,107 @@
...
+import cats.data.NonEmptyChain
+import cats.syntax.all._
+import eu.timepit.refined.auto._
+import eu.timepit.refined.numeric.Positive
+import eu.timepit.refined.types.numeric.PosBigDecimal
+import eu.timepit.refined.types.string.NonEmptyString
+import lucuma.core.optics._
+import monocle.Iso
+import monocle.Prism
+import singleton.ops._
```
cquiroz, line 15: Singleton ops is no longer available on scala 3. Could we consider an alternative?

rpiaggio:
> There is quite a bit of forrmatting changes, Could they be in a separate commit?

scalafmt is formatting on save. I think the only existing file that changed is `Angle`, where I just added the `BigDecimal` methods.

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidWedge.scala
rpiaggio: Yup, I'm going to try to implement what Todd suggested. Maybe let's hold the review until then.

rpiaggio:
Converted to draft to hold on the review for the moment.

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidWedge.scala
toddburnside: See the comment above. I don't think it is necessary anymore.

modules/core/shared/src/main/scala/lucuma/core/optics/ValidSplitEpi.scala
```diff
@@ -0,0 +1,176 @@
...
+          self.getValid(a).map(_.some),
+      (b: Option[B]) => b.foldMap(self.reverseGet)
+    )
+
+  /**
+   * Build `ValidSplitEpi` from another one, refining the return type with predicate `P`.
+   */
+  def refined[P](error: E)(implicit ev: RefinedValidate[B, P]): ValidSplitEpi[E, A, Refined[B, P]] =
+    this.andThen(ValidSplitEpi.forRefined[E, B, P](error))
+}
```
swalker2m, line 124: Maybe

```
def asFormat: Format[A, B] =
  Format(a => getValid(a).toOption, reverseGet)
```

modules/core/shared/src/main/scala/lucuma/core/optics/README.md
```diff
@@ -10,6 +10,9 @@ When we compose a split mono and a split epi end-to-end in either direction we e
 
 A `Format` is a weaker `Prism` where a *subset* of `A` forms a split epi with `B`. Every `Prism` is a `Format` where the split epi happens to be an `Iso`; and every `SplitEpi` forms a `Prism` where the subset of `A` is `A` itself.
 
+A `ValidSplitEpi` is a reformulation of a `Format`, where elements of `A` that don't map to an element of `B` return an element of `E` instead. A `Format[A, B]` is a `ValidSplitEpi[Unit, A, B]`.
+
+A `ValidWedge` is to a `Wedge` what a `ValidSplitEpi` is to a `SplitEpi`. In other words, it allows for reverse normalization.
```
swalker2m, line 15: I like this `ValidSplitEpi` concept but I wonder about the "split epi"ness of it.  As I understand them, split epi and wedge are special `Iso`s where you can always go back and forth between the `A` and the `B` even if a cycle doesn't yield the same value you started with.  As you rightly point out `ValidSplitEpi` is a reformulation of `Format`, a `ValidFormat` if you will.  I guess what I'm asking is why not drop the `ValidWedge` and call the `ValidSplitEpi` a `ValidFormat`?

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidSplitEpi.scala
```diff
@@ -0,0 +1,178 @@
...
+import lucuma.core.optics._
+import monocle.Iso
+import monocle.Prism
+
+import java.math.RoundingMode
+import java.text.DecimalFormat
+import scala.util.Try
+
+/**
+ * Convenience version of `ValidFormat` when the error type is `NonEmptyChain[NonEmptyString]` and
```
swalker2m, line 25: Oh, it looks like it was called `ValidFormat` in an earlier version?

modules/core/shared/src/main/scala/lucuma/core/optics/ValidSplitEpi.scala
swalker2m: Sorry, this doesn't really have anything to do with the `refined` method.  Just seemed like a handy converter we might add.

modules/core/shared/src/main/scala/lucuma/core/optics/README.md
rpiaggio: That's exactly what we had before (`ValidSplitEpi` named as `ValidFormat`). This had similar laws than `Format`.

However this fell short for us in some cases, specifically when we truncating decimal numbers, since in that case we are normalizing both ways, and that's how `ValidWedge` was born (and `ValidFormat` renamed to `ValidSplitEpi`).

I fiddled a bit with the names and settled on these which seem to lessen the mental burden of introducing new concepts by expanding on existing ones. The `Valid*` versions of `SplitEpi` and `Wedge` are basically the same as those for the subset of valid instances of `A`.

Also, I'm reintroducing `ValidFormat` in the next PR as a generalization of `ValidSplitEpi` and `ValidWedge` since they share the same `getValid`/`reverseGet` methods and it makes it more convenient for APIs to receive either.

All that said, I'm completely open to renaming.

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidSplitEpi.scala
rpiaggio: Yup, see previous comment. Also, good catch, I'm fixing the comment.

modules/core/shared/src/main/scala/lucuma/core/optics/README.md
swalker2m: :+1: I see, that makes sense and I have no better ideas.

modules/core/shared/src/main/scala/lucuma/core/optics/ValidWedge.scala
rpiaggio: Thank you, fixed.

modules/core/shared/src/main/scala/lucuma/core/optics/README.md
toddburnside: There is quite a long history of discussion of this issue. In the distant past, we ran into issues with `ValidFormat` when we realized we needed to be able to truncate decimals in the UI. We discussed a couple of alternatives: 

1. Weaken the laws to essentially be those of the new `ValidWedge`.
2. Create specific types for things like `TruncatedBigDecimal`, `TruncatedRefinedBigDecimal`, `TruncatedRA`, etc. that guaranteed only `n` decimal places so the `ValidFormat` laws would pass.

We opted for number 2, but reassessed it recently. We decided to have 2 versions with different guarantees instead.

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidWedge.scala
```diff
@@ -0,0 +1,98 @@
...
+      InputValidSplitEpi.bigDecimal.getValid
+        .andThen(_.map(_.setScale(decimals.value, scala.math.BigDecimal.RoundingMode.HALF_UP))),
+      bd =>
+        s"%.${decimals.value}f"
+          .format(bd)
+          .replaceAll("^-0\\.(0+)$", "0.$1") // Remove negative 0
+    )
+
+  // We can't build a generic truncatedRefinedBigDecimal. Formatting may cause the refinement to break.
+  // For example: a truncated PosBigDecimal may turn a small value into a 0, which is no longer positive.
```
toddburnside, line 80: Do we still have a way to deal with things like `HourAngle`, which has a `BigDecimal` refined to the interval `-5.0` to `5.0`. In the old version we used a `SplitEpi` called `unsafeRefinedBigDecimal` to edit it to 1 decimal place. It was unsafe because, as you say here, it could not be used for the general case, but it would work if you constrained the decimals proper.

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidSplitEpi.scala
```diff
@@ -0,0 +1,181 @@
...
+  ): InputValidSplitEpi[BigDecimal Refined P] =
+    bigDecimalWithScientificNotation.refined[P](NonEmptyChain("Invalid format"))
+
+  /**
+   * `InputValidSplitEpi` for `PosBigDecimal`, formatting with only one integer digit.
+   */
+  val posBigDecimalWithScientificNotation: InputValidSplitEpi[PosBigDecimal] =
+    refinedBigDecimalWithScientificNotation[Positive]
+
+  private def fixIntString(str: String): String = str match {
```
cquiroz, line 159: This is not important but I wonder if these should be in lucuma-ui as I guess they are specific to how we display and edit values?

rpiaggio: Are you referring to the `fix*` methods? Those are called by the `int` and `bigDecimal` validators.

I moved the predefined numeric and model validators since I'm guessing can be used by the API.

modules/core/shared/src/main/scala/lucuma/core/validation/InputValidWedge.scala
rpiaggio: I made specific instances for those cases in explore, which are also tested now. Maybe it makes sense to move them here too so they can be used by the API?

toddburnside: Cool. I guess if the API is going to use these validators, they'll need to validate the elevation range stuff. I don't know what the API is doing, though.

swalker2m: https://github.com/gemini-hlsw/lucuma-tmp-api/blob/master/modules/core/src/main/scala/lucuma/odb/api/model/ElevationRangeModel.scala#L88

I'm probably not going to update the `tmp-api` much from here but it may be of use in the real ODB.

tpolecat (approved):
The discussion makes sense and this seems pretty well thought out, so :+1:

rpiaggio:
Thank you for all the feedback. I'll merge this and move over some more validators from explore in the next PR.

PR #556, opened 2022-07-08
Elevation range validators, Display.byTag, scientific notation normalization

rpiaggio:
- Validators for elevation range.
- `Display.byTag` for `Enumerated`s.
- Normalize scientific notation, by using lowercase `e` and removing `e0`.


cquiroz (approved):
LGTM

PR #562, opened 2022-07-27
Add validator for Epoch with no scheme

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/math/validation/MathValidators.scala
```diff
@@ -17,6 +17,9 @@ trait MathValidators {
   val epoch: InputValidSplitEpi[Epoch] =
     InputValidSplitEpi.fromPrism(Epoch.fromString, "Invalid epoch")
 
+    val epochNoScheme:InputValidWedge[Epoch] =
```
cquiroz, line 20: Is it miss indented?

toddburnside: Oh, probably. I need to re-enable scalaFmt for me locally.

cquiroz (approved):
LGTM

rpiaggio (commented):
LGTM

PR #563, opened 2022-07-29
Make AnyVal extensions vals private.

rpiaggio:


modules/core/jvm/src/main/scala/lucuma/core/geom/jts/jvm/syntax/Awt.scala
```diff
@@ -10,14 +10,14 @@ import org.locationtech.jts.geom.Coordinate
 import java.awt.geom.Point2D
 
 /**
-  * Syntax for converting JTS shapes to AWT shapes.
-  */
-final class AwtOps(val self: JtsShape) extends AnyVal {
+ * Syntax for converting JTS shapes to AWT shapes.
+ */
+final class AwtOps(private val self: JtsShape) extends AnyVal {
```
cquiroz, line 15: Why do you need them to be private?

rpiaggio: The name of the val is otherwise added as a property.

For example, in this case if you have `val foo: JtsShape`, then `foo.self` is valid, resulting in an identity.

This is particularly undesirable when it interferes with other extension methods. In particular, `DisplayOps` variable is called `value` and it interferes with `coulomb`'s `value`.

cquiroz: Something like
```extension(a: JtsShape):
     def toAwt...
```

cquiroz (approved):
Thanks. perhaps we don't yet switch to scala 3 syntax as we may need to keep this compiling to scala 2 as well

PR #565, opened 2022-08-01
Add addition and subtraction to Wavelength

toddburnside:
There is currently unsafe math being performed in `lucuma-itc` by performing the addition/subtraction on the refined `.toPicometers` values and testing for "out of range". However, by the `toPicometers` can never be out of range by the logic of the refined algebra. 

The addition provided here constrains the value to a max of Wavelength.Max, and subtraction to a min of Wavelength.Min. (So that `Wavelength(2) - Wavelength(2) === Wavelength(1)`, not `Wavelength(Int.MaxValue)`

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
@@ -59,6 +59,16 @@ final case class Wavelength(toPicometers: Quantity[PosInt, Picometer]) {
 
   def angstrom: Quantity[Rational, Angstrom] = Å
 
+  /**
+    * Addition, with result constrained to Wavelength.Max
+    */
+  def +(that: Wavelength): Wavelength = Wavelength.plus(this, that)
+  
+  /**
+    * Subtraction, with result constrained to Wavelength.Min
+    */
+  def -(that: Wavelength): Wavelength = Wavelength.minus(this, that)
+
   override def toString: String =
     s"Wavelength(${toPicometers.show})"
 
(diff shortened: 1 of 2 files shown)


cquiroz (approved):
Looks great thanks

swalker2m (commented):
It seems vaguely off to add or subtract wavelengths directly, though I'm not exactly a scientist of course. I mean, you wouldn't add the wavelength of say blue light to that of yellow light and expect to get something meaningful?  Or would you?

I'm thinking of the wavelength dithers for example in GMOS longslit where we increase the desired wavelength by, say 5 nm, for some steps.  We get that by adding a length to a wavelength rather than adding a 5 nm wavelength.  It's the same thing in the end though I suppose.

I'm more unsure about the other idea of silently capping the max and min.  If we end up doing that it feels like we're masking an error?

toddburnside:
> It seems vaguely off to add or subtract wavelengths directly, though I'm not exactly a scientist of course. I mean, you wouldn't add the wavelength of say blue light to that of yellow light and expect to get something meaningful? Or would you?

You have a point. We could just isolate this math to `lucuma-itc` if you think that would be best. I'm not sure why `itc` is adding/subtracting 2 wavelengths and what it is doing with the results. ie. What the results "mean".

> I'm more unsure about the other idea of silently capping the max and min. If we end up doing that it feels like we're masking an error?

We actually have the potential of masking errors all the time. For example `Int.MaxValue + 1 === Int.MinValue`. Which probably isn't what is expected. The only real correct way would probably be to have most math return Options. Which would be quite painful, obviously. So, instead, we cover our eyes and just pretend the results will always be in range. So, this is just an attempt to do something `slightly` better. But is still incorrect. I'm not sure what the likelihood of itc going out of range is.

Given all of these uncertainties, I think I'll just fix the math in 'lucuma-itc` instead of adding it here.

cquiroz:
Those are good questions. The code was directly ported from the old basic-case repo
https://github.com/gemini-hlsw/basic-case/blob/master/src/main/scala/syntax/Wavelength.scala#L8

swalker2m:
> The code was directly ported from the old basic-case repo

This is true.  I guess it just seemed like there should be a higher bar for doing this kind of thing in `core` vs. some syntax for a particular app.

`refined` itself doesn't do anything like clipping at min or max, as far as I know.

toddburnside:
> `refined` itself doesn't do anything like clipping at min or max, as far as I know.

No, it effectively does modulus arithmetic within the range, just like std scala types. 

Closing this and sticking with the syntax in itc.

PR #566, opened 2022-08-05
Model Visits

rpiaggio:
Events and dataset information not modeled yet.

I'm not sure we gain much by having a `Step` trait that groups `FutureStep` and `StepRecord`. In fact, I purposely didn't define a `Eq[Step]` instance.

Opinions?


modules/core/shared/src/main/scala/lucuma/core/model/sequence/Step.scala
```diff
@@ -86,17 +95,117 @@ object Step extends WithUid('s') {
...
+
+object StepRecord {
+  final case class GmosNorth(
+    id:               Step.Id,
+    created:          Instant,
+    startTime:        Option[Instant],
+    endTime:          Option[Instant],
+    instrumentConfig: DynamicConfig.GmosNorth,
+    stepConfig:       StepConfig
+  ) extends StepRecord
```
swalker2m, line 130: I'm not sure it matters for the purpose of modeling this, but as I mentioned in a separate email the times come from (and must be consistent with) step events from Observe.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Visit.scala
```diff
@@ -0,0 +1,80 @@
...
+import java.time.Instant
+
+sealed trait Visit {
+  def id: Visit.Id
+  def created: Instant
+  def staticConfig: StaticConfig
+  def steps: List[StepRecord]
+
+  def startTime: Option[Instant] = steps.headOption.flatMap(_.startTime)
+  def endTime: Option[Instant]   = steps.lastOption.flatMap(_.endTime)
```
swalker2m, line 26: These times will come from sequence events, when present (which will be the usual case).

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Step.scala
rpiaggio: I see. I'll add the step events then and remove the start and end times.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Visit.scala
rpiaggio: Same here.

rpiaggio:
Closing this, the new classes will go to `lucuma-schemas`.

PR #569, opened 2022-08-12
remove Enumerated.of and Display.fromTag

tpolecat:
This disconnects the `tag` property of `Enumerated` instances from their `productPrefix`, and removes the ability to easily derive `Display` from tags (which should never be shown to the user). See discussion starting at https://gemini-software.slack.com/archives/GJQDX830D/p1660328710537259

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/CatalogName.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/CatalogName.scala
@@ -5,15 +5,14 @@ package lucuma.core.enums
 
 import lucuma.core.util.Enumerated
 
-sealed trait CatalogName extends Product with Serializable
+sealed abstract class CatalogName(val tag: String) extends Product with Serializable
 
 object CatalogName {
-  case object Simbad extends CatalogName
-  case object Horizon extends CatalogName
-  case object Gaia extends CatalogName
+  case object Simbad extends CatalogName("simbad")
+  case object Gaia extends CatalogName("gaia")
 
   /** @group Typeclass Instances */
   implicit val CatalogNameEnumerated: Enumerated[CatalogName] =
-    Enumerated.of(Simbad, Horizon, Gaia)
+    Enumerated.from(Simbad, Gaia).withTag(_.tag)
 
 }
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/CloudExtinction.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/CloudExtinction.scala
@@ -6,29 +6,30 @@ package lucuma.core.enums
 import lucuma.core.util.Display
 import lucuma.core.util.Enumerated
 
-sealed abstract class CloudExtinction(val toDeciBrightness: Int) extends Product with Serializable {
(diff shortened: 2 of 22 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/CatalogName.scala
```diff
@@ -5,15 +5,14 @@ package lucuma.core.enums
...
 
-sealed trait CatalogName extends Product with Serializable
+sealed abstract class CatalogName(val tag: String) extends Product with Serializable
 
 object CatalogName {
-  case object Simbad extends CatalogName
-  case object Horizon extends CatalogName
-  case object Gaia extends CatalogName
+  case object Simbad extends CatalogName("simbad")
+  case object Gaia extends CatalogName("gaia")
```
swalker2m, line 12: `Horizon` should have at least been `Horizons` but did we use it for anything?

cquiroz (approved):
LGTM, Thanks for taking care of this

modules/core/shared/src/main/scala/lucuma/core/enums/CloudExtinction.scala
```diff
@@ -6,29 +6,30 @@ package lucuma.core.enums
...
-  case object PointOne       extends CloudExtinction(1)
-  case object PointThree     extends CloudExtinction(3)
-  case object PointFive      extends CloudExtinction(5)
-  case object OnePointZero   extends CloudExtinction(10)
-  case object OnePointFive   extends CloudExtinction(15)
-  case object TwoPointZero   extends CloudExtinction(20)
-  case object ThreePointZero extends CloudExtinction(30)
+  case object PointOne       extends CloudExtinction("point_one", 1)
+  case object PointThree     extends CloudExtinction("point_three", 3)
+  case object PointFive      extends CloudExtinction("point_five", 5)
```
swalker2m, line 17: i'm tempted to make these `ZeroPointOne` etc to be consistent with the `OnePoint...` `TwoPoint...`, and `ThreePoint...`s. But we'd need to change the schema and patch up tmp-api accordingly.

swalker2m (approved):
:+1:

modules/core/shared/src/main/scala/lucuma/core/enums/CloudExtinction.scala
tpolecat: Maybe come back to this once we're off tmp-api?

modules/core/shared/src/main/scala/lucuma/core/enums/CatalogName.scala
tpolecat: These are used for `CatalogInfo` which is for sidereal targets, which don't exist in Horizons, so I think it's unused.

modules/core/shared/src/main/scala/lucuma/core/enums/CloudExtinction.scala
swalker2m: Sounds good.

tpolecat:
@armanbilge can you cherry-pick this into your scala3 branch?

PR #573, opened 2022-08-27
Derive Enumerated from enum.

rpiaggio:
Derives an `Enumerated` instance for `enum`s with `val tag: String`.

Example:
```scala
enum Theme(val tag: String) derives Enumerated:
   case Light extends Theme("light")
   case Dark extends Theme("dark")
```

Compilation fails if:
- Applied to something other than an `enum` (actual requirement is that `object E` must have a `values` method that returns an `Array[E]`).
- There is no ~~publicly accessible~~ `tag`, or it doesn't return a `String`.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
@@ -7,10 +7,11 @@ package util
 
 import cats.Order
 import cats.data.NonEmptyList
-import cats.syntax.all._
-import io.circe._
-import lucuma.core.syntax.string._
+import cats.syntax.all.*
+import io.circe.*
+import lucuma.core.syntax.string.*
 import monocle.Prism
+
 import scala.quoted.*
 
 /**
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
@@ -11,6 +11,7 @@ import cats.syntax.all._
 import io.circe._
 import lucuma.core.syntax.string._
 import monocle.Prism
+import scala.quoted.*
 
 /**
  * Typeclass for an enumerated type with unique string tags and a canonical ordering.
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
```diff
@@ -70,6 +72,23 @@ object Enumerated {
...
+    Select.unique(x.asTerm, "tag").asExprOf[String]
+
+  private def enumeratedImpl[E: Type](using Quotes): Expr[Enumerated[E]] =
+    '{
+      Enumerated
+        .fromNEL(NonEmptyList.fromList(${ enumValuesImpl[E] }.toList).get)
+        .withTag(x => ${ tagImpl[E]('x) })
+    }
+
+  inline def derived[E]: Enumerated[E] = ${ enumeratedImpl[E] }
```
cquiroz, line 91: Will this make the macro expand everywhere it is called? Or only once in the companion object?

cquiroz (approved):
This is a great idea, who said macros were bad 😝. Should we test what happens if you have no tag field?

Could you use a private field?

modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
rpiaggio: Technically, it is expanded wherever `Enumerated.derived` is called, which we shouldn't do more than once, or none at all: When `derives Enumerated` is used on an `enum`, this is syntactic sugar for using `Enumerated.derived` on its companion object.

rpiaggio:
> Should we test what happens if you have no tag field?

The generated code will try to reference the `tag` field, the same way as if it was typed. If there's no `tag` field, compilation fails with `value tag is not a member of XXX`.

> Could you use a private field?

Actually, yes! Since it's accessed from the companion object, it's still accessible.

toddburnside:
Nice. 🥳  I was thinking of just this thing the other day, but didn't get very far into looking into it.

PR #577, opened 2022-09-08
Timestamp

swalker2m:
When I needed timestamps in `lucuma-odb` I recalled that I had produced a `Timestamp` wrapper class in the old `gem` project.  I copied it over and converted it to be an `opaque type` instead of a simple wrapper.  (I don't know if value classes were a thing at the time.) At any rate, @cquiroz subsequently pointed out that `Timestamp` was already in `lucuma-core`. Since an `opaque type` seems like a better way to handle this than a wrapper I'm bringing the `lucuma-odb` `Timestamp` over to `lucuma-core`.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/Timestamp.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/Timestamp.scala
@@ -3,125 +3,162 @@
 
 package lucuma.core.util
 
-import cats._
-import cats.syntax.all._
+import cats.Order
+import cats.syntax.either.*
+import cats.syntax.order.*
+import io.circe.Decoder
+import io.circe.Encoder
 import lucuma.core.optics.Format
+import lucuma.core.optics.SplitEpi
+import monocle.Prism
+import org.typelevel.cats.time.instances.instant.*
 
 import java.time.Instant
+import java.time.LocalDateTime
 import java.time.ZoneOffset.UTC
 import java.time.ZonedDateTime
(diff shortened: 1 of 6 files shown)


modules/core/shared/src/main/scala/lucuma/core/util/Timestamp.scala
```diff
@@ -3,125 +3,162 @@
...
+    formatter(iso = true)
+
+  def parse(s: String): Either[String, Timestamp] =
+    Either
+      .catchOnly[DateTimeParseException](LocalDateTime.parse(s, Formatter).toInstant(UTC))
+      .orElse(Either.catchOnly[DateTimeParseException](LocalDateTime.parse(s, IsoFormatter).toInstant(UTC)))
+      .leftMap(_ => s"Could not parse as a Timestamp: $s")
+      .flatMap(fromInstant(_).toRight(s"Invalid Timestamp: $s"))
+
+  extension (timestamp: Timestamp) {
```
cquiroz, line 103: I wonder if we should offer some exetsion for `Instant`

cquiroz (approved):
LGTM

tpolecat (approved):
Looks great 👍🏻

modules/core/shared/src/main/scala/lucuma/core/util/Timestamp.scala
swalker2m: Seems like a good idea.

PR #578, opened 2022-09-08
Truncate rounds both ways. Round parallax on parse.

rpiaggio:
- Make truncating validators round both ways (when parsing and when formatting).
- Parallax parser now rounds.


toddburnside (approved):
LGTM

PR #579, opened 2022-09-10
Add Display to Gid

rpiaggio:
`Gid` now acts as a `Display` instance.

This was omitted from `Uid` on purpose: I'm not sure we ever want to show a `UUID` to users in a UI. However, showing `Gid`s is fairly common.


cquiroz (approved):
LGTM

PR #580, opened 2022-09-11
Time brightness

cquiroz:
This is a proposal for adding a time dimension to Magnitudes
Basically it replaces measurement values with `NonEmptyMap[Timestamp, BrightnessMeasure[T]]`

`NonEmptyMap` is backed by a `SortedMap` thus we'd keep the entries in order and we need at least one, signalling the common case where the value is constant

There are a few things missing, like interpolation of values and lenses but wanted to request a first impression on the impact of this change.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/BrightnessUnits.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/BrightnessUnits.scala
@@ -6,9 +6,11 @@ package core
 package math
 
 import cats.data.NonEmptyList
+import cats.data.NonEmptyMap
 import lucuma.core.math.dimensional._
 import lucuma.core.math.units._
 import lucuma.core.util.Enumerated
+import lucuma.core.util.Timestamp
 
 object BrightnessUnits {
   type Integrated
@@ -19,6 +21,7 @@ object BrightnessUnits {
   trait FluxDensityContinuum[+T]
 
   type BrightnessMeasure[T] = Measure[BigDecimal] Of Brightness[T]
+  type BrightnessMeasureOverTime[T] = NonEmptyMap[Timestamp, Measure[BigDecimal] Of Brightness[T]]
 
   // Brightness Integrated
   implicit object VegaMagnitudeIsIntegratedBrightnessUnit
--- a/modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
@@ -4,6 +4,7 @@
 package lucuma.core.model
 
 import cats.Eq
+import cats.data.NonEmptyMap
 import cats.implicits._
 import coulomb.*
 import coulomb.ops.algebra.cats.all.given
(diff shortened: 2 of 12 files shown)


tpolecat:
Seems reasonable to me. Due to the similarity with time-varying coordinates it will be tempting to abstract out `TImeVarying[A]`, but I think we should wait on doing so until we're happy with the way everything is working (by that point it may not make sense to generalize).

modules/core/shared/src/main/scala/lucuma/core/math/BrightnessUnits.scala
```diff
@@ -19,6 +22,7 @@ object BrightnessUnits {
   trait FluxDensityContinuum[+T]
 
   type BrightnessMeasure[T] = Measure[BigDecimal] Of Brightness[T]
+  type BrightnessMeasureOverTime[T] = NonEmptyMap[Instant, Measure[BigDecimal] Of Brightness[T]]
```
swalker2m, line 25: I suppose we could use the `BrightnessMeasure` alias here.

modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
```diff
@@ -4,21 +4,24 @@
...
+import lucuma.core.math.units.*
 import monocle.Focus
 import monocle.Lens
 
+import java.time.Instant
+
 final case class EmissionLine[T](
   lineWidth: Quantity[PosBigDecimal, KilometersPerSecond],
-  lineFlux:  Measure[PosBigDecimal] Of LineFlux[T]
+  lineFlux:  EmissionLine.LineFluxOverTime[T]
```
swalker2m, line 24: I don't know anything about this unfortunately.  It makes sense that the "line flux" would vary but are we sure that it does?

modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
```diff
@@ -28,18 +31,20 @@ final case class EmissionLine[T](
    *   `Integrated` or `Surface`
    */
   def to[T0](implicit conv: TagConverter[LineFlux[T], LineFlux[T0]]): EmissionLine[T0] =
-    EmissionLine[T0](lineWidth, lineFlux.toTag[LineFlux[T0]])
+    EmissionLine[T0](lineWidth, lineFlux.map(_.toTag[LineFlux[T0]]))
 }
 
 object EmissionLine {
-  implicit def eqEmissionLine[T]: Eq[EmissionLine[T]] =
+  type LineFluxOverTime[T] = NonEmptyMap[Instant, Measure[PosBigDecimal] Of LineFlux[T]]
```
swalker2m, line 38: I wondered why we don't add a `LineFluxMeasure` to parallel the `BrightnessMeasure`.

swalker2m (approved):
Source profile is already so sophisticated that it pains me to add more but we do need to match the reality of the situation.  We'll need to use the time dimension for AGS magnitude look ups and, presumably, ITC calculations so this will ripple out a bit.

modules/core/shared/src/main/scala/lucuma/core/math/BrightnessUnits.scala
cquiroz: Sounds like a good idea, to reduce confusion a bit

modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
cquiroz: I'll verify, my input from science is that emission lines can also vary over time

modules/core/shared/src/main/scala/lucuma/core/model/EmissionLine.scala
cquiroz: @rpiaggio ?

PR #582, opened 2022-09-11
Unify tracking methods for Sidereal and Nonsidereal targets

cquiroz:
We have separated ways to express tracking for sidereal and non sidereals and I think they could be unified.

This is a proposal on how we could do it but I have some questions though as I'm unfamiliar with how non-sidereals are handled, see some comments inline. I'd appreciate any insights.

This would help with asterisms handling in explore as well, basically I'd like for the generic `Target` trait to have a way to express that the coordinates could vary on time.

I also switched to use `Timestamp` on `SiderealTracking` to make it consistent with `Ephemeris`

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
@@ -51,8 +51,8 @@ final case class SiderealTracking(
   parallax:        Option[Parallax]
 ) {
 
-  def at(i: Timestamp): Option[Coordinates] =
-    plusYears(epoch.untilInstant(i.toInstant))
+  def at(i: Instant): Option[Coordinates] =
+    plusYears(epoch.untilInstant(i))
 
   /** Coordinates `elapsedYears` fractional epoch-years after `epoch`. */
   def plusYears(elapsedYears: Double): Option[Coordinates] =
--- a/modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
@@ -7,6 +7,7 @@ import cats._
 import cats.syntax.all._
 import coulomb.policy.spire.standard.given
 import lucuma.core.math._
+import lucuma.core.util.Timestamp
 import monocle.Focus
 import monocle.Lens
 
@@ -50,8 +51,8 @@ final case class SiderealTracking(
   parallax:        Option[Parallax]
 ) {
 
-  def at(i: Instant): Option[Coordinates] =
-    plusYears(epoch.untilInstant(i))
+  def at(i: Timestamp): Option[Coordinates] =
+    plusYears(epoch.untilInstant(i.toInstant))
 
   /** Coordinates `elapsedYears` fractional epoch-years after `epoch`. */
   def plusYears(elapsedYears: Double): Option[Coordinates] =
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -41,6 +41,7 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     catalogInfo:   Option[CatalogInfo]
   ) extends Target {
     def at(i: Instant) = tracking.at(i).map(SiderealCoordinates.apply)
+    def atEpoch = SiderealCoordinates(tracking.baseCoordinates)
   }
 
   object Sidereal extends SiderealOptics {
@@ -69,6 +70,7 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     sourceProfile: SourceProfile
   ) extends Target {
     def at(i: Instant) = none
+    def atEpoch = EphemerisCoordinates(Coordinates.Zero, Offset.Zero)
   }
 
   object Nonsidereal extends NonsiderealOptics {
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -23,6 +23,7 @@ import monocle.Prism
 import monocle.Traversal
 import monocle.macros.GenPrism
 
+import java.time.Instant
 import scala.collection.immutable.SortedMap
 
 /** A target of observation. */
@@ -39,7 +40,7 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     sourceProfile: SourceProfile,
     catalogInfo:   Option[CatalogInfo]
   ) extends Target {
-    def at(i: Timestamp) = tracking.at(i).map(SiderealCoordinates.apply)
+    def at(i: Instant) = tracking.at(i).map(SiderealCoordinates.apply)
   }
 
   object Sidereal extends SiderealOptics {
@@ -67,7 +68,7 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     ephemerisKey:  EphemerisKey,
     sourceProfile: SourceProfile
   ) extends Target {
-    def at(i: Timestamp) = none
+    def at(i: Instant) = none
   }
 
   object Nonsidereal extends NonsiderealOptics {
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -13,6 +13,7 @@ import lucuma.core.enums.Band
 import lucuma.core.math.BrightnessUnits._
 import lucuma.core.math._
 import lucuma.core.math.dimensional._
+import lucuma.core.util.Timestamp
 import lucuma.core.util.WithGid
 import lucuma.refined._
 import monocle.Focus
@@ -25,7 +26,7 @@ import monocle.macros.GenPrism
 import scala.collection.immutable.SortedMap
 
 /** A target of observation. */
-sealed trait Target extends Product with Serializable {
+sealed trait Target extends Tracking with Product with Serializable {
   def name: NonEmptyString
   def sourceProfile: SourceProfile
 }
@@ -37,7 +38,9 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     tracking:      SiderealTracking,
     sourceProfile: SourceProfile,
     catalogInfo:   Option[CatalogInfo]
-  ) extends Target
+  ) extends Target {
+    def at(i: Timestamp) = tracking.at(i).map(SiderealCoordinates.apply)
+  }
 
   object Sidereal extends SiderealOptics {
     implicit val eqSidereal: Eq[Sidereal] =
@@ -63,7 +66,9 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     name:          NonEmptyString,
     ephemerisKey:  EphemerisKey,
     sourceProfile: SourceProfile
-  ) extends Target
+  ) extends Target {
+    def at(i: Timestamp) = none
+  }
 
   object Nonsidereal extends NonsiderealOptics {
     implicit val eqNonsidereal: Eq[Nonsidereal] =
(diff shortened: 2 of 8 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
```diff
@@ -50,8 +51,8 @@ final case class SiderealTracking(
   parallax:        Option[Parallax]
 ) {
 
-  def at(i: Instant): Option[Coordinates] =
-    plusYears(epoch.untilInstant(i))
+  def at(i: Timestamp): Option[Coordinates] =
```
cquiroz, line 54: We are using `Timestamp` in `Ephemeris`. Do we want it here too?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -63,7 +66,9 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     name:          NonEmptyString,
     ephemerisKey:  EphemerisKey,
     sourceProfile: SourceProfile
-  ) extends Target
+  ) extends Target {
+    def at(i: Timestamp) = none
```
cquiroz, line 70: I'm unfamiliar with how to get the `EphemerisCoords` out of `EphemerisKey`. Maybe this actual needs to be
`def at(i: Timestamp): F[TrackedCoordinates]` to take into account that they may come from a remote DB?

modules/core/shared/src/main/scala/lucuma/core/model/Tracking.scala
```diff
@@ -3,11 +3,24 @@
...
 
+/** A coordinate along with a rate of change in RA and Dec for some time unit
+  *
+  * @param coord coordinates
+  * @param delta rate of change in RA and dec, where the delta(RA)/time has been
+  *              multiplied by the cosine of the dec
+  */
+sealed trait TrackedCoordinates {
+  def coord: Coordinates
+  def delta: Offset /* per time unit */
```
cquiroz, line 21: Would we rather express this as `AngularVelocity` to make it clear the time component?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
rpiaggio: I'm unfamiliar too, but if it's some kind of extrapolation to table values, I think it's safe to keep `Instant` in this interface.

modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
rpiaggio: This value is not saved in the DB. Maybe we should keep `Instant` to simplify the API?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
swalker2m: I assume we'll have a robot that keeps non-sidereal coordinates up-to-date like we do today.  I don't think the intention is to reach out to Horizons on demand.

modules/core/shared/src/main/scala/lucuma/core/syntax/Time.scala
```diff
@@ -38,7 +39,10 @@ trait ToInstantOps {
     new InstantOps(i)
 }
 
-object instant extends ToInstantOps
+object instant extends ToInstantOps {
+  extension(i: Instant)
+    def toTimestamp: Option[Timestamp] = Timestamp.fromInstant(i)
```
swalker2m, line 44: I suppose we should add (or else just always use) a version that truncates to microseconds.  `fromInstantTruncated` or something like that it is called.

modules/core/shared/src/main/scala/lucuma/core/model/Tracking.scala
swalker2m: It seems like, for `delta` to be generically useful regardless of target type, you'd have to know the time unit in question.

modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
swalker2m: I suppose the `Tracking` interface could provide an `atTimestamp` that calls the instant version.  Otoh, requiring the caller who has a `Timestamp` for some reason to do `at(t.toInstant)` isn't so bad?

cquiroz: I want that both sidereal and non-sidereals have the same interface. my impression is that it should be `Timestamp` rather than instant, is that right?

modules/core/shared/src/main/scala/lucuma/core/model/Tracking.scala
```diff
@@ -80,3 +94,33 @@ trait EphemerisCoordinatesOptics {
     delta.andThen(Offset.q)
 
 }
+
+case class SiderealCoordinates(override val coord: Coordinates) extends TrackedCoordinates {
+  override val delta: Offset =  Offset.Zero
```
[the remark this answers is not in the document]
cquiroz: This is only to unify the interface, I think the delta is 0 for sidereals at least on the usual cases. I guess we could remove it and only have it on `EphemerisCoordinates`?

modules/core/shared/src/main/scala/lucuma/core/model/SiderealTracking.scala
swalker2m: As @rpiaggio pointed out, this isn't about a time that is kept in the database.  `Timestamp` is more about being able to round-trip values.  I think I would tend to keep the coordinate lookup using `Instant`.  Non-sidereal ephemeris data will be stored at microsecond precision but we should be able to estimate coordinates any `Instant` covered by that data.  If you want to provide both though (and just have the timestamp version convert itself to an instant and call the instant version), that seems fine to me.

cquiroz: ok, that sounds good. I learned you can't add another `at(i: Instant)` because `Timestamp` is an opaque but we could do `atTimestamp` or just let it off

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
rpiaggio: Just out of curiosity: how is it done today? Where does the robot store the non-sidereal coordinates?

swalker2m: In the `ocs` codebase ODB, programs are mutable tree structures stored in memory (java-serialized to disk when modified).  Each non-sidereal observation references a relatively low-precision ephemeris table that covers the semester plus a month (or so) on either side. This is obtained from Horizons when the non-sidereal target is first resolved.

In addition, there is a robot that checks hourly for active observations, obtains high-precision ephemeris data for the upcoming or current observing night and stores them as text files in a place and format that the TCS can read and use for tracking.

cquiroz: Do we have plans to do something like that in the odb? We'd update the ephemeris table and the odb will give us the current table on the graphql response?

swalker2m: At least at the time the project was first coming together we suggested a target service separate from the ODB that would maintain this type of information.  Regardless I think ephemerides belong in a (relational) database table that is kept up-to-date automatically.  Whether to put them in graphql responses or transfer the data in some more efficient way (or take advantage of the client being able to make the same Horizons queries?) I don't know.  What do you guys think?

cquiroz: I don't have a strong opinion. I guess engage will need to query the ephemeris eventually though again it could do it directly.
Do we need API users to get ephemeris?

rpiaggio: When is it that the non-sidereal target first resolved?

Does the high-precision ephemeris data come from Horizons too?

Could engage and explore query Horizons directly? Is there a requirement that resulted in the robot writing a text file instead of the TCS querying Horizons?

modules/core/shared/src/main/scala/lucuma/core/model/Tracking.scala
cquiroz: @tpolecat I think you're the expert in Horizons, would you shed some light about the delta component? Is it required for the TCC?

tpolecat: I don't have any specific insight here. Horizons returns velocities in arcsrc/hr IIRC so we're storing them in OCS, but I don't know if they're used for guiding. We're doing our own interpolation and I don't know if these values are used at all.

tpolecat: But in ay case yes, if we're storing this it should be a velocity rather than a scalar.

swalker2m: If I recall correctly, we put them into the ephemerides files and the TCS uses them.

swalker2m: Yes, the TCS expects to see it at least but I don't know whether it takes it into account.

https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.spdb.reports.collection/src/main/scala/edu/gemini/dbTools/ephemeris/EphemerisFileFormat.scala#L65-L67

modules/core/shared/src/main/scala/lucuma/core/model/Tracking.scala
```diff
@@ -80,3 +95,38 @@ trait EphemerisCoordinatesOptics {
...
+  val declination: Lens[SiderealCoordinates, Declination] =
+    coordinates.andThen(Coordinates.declination)
+}
+
+/**
+ * Implementors can calculate the coordinates for a given object at a certain timestamp
+ */
+trait Tracking {
+  def at(i: Instant): Option[TrackedCoordinates]
+  def atEpoch: TrackedCoordinates
```
cquiroz, line 124: I added this method to give use the coordinates at epoch. For sidereals is fairly straightforward but for non-sidereals I'd use the convention that epoch would be the first item on the ephemeris.
Note this implies that we need to have at least one entry on the ephemeris table
I discussed this with Andy and he agrees that we should not have non-sidereal targets without ephemeris but you may disagree

An alternative is to make the signature
```
def atEpoch: Option[TrackedCoordinates]
```
But it does complicate a bit the code downstream where it is simpler if we can avoid the `Option`
WDYT?

modules/core/shared/src/main/scala/lucuma/core/model/Tracking.scala
```diff
@@ -3,11 +3,25 @@
...
+
+/** A coordinate along with a rate of change in RA and Dec for some time unit
+  *
+  * @param coord coordinates
+  * @param delta rate of change in RA and dec, where the delta(RA)/time has been
+  *              multiplied by the cosine of the dec
+  */
+sealed trait TrackedCoordinates {
+  def coord: Coordinates
+}
```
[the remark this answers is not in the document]
cquiroz: `Non-Sidereals` would produce a different `TrackedCoordinates` containing also the `delta` param

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -63,7 +68,10 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     name:          NonEmptyString,
     ephemerisKey:  EphemerisKey,
     sourceProfile: SourceProfile
-  ) extends Target
+  ) extends Target {
+    def at(i: Instant) = none
+    def atEpoch = EphemerisCoordinates(Coordinates.Zero, Offset.Zero)
```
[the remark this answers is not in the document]
cquiroz: Right. I considered the type class approach. It would be more flexible and avoid having an `F` on model classes.
I could experiment with that approach a bit

cquiroz:
This has been a good discussion and helped me clarify some concepts.
I'm going to close this and will implement a different solution. I wasn't too happy adding the extra interface to targets but we can work around it

PR #591, opened 2022-09-26
Inline NewType.apply

rpiaggio:
This avoids boxing and unboxing of primitive types when defining `NewType` values.

Eg, when defining `Switch extends NewType[Boolean]`, the code `val foo = Switch(false)` before compiled to:
```
foo = BoxesRunTime.unboxToBoolean(Switch.apply(BoxesRunTime.boxToBoolean(false)));
```

and now:
```
foo = false
```


cquiroz (approved):
LGTM

PR #597, opened 2022-10-18
Add csv as a possible catalog name

cquiroz:
Are we ok changing the old enums to scala 3 `enum`?


toddburnside:
> Are we ok changing the old enums to scala 3 `enum`?

Is this one of the enums that will eventually be filled by the database? In which case it probably doesn't matter since it will need to be changed later.

toddburnside (approved):
LGTM

cquiroz:
I see them more like three different sources where we get sidereal data info. They e.g. provide brightness and/or source profile
The main reason to have them is to have a record in the db of where they came from, though if we are not storing the GS we may not need gaia.

PR #599, opened 2022-10-24
Switch angle parsing to cats-parse

cquiroz:
This switches angle/RA/DEC/Coordinates parsers to cats-parse and also lets them parse angles with an arbitrary amount of decimals (though some maybe truncated)


modules/tests/shared/src/test/scala/lucuma/core/math/parser/AngleParsersSuite.scala
```diff
@@ -0,0 +1,60 @@
...
+package lucuma.core.math.parser
+
+import cats.kernel.laws.discipline.*
+import lucuma.core.math.Declination
+import lucuma.core.math.arb.ArbAngle
+import lucuma.core.math.arb.ArbRightAscension.*
+import lucuma.core.optics.laws.discipline.FormatTests
+import munit.*
+
+class AngleParsersSuite extends DisciplineSuite with AngleParsers:
```
rpiaggio, line 13: Maybe we can test that the returned values are the expected ones?

rpiaggio (approved):
I haven't used cats-parse either, so can't vouch for its usage. Otherwise, LGTM.

modules/tests/shared/src/test/scala/lucuma/core/math/parser/AngleParsersSuite.scala
cquiroz: I updated the tests to do actual comparisons

PR #600, opened 2022-10-24
unNone extension method

rpiaggio:


modules/core/shared/src/main/scala/lucuma/core/syntax/IterableOps.scala
```diff
@@ -0,0 +1,14 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.syntax
+
+import scala.collection.{IterableOps => ScalaIterableOps}
+
+trait IterableOps:
+  extension [S[_], A](aa:  ScalaIterableOps[Option[A], S, S[Option[A]]])
+    /** Idiomatic way of reducing an `Iterable[Option[A]]` to `Iterable[A]` */
+    def unNone: S[A] =  aa.flatten
```
cquiroz, line 11: Isn't this called `flattenOption` in `List`. Should use that name?

swalker2m: For that matter is it maybe covered by `FilterFunctor`?

rpiaggio: Yup, it seems to be redundant.

I liked the `unNone` name: it's shorter, more specific and consistent with `fs2.Stream`'s, but it would introduce a 3rd way of achieving the same thing ( there's already `flatten` and `flattenOption`).

PR #606, opened 2022-10-30
Port parsers to cats-parse

cquiroz:
I ported the minimal to get the test running, Some unused parsers are not included but could be added when needed


rpiaggio (approved):
LGTM

PR #609, opened 2022-11-04
Remove duplicated parser and migrate to extension methods

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/validation/package.scala
```diff
@@ -34,82 +34,64 @@ package object validation {
   type ValidFormatNec[E, A, B] = ValidFormat[NonEmptyChain[E], A, B]
   type InputValidFormat[A]     = ValidFormatNec[NonEmptyString, String, A]
 
-  implicit class StringParseOps(private val s: String) extends AnyVal {
-
-    /** Try to parse as a `BigDecimal` */
-    def toBigDecimalOption: Option[BigDecimal] =
```
cquiroz, line 40: This was already defined in string syntax

modules/core/shared/src/main/scala/lucuma/core/validation/package.scala
```diff
@@ -34,82 +34,64 @@ package object validation {
...
-  implicit class InputValidWedgeOps[A](private val self: InputValidWedge[A]) extends AnyVal {
+  extension[A](self: InputValidWedge[A])
     def withErrorMessage(msg: NonEmptyString): InputValidWedge[A] =
       self.withError(NonEmptyChain(msg))
 
     /**
      * Build an `InputValidWedge[NonEmptyList[A]]` given a `InputValidWedge[A]`
      */
-    def toNel(separator: NonEmptyString = ",".refined): InputValidWedge[NonEmptyList[A]] =
+    def toNel(separator: NonEmptyString): InputValidWedge[NonEmptyList[A]] =
```
cquiroz, line 78: I had to remove the default param or it would collide with the other `toNel` method. Is this a problem?

swalker2m: Which other `toNel`?

swalker2m (approved):
This looks good to me. 👍

modules/core/shared/src/main/scala/lucuma/core/validation/package.scala
rpiaggio: I'm guessing the one for `InputValidSplitEpi`.

@cquiroz It's not a problem but it's kind of nice to have to pass the (refined) parameter when it's a comma. What if you use `@targetName` ?

cquiroz: It doesn't make a difference to have a targetName

PR #612, opened 2022-11-07
Make seconds parser public

cquiroz:
In hindsight I should have made the `seconds` parser public


modules/core/shared/src/main/scala/lucuma/core/math/parser/AngleParsers.scala
```diff
@@ -41,7 +39,7 @@ trait AngleParsers:
     .map(_.toInt)
     .withContext("minutes")
 
-  private[parser] val seconds =
+  val seconds =
```
swalker2m, line 42: Maybe add the type declaration since it is public?

swalker2m (approved):
Looks good.  :+1:

PR #614, opened 2022-11-07
Make AngularVelocityComponent an opaque type

cquiroz:
This is a small experiment converting `AngularVelocityComponent` to an opaque type. I think several core classes could become opaque types including angle and Wavelength.
This class is particularly good for testing because we already have a benchmark.
I tested js size (using the test js ouput) and found a small reduction in size and the benchmark showed an increase from ~5500 ops/sec to ~5900 ops/sec

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala
@@ -11,9 +11,11 @@ import coulomb.ops.algebra.cats.all.given
 import coulomb.policy.spire.standard.given
 import coulomb.syntax.*
 import coulomb.units.accepted.*
-import lucuma.core.math.ProperMotion.AngularVelocityComponent
+import lucuma.core.math.ProperMotion.AngularVelocity
+import lucuma.core.math.dimensional.*
 import lucuma.core.math.units.*
 import lucuma.core.optics.SplitMono
+import lucuma.core.util.NewType
 import monocle.Focus
 import monocle.Iso
 import monocle.Lens
@@ -27,10 +29,7 @@ object VelocityAxis {
 /**
   * ProperMotion contains Ra/Dec angular velocities
   */
-final case class ProperMotion(
-  ra:  ProperMotion.AngularVelocityComponent[VelocityAxis.RA],
-  dec: ProperMotion.AngularVelocityComponent[VelocityAxis.Dec]
-) {
+case class ProperMotion( ra:  ProperMotion.RA, dec: ProperMotion.Dec) {
   // Return the ra/dec components in radians, first converting to degrees/y
   inline def toRadians: (Double, Double) =
     (ra.toRadians, dec.toRadians)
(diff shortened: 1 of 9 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala
```diff
@@ -32,36 +32,40 @@ final case class ProperMotion(
...
-    val masy: Quantity[Rational, MilliArcSecondPerYear] = μasy.toValue[Rational].toUnit[MilliArcSecondPerYear]
+    def masy: Quantity[Rational, MilliArcSecondPerYear] = μasy.toValue[Rational].toUnit[MilliArcSecondPerYear]
 
-    override def toString =
+    def toString: String =
       s"AngularVelocityComponent(${masy.show})"
-  }
 
   object AngularVelocityComponent extends AngularVelocityComponentOptics {
+    inline def apply[A](w: AngularVelocityComponent[A]): AngularVelocityComponent[A] = w
```
rpiaggio, line 54: Is this used to build the opaque type? Should the parameter `w` be a `Quantity[Long, MicroArcSecondPerYear]` ?

cquiroz: I think you're right. This works here because they are in the same file. I'll change it

modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala
```diff
@@ -32,36 +32,40 @@ final case class ProperMotion(
...
   // Return the ra/dec components in radians, first converting to degrees/y
-  def toRadians: (Double, Double) =
+  inline def toRadians: (Double, Double) =
     (ra.toRadians, dec.toRadians)
 
 }
 
 object ProperMotion extends ProperMotionOptics {
-  final case class AngularVelocityComponent[A](μasy: Quantity[Long, MicroArcSecondPerYear]) {
+  opaque type AngularVelocityComponent[A] = Quantity[Long, MicroArcSecondPerYear]
```
rpiaggio, line 41: Here's an idea, since we are revising the model: `A` seems to be just a tag.

What if we use a `NewType` for `AngularVelocityComponent` (without `A`) and then we use tagged types in `ProperMotion`?

The. new type can be just `AngularVelocity`, and it would look something like this: `case class ProperMotion(ra: AngularVelocity Of VelocityAxis.RA, dec: AngularVelocity Of VelocityAxis.Dec)`.

cquiroz: I'm not so sure what the `A` is doing there and in fact it is not limited to RA/Dec thus you could tag anything.
Could we have two newtypes
```object AngularVelocityRA extends NewType[Quantity[Long, MicroArc...]]]
object AngularVelocityDec extends NewType[Quantity[Long, MicroArc...]]]```

is the `Of` mechanism using Shapeless tags?

Comment (cquiroz) [reply modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala:41]: Isn't `Of` or a shapeless tag essentially what an `opaque type` would do?

Comment (rpiaggio) [reply modules/core/shared/src/main/scala/lucuma/core/math/ProperMotion.scala:41]: `A Of T` is actually an opaque type over `T` that is also a subtype of `T`:

``` scala
opaque infix type Of[+T, U] <: T = T
```

You can turn an `a: A` into an `A Of T` by calling `tag[T](a)`.

The difference with a non-generic opaque type is that the type `A` is exposed.

Also the tag `T` is an independent type that can be propagated to tag other types. This is not necessary here, but it's what is done to relate tagged `Units` and `Measure`.

If we use `Of` here (or anywhere else), we should move it out of the `math.dimensional` package and into `util` maybe?

cquiroz:
I switched to use the idea of `A Of T` for `AngularVelocity`, let me know what you think

rpiaggio (approved):
LGTM

PR #622, opened 2022-11-22
Fixes an edge case revelead on a test failure

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/math/validation/MathValidators.scala
```diff
@@ -53,7 +53,7 @@ trait MathValidators {
       a =>
         f"${truncateAngle(a).toSignedBigDecimalDegrees}%.2f"
           .replace("-0.00", "0.00")
-          .replace("180.00", "-180.00")
+          .replace("$180.00", "-180.00")
```
toddburnside, line 56: What is the `$` for? Is there actually a `$` in the original string?

toddburnside: Ah. I guess the old replace was changing "-180.00" to "--180.00"? But, the new replace doesn't seem to fix that, to me.

cquiroz: exactly the `--180` triggered a test error. `replace` takes a regex

toddburnside: Ah. OK. I didn't know `replace` takes a regex. But, shouldn't the `$` be a `^`?

toddburnside (approved):
LGTM

PR #626, opened 2022-11-28
Convert Angle to an opaque type

cquiroz:
Angle is essentially a wrapper for `Long` thus can become an `opaque type`
This PR does the conversion for `Angle` and `HourAngle`

There is a nice benchamark bump from 5200 ops/sec to 6030 ops/sec

Size on the js test suite increased a bit though. around 0.1%

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
@@ -13,6 +13,7 @@ import lucuma.core.optics.*
 import monocle.Iso
 import monocle.Prism
 
+import scala.annotation.targetName
 import scala.math
 
 /**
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
```diff
@@ -23,175 +24,15 @@ import scala.math
...
-  /**
-   * angle difference or separation between two angles in the range [0 .. π]
-   */
-  def difference(a: Angle): Angle =
-    Angle.difference(this, a)
-}
+opaque type Angle = Long
 
 object Angle extends AngleOptics {
+  inline private def apply(w: Long): Angle = w
```
swalker2m, line 34: Could/should we keep the `0 <= toMicroarcseconds < 360º` sanity check? That is, `Angle` isn't just any `Long`, it has to be microarcseconds and in range.  I think this is still guaranteed by the constructors but it seems like it would be easy to slip in a non-valid constructor at some point.

cquiroz: I wasn't sure about that but we could restore it. I'm not sure if we can inline in that case

rpiaggio: Maybe we should `Refine` it?

swalker2m: > Maybe we should Refine it?

That sounds right, I think.  We wrote `Angle` long before we embraced refine.

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
```diff
@@ -487,65 +485,12 @@ trait AngleOptics extends OpticsHelpers { this: Angle.type =>
...
-   */
-  def -(ha: HourAngle): HourAngle =
-    HourAngle.fromMicroseconds(toMicroseconds - ha.toMicroseconds)
-
-  /** String representation of this HourAngle, for debugging purposes only. */
-  override def toString: String =
-    HourAngle.fromStringHMS.taggedToString("HourAngle", this)
-
-}
+opaque type HourAngle <: Angle = Long
```
cquiroz, line 488: In the past `HourAngle` extended `Angle` presumably to reuse some of the methods though HourAngle requires the value to be divisible by 15.
I suppose we may want to pass `HourAngle` to methods taking `Angle?

swalker2m: Cool, I didn't know you could do that in an `opaque type` definition.  What happens now  when you try to, say, add an `Angle` to an `HourAngle`?  Is it any different from what happened in the previous version?

cquiroz: It should work. I'll check if there is a test and add one if not

swalker2m:
> Maybe just restore the assertions on the constructors?

Seems fine to me.

swalker2m (approved):
Looks good. :+1:

PR #629, opened 2022-12-07
Make SED optional

cquiroz:
Currently we require an SED which in many cases we don't know and thus we default to e.g. A0V which is missleading

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
@@ -3,18 +3,18 @@
 
 package lucuma.core.model
 
-import cats._
-import cats.implicits._
-import eu.timepit.refined.auto._
-import eu.timepit.refined.cats._
+import cats.*
+import cats.implicits.*
+import eu.timepit.refined.auto.*
+import eu.timepit.refined.cats.*
 import eu.timepit.refined.types.numeric.PosBigDecimal
 import eu.timepit.refined.types.string.NonEmptyString
 import lucuma.core.enums.Band
-import lucuma.core.math.BrightnessUnits._
-import lucuma.core.math._
-import lucuma.core.math.dimensional._
+import lucuma.core.math.BrightnessUnits.*
+import lucuma.core.math.*
+import lucuma.core.math.dimensional.*
 import lucuma.core.util.WithGid
-import lucuma.refined._
+import lucuma.refined.*
 import monocle.Focus
 import monocle.Lens
 import monocle.Optional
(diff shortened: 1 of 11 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -59,14 +59,14 @@ object Target extends WithGid('t'.refined) with TargetOptics {
     val NameOrder: Order[Sidereal] = Order.by(x => (x.name, x.tracking))
   }
 
-  final case class Nonsidereal(
+  case class Nonsidereal(
```
swalker2m, line 62: Is it no longer possible to extend `case class`es?

swalker2m (approved):
This looks good to me.  Maybe @rpiaggio has more insight though?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
cquiroz: In scala 3 you cannot extend classes unless yoou specify them with the `open` keyword
https://docs.scala-lang.org/scala3/reference/other-new-features/open-classes.html

You can still extend in particular cases (e.g. tests) but by default you can't unless explicitly enabled in the code

cquiroz: In general I`ve been removing final in scala 3 code.
About `-source feature` maybe @armanbilge knows more about it?

rpiaggio:
I have the feeling we actively decided to not make it optional at some point. Let me dig through emails and chats.

cquiroz:
I had a discussion with users and we came up with this ticket:
https://app.shortcut.com/lucuma/story/1852/no-default-target-morphology-or-sed

Essentially we assign targets a morphology which in most cases is incorrect. By making it optional we force the user to think about it and set one

andrewwstephens:
The ITC relies on knowing the morphology and SED, and getting it wrong will affect the generated sequence.  I'm hoping that at some point in the future we can divine these properties from online catalogs, e.g. Simbad and NED, but even then we will never get 100% of targets and the user will need to fill in the blanks.

swalker2m:
> I have the feeling we actively decided to not make it optional at some point.

If I recall correctly, that discussion was about the `SourceProfile` as a whole.

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
armanbilge: Btw, even though the _compiler_ enforces the `open` keyword, unless you make the class `final` it will not be `final` in the bytecode. This may have some (small) performance implications.

rpiaggio:
> If I recall correctly, that discussion was about the `SourceProfile` as a whole.

Yup, that was the case!

PR #638, opened 2023-01-06
WavelengthDither

swalker2m:
* Adds a `WavelengthDither` type, which will be used in GMOS long slit mode.  `WavelengthDither` is like `Wavelength` (a length stored as an `Int` with picometer units) but represents an _offset_ from a wavelength.  As such it may be zero, positive or negative whereas `Wavelength` is strictly positive.
* Converts `Wavelength` from a wrapper class to an `opaque type`.
* Standardizes the naming across the two types.


modules/core/shared/src/main/scala/lucuma/core/math/Wavelength.scala
```diff
@@ -29,129 +30,162 @@ import scala.util.Try
   * which means the largest representable wavelength is 2.147483647 mm.
   * @param toPicometers This wavelength in positive integral picometers (10^-12^ of a meter).
   */
-final case class Wavelength(toPicometers: Quantity[PosInt, Picometer]) {
+opaque type Wavelength = Quantity[PosInt, Picometer]
```
cquiroz, line 33: Cool

modules/core/shared/src/main/scala/lucuma/core/math/WavelengthDither.scala
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * A wavelength dither is a small offset (positive or negative) intended to be
+ * applied to a wavelength, such as the observing wavelength.  It is therefore a
+ * small distance and not a `Wavelength` per se.  A `Wavelength + WavelengthDither`
+ * should produce a new `Wavelength`.
+ */
+opaque type WavelengthDither = Quantity[Int, Picometer]
+
+object WavelengthDither {
```
cquiroz, line 26: Would you ever convert `Waveleng` to `WavelengthDither`?

modules/tests/shared/src/test/scala/lucuma/core/math/WavelengthSuite.scala
```diff
@@ -6,44 +6,61 @@ package lucuma.core.math
...
-import org.scalacheck.Prop._
+import org.scalacheck.Prop.*
 import spire.math.Rational
 
 import java.math.RoundingMode
+import scala.language.implicitConversions
 
 final class WavelengthSuite extends munit.DisciplineSuite {
-  import ArbWavelength._
+  //  import ArbQuantity._   Why won't this work?
```
cquiroz, line 34: commented line?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/WavelengthDither.scala
swalker2m: It doesn't occur to me when you ever would have a `Wavelength` value that serves as an offset.

modules/tests/shared/src/test/scala/lucuma/core/math/WavelengthSuite.scala
swalker2m: Thanks, removed.  I'm still unsure why it didn't work but content to let the mystery be.

cquiroz: I think I've seen that before where you need to import each member and the wildcard doesn't work

PR #646, opened 2023-01-18
Add `Unbounded` to PosAngleConstraint

toddburnside:
Will be merged as part of the grand transition to using the new ODB.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/PosAngleConstraint.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/PosAngleConstraint.scala
@@ -21,6 +21,22 @@ sealed trait PosAngleConstraint extends Product with Serializable
 object PosAngleConstraint extends PosAngleConstraintOptics {
 
   /**
+   * Specifies that the there is no constraint on the position angle.
+   * It may be set to any angle for this observation.
+   */
+  case object Unbounded extends PosAngleConstraint {
+    override def toString: String = "Unbounded"
+  }
+
+  /**
+   * Constructs an `Unbounded` instance with a wider `PosAngle` type.
+   *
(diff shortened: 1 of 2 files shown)


swalker2m (approved):
Looks good to me. :+1:  Seems like @tpolecat mentioned that there's no longer a need for widening constructors but I'm not certain.

modules/core/shared/src/main/scala/lucuma/core/model/PosAngleConstraint.scala
```diff
@@ -20,6 +20,22 @@ sealed trait PosAngleConstraint extends Product with Serializable
 
 object PosAngleConstraint extends PosAngleConstraintOptics {
 
+  /**
+   * Specifies that the there is no constraint on the position angle.
+   * It may be set to any angle for this observation.
+   */
+  case object Unbounded extends PosAngleConstraint {
```
cquiroz, line 27: We used to have in most places `Option[PosAngleConstraint]` where `None` would be used as `UNBOUNDED`
It maybe worth check on lucuma-core where this pattern has been used and fix it

toddburnside: I don't know of any places in lucuma-core, but I'll look. I have another pull request coming for `lucuma-catalog`. Plus, of course, the big PR for explore.

cquiroz: Right, we may encode that on catalog or explore

toddburnside: The encoder is in lucuma-schemas

PR #647, opened 2023-01-18
TimeSpan

swalker2m:
Adds a`TimeSpan` type that corresponds to the ODB GraphQL schema `NonNegDuration`.  My goal is 1:1 correspondence between the data type and a convenient representation in the GraphQL schema.   Note that `java.time.Duration` cannot always be represented as a `Long` in microseconds (which is the base type in the GraphQL schema) and there is already a canonical JSON codec for `Duration` that includes the second and nano adjustment.


modules/core/shared/src/main/scala/lucuma/core/enums/F2ReadMode.scala
```diff
@@ -18,18 +19,18 @@ sealed abstract class F2ReadMode(
   val shortName: String,
   val longName: String,
   val description: String,
-  val minimumExposureTime: FiniteDuration,
-  val recommendedExposureTime: FiniteDuration,
-  val readoutTime: FiniteDuration,
+  val minimumExposureTime: TimeSpan,
+  val recommendedExposureTime: TimeSpan,
+  val readoutTime: TimeSpan,
```
swalker2m, line 24: Replacing `FiniteDuration`, `Duration` and `NonNegDuration` where possible with `TimeSpan`.

modules/core/shared/src/main/scala/lucuma/core/enums/F2ReadMode.scala
```diff
@@ -18,18 +19,18 @@ sealed abstract class F2ReadMode(
...
   val readCount: Int,
   val readNoise: Double
 ) extends Product with Serializable
 
 object F2ReadMode {
 
-  /** @group Constructors */ case object Bright extends F2ReadMode("Bright", "bright", "Bright Object", "Strong Source", 1500.millis, 5000.millis, 8000.millis, 1, 11.7)
-  /** @group Constructors */ case object Medium extends F2ReadMode("Medium", "medium", "Medium Object", "Medium Source", 6000.millis, 21000.millis, 14000.millis, 4, 6.0)
-  /** @group Constructors */ case object Faint extends F2ReadMode("Faint", "faint", "Faint Object", "Weak Source", 12000.millis, 85000.millis, 20000.millis, 8, 5.0)
+  /** @group Constructors */ case object Bright extends F2ReadMode("Bright", "bright", "Bright Object", "Strong Source", 1500.msTimeSpan, 5000.msTimeSpan, 8000.msTimeSpan, 1, 11.7)
```
swalker2m, line 31: the `.msTimeSpan`, `.secTimeSpan` etc. check the value to guarantee that they are in range at compile time

modules/core/shared/src/main/scala/lucuma/core/model/sequence/DynamicConfig.scala
```diff
@@ -6,18 +6,17 @@ package lucuma.core.model.sequence
...
 import monocle.macros.GenPrism
 import org.typelevel.cats.time._
 
-import java.time.Duration
-
 sealed trait DynamicConfig
 object DynamicConfig {
   final case class GmosNorth(
-    exposure:      Duration,
+    exposure:      TimeSpan,
```
swalker2m, line 19: Previously we allowed negative `Duration` but with `TimeSpan` it must be non-negative.  I think we weren't intending to use negative exposure times for anything since it doesn't correspond to reality.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/StepTime.scala
```diff
@@ -4,41 +4,40 @@
...
-  configChange: Duration,
-  exposure:     Duration,
-  readout:      Duration,
-  write:        Duration,
-  total:        Duration
+  configChange: TimeSpan,
+  exposure:     TimeSpan,
+  readout:      TimeSpan,
+  write:        TimeSpan,
+  total:        TimeSpan
```
swalker2m, line 17: All of these will take at least 0 time or more.

modules/core/shared/src/main/scala/lucuma/core/syntax/timespan/package.scala
```diff
@@ -0,0 +1,84 @@
...
+
+package lucuma.core.syntax.timespan
+
+import lucuma.core.util.TimeSpan
+
+extension (inline i: Int) {
+
+  inline def microsecondTimeSpan: TimeSpan =
+    inline if (i >= 0) TimeSpan.fromMicroseconds(i).get
+    else scala.compiletime.error("TimeSpan must be non-negative.")
```
swalker2m, line 12: I may not be doing this right but I wanted to be able to safely create literal values like `20.secondTimeSpan`.

modules/core/shared/src/main/scala/lucuma/core/syntax/timespan/package.scala
```diff
@@ -0,0 +1,84 @@
...
+  inline def hourTimeSpan: TimeSpan =
+    inline if (i >= 0) TimeSpan.fromHours(BigDecimal(i)).get
+    else scala.compiletime.error("TimeSpan must be non-negative.")
+
+  inline def hrTimeSpan: TimeSpan =
+    hourTimeSpan
+
+}
+
+extension (inline l: Long) {
```
swalker2m, line 47: Tried to generalize this for anything with an `Integral` but couldn't quite swing it with the `inline`s.

swalker2m (commented):
Made a draft PR with changes in the schema to use `TimeSpan`: https://github.com/gemini-hlsw/lucuma-odb/pull/225

modules/core/shared/src/main/scala/lucuma/core/enums/F2ReadMode.scala
cquiroz: 👍

modules/core/shared/src/main/scala/lucuma/core/model/sequence/DynamicConfig.scala
cquiroz: is a 0 exposure time allowed?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/StepTime.scala
cquiroz: Would `total` be calculated from the other times?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/DynamicConfig.scala
swalker2m: yeah, biases for example are 0 seconds

modules/core/shared/src/main/scala/lucuma/core/model/sequence/StepTime.scala
swalker2m: actually yes, this was extracted from the schema where total is offered but maybe we don't need it here

modules/core/shared/src/main/scala/lucuma/core/util/TimeSpan.scala
```diff
@@ -0,0 +1,170 @@
...
+   * Constructs a TimeSpan from the given number of microseconds, if it is
+   * non-negative.
+   */
+  def fromMicroseconds(µs: Long): Option[TimeSpan] =
+    NonNegLong.from(µs).toOption
+
+  def unsafeFromMicroseconds(µs: Long): TimeSpan =
+    fromMicroseconds(µs).getOrElse(sys.error(s"The µs value ($µs) must be non-negative."))
+
+  def fromNonNegMicroseconds(µs: NonNegLong): TimeSpan =
```
cquiroz, line 47: I guess this could be `apply` or have another `apply` method?

cquiroz (approved):
LGTM, I only wonder if this will break explore while using tmp-api

modules/core/shared/src/main/scala/lucuma/core/model/sequence/DynamicConfig.scala
cquiroz: I've never understood that, especially that when I do a bias in the real world they do take time 😄

swalker2m: Ha, yes it still has to configure and readout presumably even if there is no exposure time.

modules/core/shared/src/main/scala/lucuma/core/syntax/timespan/package.scala
toddburnside: > Yes, I'm not sure what to do about that. If it is a random `Int` value it could be negative and technically it'd need an `Option` result or else it would be "unsafe".

I think that is the behavior you are looking for in order to be able to create TimeSpans from literals.

> Does this count as a macro? I'd hoped it stopped short of that.

No te preocupes. I don't think this counts as a macro. 😄

cquiroz: plain inlines are probably not technically macros

modules/core/shared/src/main/scala/lucuma/core/util/TimeSpan.scala
swalker2m: :+1: added `apply`.

swalker2m:
> LGTM, I only wonder if this will break explore while using tmp-api

You're right, it might be better to wait until we're switched over to `lucuma-odb`.

toddburnside (approved):
LGTM

toddburnside:
> > LGTM, I only wonder if this will break explore while using tmp-api
> 
> You're right, it might be better to wait until we're switched over to `lucuma-odb`.

There is also #646, which also needs to be merged at the Time of the Great Transition. Or, maybe we just need to freeze Explore `staging` at this point in time. I should be done enough with the configuration stuff in a couple of days for it to be in staging. I'm not sure what the status of the target editing is.

toddburnside:
> Target editing is close to working. Another day or two.

My recommendation is we freeze library updates to the main branch of explore. We could then merge some of the other library PRs that are pending.

swalker2m:
> My recommendation is we freeze library updates to the main branch of explore. We could then merge some of the other library PRs that are pending.

Any objection then to merging this and #646 ?

cquiroz:
I'd say we merge and keep the main explore branch on the previous core version until the odb integration is complete

toddburnside:
> I'd say we merge and keep the main explore branch on the previous core version until the odb integration is complete

Agreed. Although I think the freeze should extend to most of the `lucuma*` libraries so I can start merging some other PRs.

So, merge away.

PR #652, opened 2023-01-26
Convenience methods for TimeSpan

rpiaggio:


modules/core/shared/src/main/scala/lucuma/core/util/TimeSpan.scala
```diff
@@ -34,6 +34,9 @@ object TimeSpan {
   val Max: TimeSpan =
     NonNegLong.MaxValue
 
+  val Zero: TimeSpan =
+    unsafeFromMicroseconds(0L)
```
swalker2m, line 38: Or just `= Min` I suppose.

swalker2m (approved):
:+1:

PR #656, opened 2023-01-31
Limit arbitrary atom and sequence size

swalker2m:
Currently arbitrary atoms contain an arbitrarily sized list of steps and arbitrary sequences contain an arbitrarily sized list of atoms.  The combination can get quite big, especially for the `FutureExecutionConfig` which contains two sequences.  Since this impacts test run time considerably the goal is to limit these sizes to something more manageable.

### DIFF
--- a/modules/testkit/src/main/scala/lucuma/core/model/sequence/arb/ArbAtom.scala
+++ b/modules/testkit/src/main/scala/lucuma/core/model/sequence/arb/ArbAtom.scala
@@ -16,22 +16,26 @@ trait ArbAtom {
   import ArbUid._
   import ArbStep._
 
-  implicit val arbAtomGmosNorth: Arbitrary[Atom.GmosNorth] = Arbitrary(
+  private def genBoundedAtom[A: Arbitrary, B](limit: Int, f: (Atom.Id, List[A]) => B): Gen[B] =
     for {
       id    <- arbitrary[Atom.Id]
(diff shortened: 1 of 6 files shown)


rpiaggio:
I think this can be achieved with the ScalaCheck parameter `maxSize`.

We are doing this for example in https://github.com/gemini-hlsw/lucuma-core/blob/master/modules/tests/shared/src/test/scala/lucuma/core/model/sequence/AtomSuite.scala#L20

swalker2m:
> Ah, but in this case you are limiting the size in the arbitrary generation wherever it is used, whereas the ScalaCheck param only limits the run.

I wasn't aware, or at least forgot about, the test parameters solution.  Thanks for pointing that out!

I'm fine with it, though you have to remember to set it in anything that ultimately includes an arbitrary sequence and the size limit applies to everything else in your suite which might sometimes be too broad?  It's kind of "arbitrary" but I don't think there's much testing to gain with very large atoms and sequences so it seemed ok to just fix the upper bound at a small value.

What do you think?  I can go either way.  If we keep this change then we should remove the max size overrides.

rpiaggio:
I'm all for limiting the arbitraries and removing the max size overrides. That way, we don't have to set them in every test that uses the arbs.

rpiaggio (approved):
LGTM

tpolecat (approved):
Agree with all of this, seems good.

PR #657, opened 2023-01-31
BoundedInterval

rpiaggio:
`spire.math.Interval` has no provision to work with bounded intervals, forcing the user to contemplate the empty and unbounded cases even when we know the interval to be bounded.

To simplify working with these cases, there was previously a set of implicits. However, they didn't contemplate the case where the interval is a single point (which in Spire is a distinct and unrelated case to a bounded interval).

This PR implements a type alias that groups the bounded and point cases and offers a set of extension methods to work with them ergonomically.


modules/core/shared/src/main/scala/lucuma/core/math/BoundedInterval.scala
```diff
@@ -0,0 +1,69 @@
...
+
+object BoundedInterval:
+  def closed[A: Order](lower: A, upper: A): Option[BoundedInterval[A]] = 
+    Interval.closed(lower, upper).some.filterNot(_.isEmpty).map(_.asInstanceOf[BoundedInterval[A]])
+
+  def unsafeClosed[A: Order](lower: A, upper: A): BoundedInterval[A] = 
+    closed(lower, upper).getOrElse(sys.error(s"Could not build a bounded closed interval with lower bound [$lower] and upper bound [$upper]."))
+
+  def openUpper[A: Order](lower: A, upper: A): Option[BoundedInterval[A]] = 
+    Interval.openUpper(lower, upper).some.filterNot(_.isEmpty).map(_.asInstanceOf[BoundedInterval[A]])
```
swalker2m, line 30: Are `closed` and `openUpper` the only cases we're interested in?

swalker2m (approved):
I've never worked with Spire's `Interval` but this makes sense to me.  :+1:

modules/core/shared/src/main/scala/lucuma/core/math/BoundedInterval.scala
rpiaggio: Those are the only ones we are using for the moment, but I just added the other ones for completeness.

rpiaggio:
I went ahead and switched to an opaque type.

Although we need stubs for all methods, this gives us more control over the specific types, returning `BoundedInterval` whenever possible.

swalker2m (approved):
This looks excellent.  Seems like a good idea.  :+1:

PR #661, opened 2023-02-02
NonEmptyList / Zipper syntax

swalker2m:
Adds a bit of syntax to create `Zipper`s from `NonEmptyList` where the focused element is a min or max according to `Order`.  I suppose this could alternatively just be added to the `Zipper` companion.


swalker2m:
Or maybe just add `focusMin` and `focusMax` to `Zipper` itself.

PR #662, opened 2023-02-02
Zipper focus ops

swalker2m:
Adds a few focus setting ops to `Zipper`, all of which obtain an equivalent `Zipper` but with an updated focus

* `focusFirst` focused on the first element
* `focusMax` focused on the first maximum element according to `Order`
* `focusMin` focused on the first minimum element according to `Order`


toddburnside (approved):
LGTM

PR #663, opened 2023-02-06
ValidFilter + value in errors

rpiaggio:
- Introduce a `ValidFilter` optic, composable with `ValidWedge` and `ValidSplitEpi`, to perform simple validations without transforming the value. Allows composing a `ValidSplitEpi` parser validator with, for example, checking that a parsed numeric value is within certain bounds.
- Add the input value as a parameter to the error instances. Allows error messages like `Value XXX is out of bounds`.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/units.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/units.scala
@@ -142,6 +142,8 @@ trait units {
 
   given ValueConversion[Parallax.LongParallaxμas, Rational] = Rational(_)
 
+  given refinedValueConversion[V, P]: ValueConversion[V Refined P, V] = _.value
+
   inline given [UF, UT]: TruncatingUnitConversion[PosInt, UF, UT] = v =>
     refineV[Positive](coulomb.conversion.standard.unit.ctx_TUC_Int(v))
       .getOrElse(1.refined[Positive])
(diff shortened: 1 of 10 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -142,6 +142,8 @@ trait units {
 
   given ValueConversion[Parallax.LongParallaxμas, Rational] = Rational(_)
 
+  given refinedValueConversion[V, P]: ValueConversion[V Refined P, V] = _.value
```
rpiaggio, line 145: This is unrelated. Allows easy "unrefining" of coulomb quantities.

modules/core/shared/src/main/scala/lucuma/core/optics/ValidFormat.scala
```diff
@@ -13,4 +13,10 @@ trait ValidFormat[E, A, B] {
   val getValid: A => Either[E, B]
 
   val reverseGet: B => A
+
+  /** Like getValid, but throws IllegalArgumentException when Invalid. */
+  def unsafeGet(a: A): B =
```
rpiaggio, line 18: Pulled upwards from subtypes.

toddburnside (approved):
LGTM

PR #664, opened 2023-02-07
WavelengthCoverage

rpiaggio:
Introduces a `WavelengthCoverage` model class, which represents a wavelength width. It can be centered around (or starting at, or ending at) a `Wavelength` to obtain a `BoundedInterval[Wavelength]` representing coverage.

Replaces previous `Coverage` class.

(Maybe we should name this `WavelengthWidth`??)


swalker2m (approved):
Looks great! :+1:

swalker2m:
> Maybe we should name this WavelengthWidth??

Coverage seems to be the term the astronomers use.  I don't know; it seems good to me.

rpiaggio:
I was wondering if we should distinguish between the terms for the range (ie: `[λ1, λ2]`) and the width of the range (ie: `λ2 - λ1`; `-` is a minus here). We seem to be using "coverage" interchangeably for both.

modules/core/shared/src/main/scala/lucuma/core/math/WavelengthCoverage.scala
```diff
@@ -0,0 +1,203 @@
...
+     * Alias for `toPicometers`.
+     */
+    def pm: Quantity[PosInt, Picometer] =
+      toPicometers
+
+    /**
+      * Create a wavelength range centered on the given value. Will be truncated at 
+      * Wavelength.Min and Wavelength.Max.
+      */
+    def centeredAt(λ: Wavelength): BoundedInterval[Wavelength] =
```
toddburnside, line 50: I assume planning to use this class to calculate `coverageInterval` in the SpectroscopyModesMatrix in explore? It seems there we have another case where the interval gets `clipped` within a larger interval. So given a central wavelength, the coverage interval can't extend beyond the capabilities of the configuration. Do you want to include that case here, or just handle it manually in SpectroscopyModesMatrix?

toddburnside (approved):
LGTM

toddburnside:
> I was wondering if we should distinguish between the terms for the range (ie: `[λ1, λ2]`) and the width of the range (ie: `λ2 - λ1`; `-` is a minus here). We seem to be using "coverage" interchangeably for both.

I see what you mean. Maybe `CoverageWidth` or `WavelengthCoverageWidth`?

swalker2m:
> I see what you mean. Maybe `CoverageWidth` or `WavelengthCoverageWidth`?

At least in one version of the inception document it seems we label this "λ Range" (in the configuration panel) and use "λ Coverage" for the interval itself (in advanced configuration) so maybe `WavelengthRange`?  If we're using a `BoundedInterval[Wavelength]` for the coverage itself then the "coverage" term goes away in the type names I guess.

rpiaggio:
Ok, renaming to `WavelengthRange` since it seems consistent with the inception document.

rpiaggio:
Ooops merged before pushing the rename commit. I'll send another PR.

modules/core/shared/src/main/scala/lucuma/core/math/WavelengthCoverage.scala
rpiaggio: > Do you want to include that case here, or just handle it manually in SpectroscopyModesMatrix?

It's done via `BoundedInterval.intersect` in `SpectroscopyModesMatrix`.

PR #665, opened 2023-02-07
Rename WavelengthCoverage to WavelengthRange

rpiaggio:


swalker2m (approved):
This seems fine with me, and as I mentioned seems to match the inception document.  So :+1:.  Perhaps we should change Explore's initial configuration panel to match?  It looks like we use "λ Coverage" inconsistently in the UI as I think you pointed out.

swalker2m:
> Perhaps we should change Explore's initial configuration panel to match?

I don't know the history of this.  If Andy is happy the way it is then so am I.

PR #666, opened 2023-02-08
Make Proposal.toOActivation optional

toddburnside:
`toOActivation` is (and has been) optional in the API, so it should be optional in the model as well. I just never noticed this before, because Explore always sends a value for it.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-ThisBuild / tlBaseVersion                         := "0.64"
+ThisBuild / tlBaseVersion                         := "0.65"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Proposal.scala
@@ -21,7 +21,7 @@ final case class Proposal(
   title:         Option[NonEmptyString],
   proposalClass: ProposalClass,
   category:      Option[TacCategory],
-  toOActivation: ToOActivation,
+  toOActivation: Option[ToOActivation],
   abstrakt:      Option[NonEmptyString],
   partnerSplits: SortedMap[Partner, IntPercent]
 )
@@ -35,7 +35,7 @@ object Proposal {
   val partnerSplits = Focus[Proposal](_.partnerSplits)
 
   val Default: Proposal =
-    Proposal(None, ProposalClass.Queue(80.refined), None, ToOActivation.None, None, SortedMap.empty)
+    Proposal(None, ProposalClass.Queue(80.refined), None, ToOActivation.None.some, None, SortedMap.empty)
 
   implicit val eqProposal: Eq[Proposal] = Eq.instance {
     case (Proposal(t1, pc1, c1, to1, a1, ps1), Proposal(t2, pc2, c2, to2, a2, ps2)) =>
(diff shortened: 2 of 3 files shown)


rpiaggio (approved):
LGTM

swalker2m:
The code looks good, but I wonder if we should instead change the API to make it required?  `ToOActivaton` has a `None` value already.

swalker2m:
The `TooInput` requires it to be specified upon creation, which makes sense to me.

toddburnside:
> The code looks good, but I wonder if we should instead change the API to make it required? `ToOActivaton` has a `None` value already.

In Explore, we send `None` on creation. I'd be fine with making in required in the API.

swalker2m:
Hmm, if you don't provide a value, you should get an error:

```
      case _ => Result.failure("All of proposalClass, toOActivation, and partnerSplits are required on creation."
```

toddburnside:
> Hmm, if you don't provide a value, you should get an error:

We always send a value, so I never saw that. However, it seems it was possible to create one without a value, because I was getting decoder errors when retrieving it. I'm not sure how it got to that state. I could try duplicating it in the playground.

toddburnside:
Related to this (and why I was having problems creating a proposal and maybe ended up with one without a toOActivation), the API used to allow either an empty `partnerSplits` array, OR the allocations had to add up to 100. Now it requires them to add up to 100. So, I changed out `toInput` method to `unassign` on empty list and that works. So, maybe the error check above isn't working?

But, thinking about it, maybe it should allow an empty list on creation. It returns an empty list if I don't provide one, so that seems a bit inconsistent.

swalker2m:
> So, it is not requiring proposalClass, toOActivation, or partner splits.

It isn't required according to the schema itself (because the same input is used for editing the proposal), but we check and validate the inputs upon receipt.  In the case of creation, if no TOO value is specified, it should generate an error.  If I update it, then the only thing that would change would be the `Proposal` type itself where it appears optional I think by mistake.

swalker2m:
See:

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/service/src/main/scala/lucuma/odb/graphql/input/ProposalInput.scala#L73-L74

toddburnside:
You beat me to it. I was just typing to say nothing is required in the ProposalInput because it is being used for both creating and editing. 

Despite the code you pointed to, playing in the playground I can create a proposal with only a `proposalClass`

swalker2m:
> Despite the code you pointed to, playing in the playground I can create a proposal with only a proposalClass

Interesting, there must be a bug somewhere.

Thinking back to the existing Phase 1 Tool, it may be set up as an `Option` with a similar enum.  The idea being that they wanted to distinguish between `None` as in the user hasn't actively chosen anything yet and `Some(Too.None)` as in the user explicitly chose no ToO.  I don't know if we need to distinguish that still and allow proposal creation without having yet selected the ToO option.  If we do then we'd need some kind of validation step before it can be submitted for time allocation -- maybe that is planned anyway?

toddburnside:
> If we do then we'd need some kind of validation step before it can be submitted for time allocation -- maybe that is planned anyway?

I think we'll probably need some sort of validation. In Explore we default to:
- proposalClass => Queue with minPercentTime of 80
- toOActivation => None
But, it is difficult to provide a reasonable default for the partner splits. At some point we discussed with Andy the possibility of trying to come up with a default based on some information about the user, but he decided it would be wrong way too often.

I have no problem with making the `proposalClass` and `toOActivation` required for creation, though.

swalker2m:
Okay, I think I'm inclined to make TOO `!` required in the `Proposal` schema type and see whatever is going wrong with the check on creation.

toddburnside:
> Okay, I think I'm inclined to make TOO `!` required in the `Proposal` schema type and see whatever is going wrong with the check on creation.

I can create an issue for that.

What about the partnerSplits? Should the API accept an empty array? Or do I exclude it initially and have the API return an empty array? I think the former makes most sense, but I can handle it either way.

swalker2m:
> What about the partnerSplits? Should the API accept an empty array? Or do I exclude it initially and have the API return an empty array? I think the former makes most sense, but I can handle it either way.

I'm not sure.  There's no good default for it other than an empty array so It seems like we should accept that.  Otherwise you have to specify the partner request(s) up front.

toddburnside:
Created issue https://github.com/gemini-hlsw/lucuma-odb/issues/253. Further discussion can take place there.

PR #669, opened 2023-02-09
Limited precision values for BrightnessValue

rpiaggio:
Make `LimitedBigDecimals` available in testkit too.


toddburnside (approved):
LGTM

PR #678, opened 2023-02-17
Fix BoundedInterval constructor, tests, duration for Instant

rpiaggio:
The constructor `BoundedInterval.fromInterval` had a bug when passing an `Above`, `Below` or `All`. It mistakenly constructed a `BoundedInterval` when it shouldn't.

The existence of `BoundedInterval` can now simplify code in a number of places.

Extension method `duration` for `IntervalSeq[Instant]` had a bug, where it called itself indefinitely.

Most tests that were in `IntervalSuite` are now in `BoundedIntervalSuite`.


toddburnside (approved):
LGTM

PR #679, opened 2023-02-20
Object Tracking

cquiroz:
This is an class I use in explore but that I'd like to use in other places (lucuma-catalog)
The main purpose is to generalize tracking an object so we can use the same logic for sidereal/non-sidereal/asterism


rpiaggio (approved):
LGTM

PR #681, opened 2023-02-22
Add average parallactic calculation

cquiroz:
This was ported from the OCS and included directly on explore where it was validated.
This is going to be useful in other places thus it makes sense no move it to core


toddburnside (approved):
LGTM

PR #684, opened 2023-03-01
objectUrl in CatalogInfo

rpiaggio:
I'm not sure if it's of interest to make this generally available in core.

Otherwise, we can make it an extension method in `lucuma-ui` if it's going to be shared by multiple UIs, or just in `explore`, where we actually need it for the time being.


toddburnside (approved):
I don't have a problem with it in core. Maybe others do.

PR #699, opened 2023-04-05
Avoid errors and warnings

rpiaggio:
We're throwing an NPE when attempting to access `ElevationRange.HourAngle.DecimalHour` on Scala 3.2.2, so hoping that changing from an `object` to `lazy val` will fix that.

Also, we get a warning for `inline given` defined as a function in newer versions, so removing the `inline` there.


tpolecat:
Looks like you need to bump the base version.

rpiaggio:
> Looks like you need to bump the base version.

Done

toddburnside (approved):
LGTM

PR #700, opened 2023-04-06
SignalToNoise

swalker2m:
Adds a `SignalToNoise` value that can be precisely mapped to the database.  In other words, if you can make a `SignalToNoise` it can be stored and then an exact `Eq` value can be retrieved.

After updating the schema (in a separate PR), this will address an issue in which we allow any `PosBigDecimal` to be passed in as a signal-to-noise value but can only actually store a subset of them.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
@@ -1,0 +1,128 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math
+
+import cats.Order
+import cats.Show
+import eu.timepit.refined.types.numeric.PosBigDecimal
+import io.circe.Decoder
+import io.circe.Encoder
+import io.circe.JsonNumber
+import lucuma.core.optics.Format
+import monocle.Prism
+
+import scala.util.control.Exception.catching
+
+// Signal-to-noise stored as milli-signal to noise Long capped at 9,999,999,999
+// so that it fits in a database numeric(10, 3).
+
+opaque type SignalToNoise = Long
+
+object SignalToNoise {
+
+  /**
+   * Maximum supported signal-to-noise value: 9,999,999.999.
+   */
+  val Max: SignalToNoise =
+    9_999_999_999L
+
+  /**
+   * Minimum supported signal-to-noise value: 0.001.
+   */
+  val Min: SignalToNoise =
+    1L
+
+  extension (s2n: SignalToNoise) {
+
+    /**
+     * Converts this SignalToNoise value to a `BigDecimal`, which is guaranteed
+     * to be positive.
+     */
+    def toBigDecimal: BigDecimal =
+      BigDecimal(s2n, 3)
+
+    /**
+     * Converts this SignalToNoise value to a `PosBigDecimal`.
+     */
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
```diff
@@ -0,0 +1,128 @@
...
+import io.circe.Decoder
+import io.circe.Encoder
+import io.circe.JsonNumber
+import lucuma.core.optics.Format
+import monocle.Prism
+
+import scala.util.control.Exception.catching
+
+// Signal-to-noise stored as milli-signal to noise Long capped at 9,999,999,999
+// so that it fits in a database numeric(10, 3).
```
swalker2m, line 18: Why `numeric(10, 3)`?  It seemed "big enough" as far as I can tell without being ridiculously too large.  We can always increase it later if necessary.

modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
```diff
@@ -0,0 +1,128 @@
...
+
+opaque type SignalToNoise = Long
```
rpiaggio, line 20: May I suggest using something like:

```scala
protected type SNRefinedType = Long Refined Interval.Open[0L, 10000000000L]
object SignalToNoise extends RefinedTypeOps[SNRefinedType, Long] with NewType[SNRefinedType] 
type SignalToNoise = SignalToNoise.Type
```

That way the underlying type is refined and we get the consturctors for free (and consistent with refined constructors).

rpiaggio: Disregard that, the refined constructors will return a `SNRefinedType`. There doesn't seem to be a way to do what I suggest since it results in a recursive definition between the object and the opaque type.

rpiaggio (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
rpiaggio: Oh wait, this seems to work, without using `NewType`:

```scala
opaque type SignalToNoise = Long Refined Interval.Open[0L, 1000000000L]
object SignalToNoise extends RefinedTypeOps[SignalToNoise, Long] with NewType[SignalToNoise]
```
```scala
scala> SignalToNoise.from(1L)
val res2: Either[String, Test.SignalToNoise] = Right(1)
                                                                                                                                                                                               
scala> SignalToNoise.from(0L)
val res3: Either[String, Test.SignalToNoise] = Left(Left predicate of ((0 > 0) && (0 < 1000000000)) failed: Predicate failed: (0 > 0).)
```

cquiroz (approved):
LGTM

cquiroz:
We should probably add this to the ITC graphql schema

modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
swalker2m: Thanks, that's a pretty neat trick!  I was trying to keep the underlying representation as milli-sn somewhat hidden though because it didn't seem completely clear that it won't need to change one way or another.  For example if it were to change such that `0.01` is the smallest SN we can represent, client code like `SignalToNoise.from(1L)` would suddenly mean something different.

rpiaggio: Understood :+1:. I like the idea of having explicit constructor names.

PR #703, opened 2023-04-12
Add PosBigDecimal optics to SignalToNoise

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
```diff
@@ -112,6 +132,10 @@ object SignalToNoise {
   given Order[SignalToNoise] =
     Order.fromLessThan(_ < _)
 
+  given AdditiveCommutativeSemigroup[SignalToNoise] = new AdditiveCommutativeSemigroup[SignalToNoise] {
```
rpiaggio, line 135: Nice

modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
```diff
@@ -112,6 +132,10 @@ object SignalToNoise {
...
+    def plus(x: SignalToNoise, y: SignalToNoise): SignalToNoise = (x + y)
```
swalker2m, line 136: Because I chose to make the underlying type simple (a `long`) this compiles but produces an illegal value sometimes?   What is `Max` + `Max` for example?

Maybe I should have used the refined type range for the representation.

modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
```diff
@@ -80,6 +92,14 @@ object SignalToNoise {
   val FromBigDecimalRounding: Format[BigDecimal, SignalToNoise] =
     Format(bd => fromMilliDecimal((bd * 1000).setScale(0, BigDecimal.RoundingMode.HALF_UP)), _.toBigDecimal)
 
+  /**
+   * Creates a `SignalToNoise` value assuming that the given PosBigDecimal is in
+   * range [Min, Max].  Rounds finer scale values to milli-sn.
+   *
+   * @group Optics
+   */
+  val FromPosBigDecimalRounding: Format[PosBigDecimal, SignalToNoise] =
```
rpiaggio, line 101: May I suggest using a `ValidSplitEpi` here? It combines the functionality of `Format` and `SplitEpi`: the valid inputs can be normalized.

Probably `FromBigDecimalRounding` above too.

I suspect that going this way we won't need custom generators in the tests.

modules/core/shared/src/main/scala/lucuma/core/math/SignalToNoise.scala
swalker2m: I wasn't aware that we needed to add them, but if we do really add would be `(SignalToNoise, SignalToNoise) => Option[SignalToNoise]`?

cquiroz: because of the range limits?

swalker2m: Right, if we can end up with a `SignalToNoise` value that is bigger than `Max` the database would reject it when it processes inputs.  A `reverseGet` result passed to `getOption` on one of the optics would produce `None`, etc.

cquiroz: Ok, let's rather remove the instance, we can convert to numbers and do operations that way

modules/tests/shared/src/test/scala/lucuma/core/math/SignalToNoiseSuite.scala
```diff
@@ -18,12 +22,18 @@ final class SignalToNoiseSuite extends munit.DisciplineSuite {
...
+  checkAll("Order",                     OrderTests[SignalToNoise].order)
+  checkAll("JSON Codec",                CodecTests[SignalToNoise].codec)
+  checkAll("FromBigDecimalExact",       PrismTests(SignalToNoise.FromBigDecimalExact))
+  checkAll("FromBigDecimalRounding",
+    ValidSplitEpiTests(SignalToNoise.FromBigDecimalRounding).validSplitEpiWith(bigDecimalSignalToNoise))
+  checkAll("FromPosBigDecimalExact",    PrismTests(SignalToNoise.FromPosBigDecimalExact))
+  checkAll("FromPosBigDecimalRounding",
+    ValidSplitEpiTests(SignalToNoise.FromPosBigDecimalRounding).validSplitEpiWith(posBigDecimalSignalToNoise)
+  )
+  checkAll("FromString",                FormatTests(SignalToNoise.FromString).formatWith(stringSignalToNoise))
```
rpiaggio, line 37: Do the tests fail with regular `BigDecimal` and `PosBigDecimal` arbitraries?

swalker2m: I've never used `ValidSplitEpi` and am not sure what it is, but if it is like `Format` I think we're good?  The tests each include:

* valid big decimals: in range and not overly precise
* "coverage" examples: big decimals that are in range but way too precise -- these can be rounded to be valid.  After the initial round trip the canonical format emerges.
* arbitrary big decimals which may or may not be in range

```scala
  private val validBigDecimalSignalToNoise: Gen[BigDecimal] =
    Gen
       .choose(1L, 9_999_999_999L)
       .map(BigDecimal(_, 3))

  private val coverageExamples: Gen[BigDecimal] =
     Gen
       .choose(1L, 9_999_999_999_999L)
       .map(BigDecimal(_, 6))

  val bigDecimalSignalToNoise: Gen[BigDecimal] =
     Gen.oneOf(validBigDecimalSignalToNoise, coverageExamples, arbitrary[BigDecimal])
```

Is that what you meant?

rpiaggio: My bad, I missed the fact that `bigDecimalSignalToNoise` includes `arbitrary[BigDecimal]` (and similarly `posBigDecimalSignalToNoise`), and so I thought we were leaving cases out. Sorry for that!

rpiaggio (approved):
LGTM

swalker2m (approved):
Thanks for adding those, looks great.  :+1:

PR #704, opened 2023-04-14
Rename Attachment.Id to ObsAttachment.Id

toddburnside:


rpiaggio (approved):
LGTM

Will there be other kinds of attachments?

toddburnside:
> Will there be other kinds of attachments?

Yes. There will be proposal attachments. But, at this point I'm thinking they won't get their own ids.

PR #705, opened 2023-04-14
Update JTS and test a bug present on the previous JTS port

cquiroz:
When doing intersection of offsets a bug was discovered that sent us down a rabbit hole as it needed a full update of the geometry library which fixes the bug

This PR incorporate that lib and adds a test case to verify the bug is fixed

I may not merge this right away to test with explore first


modules/core/shared/src/main/scala/lucuma/core/geom/jts/JtsShapeInterpreter.scala
```diff
@@ -20,7 +20,7 @@ import java.util.Collections
 object JtsShapeInterpreter extends ShapeInterpreter {
 
   private val EmptyGeometry: Geometry =
-    Jts.geometryFactory.buildGeometry(Collections.emptyList[Geometry])
+    Jts.geometryFactory.createPoint
```
cquiroz, line 23: This change is needed or many of the tests break

swalker2m: I don't really understand what an empty point means but I suppose if it works ...

build.sbt
```diff
@@ -14,7 +14,7 @@ lazy val kindProjectorVersion  = "0.13.2"
 lazy val monocleVersion        = "3.2.0"
 lazy val scalaJavaTimeVersion  = "2.5.0"
 lazy val geminiLocalesVersion  = "0.7.0"
-lazy val jtsVersion            = "0.3.0"
+lazy val jtsVersion            = "0.4-6289e97-SNAPSHOT"
```
swalker2m, line 17: Did you need further changes after you updated the version?

modules/core/shared/src/main/scala/lucuma/core/geom/jts/JtsShapeInterpreter.scala
cquiroz: They are in fact different entites a point is just a point:
```
println(Jts.geometryFactory.createPoint)
   POINT EMPTY```

While the previous one is an empty geometry collection
```
println(Jts.geometryFactory.buildGeometry(Collections.emptyList[Geometry]))
   GEOMETRYCOLLECTION EMPTY
```

I'm not sure the reason why but the updated JTS lib cannot do operations like union or intersection on GEOMETRYCOLLECTION

PR #707, opened 2023-04-19
TimeSpan math ops

swalker2m:
While it's true that you cannot arbitrarily add two `TimeSpan`s and expect to get a valid `TimeSpan` result, it's still useful to be able to add, subtract, multiply and divide them without worrying about it in most situations.  A `TimeSpan.Max` is essentially infinity for our purposes, and treating it as the bounds of math operations makes the 99% (or better) case easier. For example, if exposure time takes `TimeSpan.Max` then the fact that detector readout and write to permanent storage require some additional amount of time is not very useful anyway. 

The pipe,`|`,  in `t0 +| t1` etc. is meant to be suggestive of a hitting a wall or bounds.  `TimeSpan` forms a `Monoid` under `+|` which is also useful.

I don't know.  Is all of this dumb?

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/TimeSpan.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/TimeSpan.scala
@@ -3,11 +3,14 @@
 
 package lucuma.core.util
 
+import cats.Monoid
 import cats.Order
 import cats.Order.catsKernelOrderingForOrder
 import cats.syntax.option.*
 import cats.syntax.order.*
+import eu.timepit.refined.types.numeric.NonNegInt
 import eu.timepit.refined.types.numeric.NonNegLong
+import eu.timepit.refined.types.numeric.PosInt
 import lucuma.core.optics.Format
 import monocle.Iso
 import monocle.Prism
@@ -15,6 +18,7 @@ import monocle.Prism
 import java.math.RoundingMode.HALF_UP
 import java.time.Duration
 import java.time.temporal.ChronoUnit.MICROS
+import scala.annotation.targetName
 import scala.util.Try
 
 /**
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/util/TimeSpan.scala
```diff
@@ -141,6 +144,39 @@ object TimeSpan {
...
+    @targetName("boundedSubtract")
+    def -|(other: TimeSpan): TimeSpan =
+      fromMicroseconds(timeSpan.toMicroseconds - other.toMicroseconds).getOrElse(Min)
+
+    /**
+     * Multiplies a TimeSpan by an integer, limiting the resulting value to the
+     * range (`Min`, `Max`).
+     */
+    @targetName("boundedMultiply")
+    def *|(multiplier: Int): TimeSpan = {
```
rpiaggio, line 166: Is there a reason for division to require a `PosInt` but multiplication to allow negative numbers?

Also, can we support decimals (with rounding on the result) ?

swalker2m: Actually I just want a `NotZeroInt` but I don't think that exists?  Dividing by a negative is never going to produce anything but `Zero` so I went with `PosInt`.

Decimal options with rounding seem fine.  I just didn't need them for what I'm working on.

rpiaggio: Won't multiplying by a negative also always produce `Zero` ?

swalker2m: I imagine with `refined` you can scare up a `Not[Equal[_0]]` or some such but it seems awkward, I don't know.

swalker2m: > Won't multiplying by a negative also always produce Zero ?

This is true.  It could be a `NonNegInt` but it's easier to work with a raw `Int` and there's a valid way to use the full range of `Int` as the multiplier so I switched to that.  I guess the inconsistency is a little irksome though.

rpiaggio (approved):
LGTM

swalker2m:
@rpiaggio I added a `NonZeroInt` to make multiplication and division more consistent (in that both would allow negatives).  If this is better (?) does `NonZeroInt` already exist somewhere to anybody's knowledge?  I made a `lucuma.core.refined` package for it, but maybe there's a better existing place?

I wonder if `TimeSpan` and `Timestamp` would be better off in `lucuma.core.data` for that matter.

PR #708, opened 2023-04-21
Setup published npm package

armanbilge:


build.sbt
```diff
@@ -126,3 +126,46 @@ lazy val benchmarks = project
...
+  .settings(
+    scalaJSLinkerConfig ~= (_.withModuleKind(ModuleKind.ESModule)),
+    npmPublish := {
+      import scala.sys.process._
+      val _      = (Compile / fullLinkJS).value
+      val outDir = (Compile / fullLinkJS / scalaJSLinkerOutputDirectory).value
+      IO.write(
+        outDir / "package.json",
+        s"""|{
+            |  "name": "lucuma-core",
```
armanbilge, line 145: This will publish here:
https://www.npmjs.com/package/lucuma-core

build.sbt
```diff
@@ -126,3 +126,46 @@ lazy val benchmarks = project
   .dependsOn(core.jvm)
   .settings(name := "lucuma-core-benchmarks")
   .enablePlugins(NoPublishPlugin, JmhPlugin)
+
+// for publishing CSS to npm
```
cquiroz, line 130: I guess publishing scala.js to npm?

armanbilge: Woops, copy-pasted this from lucuma-css :)

modules/npm/src/main/scala/lucuma/core/Foo.scala
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core
+
+import scala.scalajs.js
+import scala.scalajs.js.annotation.JSExportTopLevel
+
+@JSExportTopLevel("Foo")
+class Foo extends js.Object
```
armanbilge, line 10: So I think this will work as:
```js
import { Foo } from 'lucuma-core';
```

If we write `@JSExportTopLevel("Foo", "somewhere")` then it becomes:
```js
import { Foo } from 'lucuma-core/somewhere.js';
```

but not 100% on this :)

build.sbt
cquiroz: We need credentials right?

modules/npm/src/main/scala/lucuma/core/Foo.scala
armanbilge: Note, you can't nest any further than that. See:
- https://github.com/scala-js/scala-js/issues/4742#issuecomment-1518126573

cquiroz: Seems right, I'll experiment a bit

build.sbt
armanbilge: Already set up.

cquiroz:
Thanks a lot this seems great

cquiroz (approved):
LGTM

PR #711, opened 2023-04-25
Allow SignalToNoise to be zero

cquiroz:
Turns out there are edge cases where SignalToNoise can be 0
You can have an all noise picture for example and the ITC returns 0 in some edge cases (extremely faint stars, or odd component combinations)

### DIFF
--- a/modules/tests/shared/src/test/scala/lucuma/core/math/SignalToNoiseSuite.scala
+++ b/modules/tests/shared/src/test/scala/lucuma/core/math/SignalToNoiseSuite.scala
@@ -39,8 +39,8 @@ final class SignalToNoiseSuite extends munit.DisciplineSuite {
   checkAll("FromBigDecimalExact",       PrismTests(SignalToNoise.FromBigDecimalExact))
   checkAll("FromBigDecimalRounding",
     ValidSplitEpiTests(SignalToNoise.FromBigDecimalRounding).validSplitEpiWith(bigDecimalSignalToNoise))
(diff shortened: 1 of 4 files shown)


modules/tests/shared/src/test/scala/lucuma/core/math/SignalToNoiseSuite.scala
```diff
@@ -15,24 +18,30 @@ import lucuma.core.optics.Format
...
+    Gen.posNum[BigDecimal].map(refineV[NonNegative](_).getOrElse(throw new RuntimeException))
+  }
+
   checkAll("Order",                     OrderTests[SignalToNoise].order)
   checkAll("JSON Codec",                CodecTests[SignalToNoise].codec)
   checkAll("FromBigDecimalExact",       PrismTests(SignalToNoise.FromBigDecimalExact))
   checkAll("FromBigDecimalRounding",
     ValidSplitEpiTests(SignalToNoise.FromBigDecimalRounding).validSplitEpiWith(bigDecimalSignalToNoise))
-  checkAll("FromPosBigDecimalExact",    PrismTests(SignalToNoise.FromPosBigDecimalExact))
+  checkAll("FromPosBigDecimalExact",    PrismTests(SignalToNoise.FromNonNegBigDecimalExact))
```
swalker2m, line 42: I suppose we could update the label as well since it still reads "FromPosBigDecimal".

swalker2m (approved):
This seems good to me.  I can't remember now why I didn't include `0`.  We'll need to change a database constraint when we update the ODB for this.

PR #712, opened 2023-04-25
Allow Typescript sources to be copied into npm package

armanbilge:


cquiroz (approved):
Uh great, Will this work for both npmPublish and npmPackage?

armanbilge:
Yes. `npmPackage` copies the sources. `npmPublish` runs `npmPackage` and then publishes.

modules/npm/src/main/typescript/main.d.ts
```diff
@@ -0,0 +1,3 @@
+declare class Foo {
```
armanbilge, line 1: Oh, this is missing an `export`. Oh well :)
```suggestion
export declare class Foo {
```

PR #714, opened 2023-04-26
Timing window model

rpiaggio:


modules/core/shared/src/main/scala/lucuma/core/enums/TimingWindowInclusion.scala
```diff
@@ -0,0 +1,26 @@
...
+  case Include, Exclude
+
+  def toBooleanInclude: Boolean = this == Include
+
+  def tag: String = this match
+    case Include => "include"
+    case Exclude => "exclude"
+
+object TimingWindowInclusion:
+  def fromBooleanInclude(include: Boolean): TimingWindowInclusion =
```
cquiroz, line 20: I guess you could also have an `Iso[TimingWindowInclusion, Boolean]`

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/TimingWindowInclusion.scala
rpiaggio: Good idea

PR #718, opened 2023-05-05
Experimentally export a method to get the night plot data in js/ts

cquiroz:


modules/npm/src/main/typescript/main.d.ts
```diff
@@ -1,3 +1,8 @@
-declare class Foo {
-  constructor();
+declare class PlotPoint {
+  constructor(instant: Date, airmass: Number, altitude: Number);
+  airmass: Number;
+  altitude: Number;
+  instant: Date;
 }
+
+declare function nightPlot(site: Site, start: Date, coords: String): PlotPoint[];
```
armanbilge, line 8: It should be `number` instead of `Number` and `string` instead of `String`.

https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types

cquiroz: Uh thanks for noticing that

armanbilge: Also I wonder if it should be `export declare class` and `export declare function`? I'm not sure 🤔

cquiroz: It is working as:
```typescript
import { GS } from 'lucuma-core'
import { PlotPoint } from 'lucuma-core'
import { nightPlot } from 'lucuma-core'

const points: Array<PlotPoint> = nightPlot(GS, new Date(), "09 55 33.173 +69 03 55.06")

console.log(points.length)
```

cquiroz: Also note that I have a `Site` param which isn't defined!

armanbilge: Does that code pass the typescript compiler?

armanbilge (approved):

modules/npm/src/main/typescript/main.d.ts
```diff
@@ -1,8 +1,12 @@
-declare class PlotPoint {
+export declare class PlotPoint {
   constructor(instant: Date, airmass: number, altitude: number);
```
armanbilge: If `PlotPoint` is not a `js.Object` then I don't think it has a JS-accessible ctor?

PR #719, opened 2023-05-06
Add TimingWindow.toIntervalSeq method

rpiaggio:
This turns a timing window definition into something we can work this (`IntervalSeq[Instant]`).


toddburnside (approved):
LGTM

PR #724, opened 2023-05-16
Angle and Offset updates

swalker2m:
A couple of math updates:

* `Angle` gets unsigned milliarcsecond and arcsecond optics
* `Offset` gets a `distance` to compute angular separation between two offsets

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Offset.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Offset.scala
@@ -51,6 +51,16 @@ final case class Offset(p: Offset.Component[Axis.P], q: Offset.Component[Axis.Q]
   def rotate(θ: Angle): Offset =
     Offset.rotateBy(θ)(this)
 
+  /**
+   * Angular separation between two offset positions.
+   */
+  def distance(o: Offset): Angle = {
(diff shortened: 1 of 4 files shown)


rpiaggio (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/Offset.scala
```diff
@@ -51,6 +51,16 @@ final case class Offset(p: Offset.Component[Axis.P], q: Offset.Component[Axis.Q]
   def rotate(θ: Angle): Offset =
     Offset.rotateBy(θ)(this)
 
+  /**
+   * Angular separation between two offset positions.
+   */
+  def distance(o: Offset): Angle = {
```
cquiroz, line 57: I guess this is for relatively small offsets?

swalker2m: In practice yes, but it's angular separation not RA/Dec coordinates so the declination is not relevant.  It should work regardless of the separation.

cquiroz: ah ok cool!

PR #725, opened 2023-05-16
Define amp count options

swalker2m:
A small update to the GMOS detector options.  It turns out that not just any combination of amp counts and detectors is valid.  E2V offers 3 and 6 while Hamamatsu 6 and 12.  (This fact will be needed later when GMOS readout time tables are validated.)


rpiaggio (approved):
LGTM

PR #727, opened 2023-05-16
Planned time model classes

swalker2m:
This is a collection of model classes for keeping up with time estimates during the planned time calculation. At this point, none of these objects are used in the codebase but they will be referenced in the sequence model in the next PR (and ultimately computed in the database).

__`StepEstimate`__ Contains detailed time estimation information for a single step.  This will replace the simplistic `StepTime`.  In particular, a step estimate contains a collection of `ConfigChangeEstimate` and a collection of `DetectorEstimate`.  In each case, the maximum time element in the collection is taken as the overall "config change" or "detector" time.  This is because configuration changes and detectors are read in parallel so there's no need to sum them. To hold the collection along with the selection, a `Zipper` is employed.

As an example, if step requires an offset that takes 7 seconds, and also moves the filter wheel requiring 20 seconds, you only need to figure that the configuration change as a whole is 20 seconds.

* __`ConfigChangeEstimate`__ An individual configuration change.  For example, to hold the cost of moving the science fold in order to perform a calibration.

* __`DetectorEstimate`__ A detector can produce a number of datasets per step, each of which is individually estimated in a `DatasetEstimate`.  An instrument can have multiple detectors, so the step has a collection of `DetectorEstimate`.  In the case of GMOS, this will always be one and it will always produce a single dataset per step.

* __`DatasetEstimate`__ The exposure, read and write times for a single dataset.

__`PlannedTime`__ Planned time breaks a time amount across `ChargeClass`es.  Each step/observe is associated with an `ObserveClass` which in turn has a corresponding `ChargeClass`.  The planned time for a single step, then, is all associated with a single `ChargeClass`.  `PlannedTime` can be summed though, such that the planned time for an `Atom` is the sum of the planned times for its steps and each step may have a different `ObserveClass`.  Similarly the planned time for a `Sequence` is the sum of the planned time for its atoms.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/DetectorEstimate.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/DetectorEstimate.scala
@@ -1,0 +1,56 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.model.sequence
+
+import cats.Order
+import cats.data.NonEmptyList
+import eu.timepit.refined.types.numeric.NonNegInt
+import lucuma.core.data.Zipper
+import lucuma.core.util.TimeSpan
+
+/**
+ * Time required for an individual detector to produce its datasets for a given
+ * step.  Many instruments (like GMOS-N and GMOS-S) have a single detector that
+ * produces a single dataset per step.  Newer instruments (like GHOST's blue
+ * red detectors) have multiple detectors and each may produce multiple
+ * datasets.
+ */
+case class DetectorEstimate(
+  name:        String,
+  description: String,
+  dataset:     DatasetEstimate,
+  count:       NonNegInt
+) {
+
+  /**
+   * Time for this detector to produce `count` datasets.
+   */
+  lazy val estimate: TimeSpan =
+    dataset.estimate *| count.value
+
+}
+
+object DetectorEstimate {
+
+  def zero(name: String, description: String): DetectorEstimate =
+    DetectorEstimate(name, description, DatasetEstimate.Zero, NonNegInt.unsafeFrom(0))
+
+  given Order[DetectorEstimate] =
+    Order.by { e => (
+      e.dataset,
+      e.count.value,
+      e.name,
+      e.description
+    )}
+
+  /**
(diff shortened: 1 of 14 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/DetectorEstimate.scala
```diff
@@ -0,0 +1,56 @@
...
+  description: String,
+  dataset:     DatasetEstimate,
+  count:       NonNegInt
+) {
+
+  /**
+   * Time for this detector to produce `count` datasets.
+   */
+  lazy val estimate: TimeSpan =
+    dataset.estimate *| count.value
```
cquiroz, line 30: I wonder if we should have a `*|` with refined ints

modules/core/shared/src/main/scala/lucuma/core/model/sequence/PlannedTime.scala
```diff
@@ -0,0 +1,116 @@
...
+
+    /**
+     * Gets the time charged for the given charge class (or TimeSpan.Zero if no
+     * charge is recorded).  Alias for `getOrZero`.
+     */
+    def apply(chargeClass: ChargeClass): TimeSpan =
+      getOrZero(chargeClass)
+
+    // N.B., if you refer to `apply` within this extensions object, it will
+    // find the Map apply :-/ For that reason I added `getOrZero`.
```
cquiroz, line 53: Is this because it is an opaque type?

modules/testkit/src/main/scala/lucuma/core/model/sequence/arb/ArbDetectorEstimate.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package lucuma.core.model.sequence
+package arb
+
+import eu.timepit.refined.types.numeric.NonNegInt
+import org.scalacheck.Arbitrary
+import org.scalacheck.Arbitrary.*
+import org.scalacheck.Cogen
+import org.scalacheck.Gen
+
```
cquiroz, line 12: extra line

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/sequence/PlannedTime.scala
swalker2m: Right, I would expect it to use the apply in the same extensions object but it doesn't seem to.

PR #730, opened 2023-05-24
Planned Time and sequence model update

swalker2m:
This PR offers three updates to the sequence model:

1. Moves the GMOS related classes to a `gmos` package.
2. Parameterizes the sequence on the instrument static, `S`, and dynamic, `D`, types.  All the instruments should have the same config / sequence / atom / step hierarchy but will differ on instrument details.  I plan to offer codecs for all these types in a `lucuma-odb-schema` library.
3. Re-imagines the sequence as a sequence _summary_ instead of a fully expanded list of atoms.  There is still the `nextAtom` and `possibleFuture`, but the future may not be complete.  It includes `hasMore` and `atomCount` to provide an indication of completion or not to the client.  Augmenting the potentially incomplete future, there is a `SequenceDigest` that contains calculations (like the planned time) that are summed across the entire future.  Eventually a list of at least the time for each future atom will probably be required by the scheduler.

I've removed `ManualConfig` for now since it is not yet supported and I hope we can just use normal sequence classes for it.

More details:

`InstrumentExecutionConfig` -> `ExecutionConfig[S, D]` -> `ExecutionSequence[D]` -> `Atom[D]` -> `Step[D]`

`S` Instrument static (unchangeable) config.
`D` Instrument dynamic (may change from step to step) config.

  * At the top level `FutureExecutionConfig` was renamed to `ExecutionConfig` and parameterized on the static, `S`, and dynamic `D` type of the instrument.
  * Parameterizes the `ExecutionSequence`, `Atom` and `Step` on a type, `D`, for the dynamic config.  For example, the GMOS North or GMOS South dynamic config.  Sequence generation is generic on the static and dynamic types so constructing these items without corresponding parameterization was cumbersome and resulted in needing an instance per instrument for each of sequence, atom, and step.
  * To get unparameterized `GmosNorth` and `GmosSouth` execution configs, there is now an `InstrumentExecutionConfig`.
  * Adds an optional `description` to an `Atom` to make it easier to pick out the significant change or feature behind the atom.  For example, for GMOS Longslit science sequences this can be the offset in `q` and the wavelength for the atom (e.g., "q 15.0″, λ 600 nm").  Whether it is useful for Explore to show this information is open for debate.  I think so, but a main motivator was to allow me to see what is happening in the query output.
  * The whole hierarchy, `ExecutionConfig` -> `ExecutionSequence` -> `Atom` -> `Step` has been given an `ObserveClass`  entry which summarize the corresponding elements of lower levels.  For example, the sequence observe class is the combination of its atom's observe classes which in turn sum their step observe classes.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
@@ -7,6 +7,7 @@ import cats.Eq
 import cats.data.NonEmptyList
 import cats.syntax.all._
 import eu.timepit.refined.auto._
+import eu.timepit.refined.types.string.NonEmptyString
 import lucuma.core.data.Zipper
 import lucuma.core.enums.ObserveClass
 import lucuma.core.util.TimeSpan
@@ -25,7 +26,7 @@ import monocle.macros.GenPrism
  */
 case class Atom[D](
   id:          Atom.Id,
-  description: Option[String],
+  description: Option[NonEmptyString],
   steps:       NonEmptyList[Step[D]]
 ) {
 
@@ -44,7 +45,7 @@ object Atom extends WithUid('a'.refined) {
     Focus[Atom[D]](_.id)
 
   /** @group Optics */
-  def description[D]: Lens[Atom[D], Option[String]] =
+  def description[D]: Lens[Atom[D], Option[NonEmptyString]] =
     Focus[Atom[D]](_.description)
 
   /** @group Optics */
@@ -54,7 +55,7 @@ object Atom extends WithUid('a'.refined) {
   given [D](using Eq[D]): Eq[Atom[D]] =
     Eq.by { a => (
       a.id,
-      a.description,
+      a.description.map(_.value),
       a.steps
     )}
 }
--- a/modules/testkit/src/main/scala/lucuma/core/model/sequence/arb/ArbAtom.scala
+++ b/modules/testkit/src/main/scala/lucuma/core/model/sequence/arb/ArbAtom.scala
@@ -5,6 +5,7 @@ package lucuma.core.model.sequence
 package arb
 
 import cats.syntax.all._
+import eu.timepit.refined.types.string.NonEmptyString
 import lucuma.core.data.Zipper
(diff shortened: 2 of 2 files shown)


swalker2m:
Corresponding schema changes are in https://github.com/gemini-hlsw/lucuma-odb/pull/453.  My plan is to publish `lucuma-odb-schema` as a library so that the codecs can just be used by interested clients.

swalker2m:
I've been working on this for some time, and when I started we were more cavalier about breaking changes.  If it would be helpful to restore, say, `StepTime` or any other of the changes temporarily let me know.

swalker2m:
N.B. The important changes are in the first ~1/3 of the change set.

toddburnside (approved):
LGTM

tpolecat (approved):
:+1: Looks great to me.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
```diff
@@ -4,58 +4,58 @@
...
-  def steps: List[Step]
+/**
+ * An Atom is a collection of steps that must be treated as indivisible
+ * at execution time.
+ *
+ * @tparam D dynamic config type (e.g., DynamicConfig.GmosNorth)
+ */
+case class Atom[D](
+  id:          Atom.Id,
+  description: Option[String],
```
rpiaggio, line 28: `Option[NonEmptyString]` ?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
```diff
@@ -4,58 +4,58 @@
...
-  val gmosSouth: Prism[Atom, GmosSouth] =
-    GenPrism[Atom, GmosSouth]
+  def description[D]: Lens[Atom[D], Option[String]] =
+    Focus[Atom[D]](_.description)
+
+  /** @group Optics */
+  def steps[D]: Lens[Atom[D], NonEmptyList[Step[D]]] =
+    Focus[Atom[D]](_.steps)
+
+  given [D](using Eq[D]): Eq[Atom[D]] =
```
rpiaggio, line 54: `derives Eq` ?

PR #738, opened 2023-06-06
Execution Event Types and Visit ID

swalker2m:
Switches the execution event types to `enum`s and changes the `Visit.Id` to `Gid`.  Steps and Atoms are `UUID`-based because they are not always associated with anything in the database.  I could think of no reason why `Visit` and `ExecutionEvent` should be `UUID`s though.

Nobody is using any of these things yet so it should be innocuous.


toddburnside (approved):
LGTM

PR #739, opened 2023-06-08
convert step type

swalker2m:
Before step type makes its way into the database, I'd like to change the `tag` to conform to the snake case standard.


toddburnside (approved):
LGTM

PR #758, opened 2023-07-07
move extinction from odb to here

tpolecat:
This was in `lucuma.odb.data` but it really should be here, in `lucuma.core.model`

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/Extinction.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/Extinction.scala
@@ -1,0 +1,46 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.model
+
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/Extinction.scala
```diff
@@ -0,0 +1,46 @@
...
+
+object Extinction:
+
+  def apply(millimags: NonNegShort): Extinction =
+    millimags
+
+  val FromMillimags: Prism[Short, Extinction] =
+    Prism((s: Short) => NonNegShort.from(s).toOption)(_.value)
+
+  val FromMags: Format[Double, Extinction] =
```
swalker2m, line 27: Should it be `Format[BigDecimal, Extinction]`?

swalker2m (approved):
Looks great. :+1:

modules/core/shared/src/main/scala/lucuma/core/model/Extinction.scala
tpolecat: Well it totally works. I must have been hallucinating.

PR #762, opened 2023-07-11
Geom improvements

cquiroz:
Some performance oriented improvements.
Mostly avoid allocations in geometry calculations. I ran a benchmark and observe a small but measurable improvement when doing ags calculations


toddburnside (approved):
LGTM

PR #767, opened 2023-07-21
ExecutionDigest

swalker2m:
Adds an `ExecutionDigest`, which is just the combination of setup time and the acquisition and science `SequenceDigest`s.


cquiroz (approved):
LGTM

PR #769, opened 2023-07-26
Remove digest artifacts from sequence

swalker2m:
The ODB API will be updated to provide `execution` / `digest` and `execution` / `config` queries.  The former (`digest`) returns summary information for which it is required to examine the entire sequence (a fold over all the atoms).  The later (`config`) returns the familiar `nextAtom`, `possibleFuture`, and `hasMore` for each of the `acquisition` and `science` sequences.  This enables the `digest` information to be neatly cached while the expanded `config` can be quickly computed upon request thanks to the tight `futureLimit` that is imposed upon the sequences.

This PR enables the pending ODB update by moving the `atomCount` (and `setupTime`) to the digest.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-ThisBuild / tlBaseVersion                         := "0.81"
+ThisBuild / tlBaseVersion                         := "0.82"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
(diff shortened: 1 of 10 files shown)


rpiaggio:
So the `digest` is now parallel to `execution` instead of embedded?

rpiaggio (approved):
LGTM

swalker2m:
> So the `digest` is now parallel to `execution` instead of embedded?

You can ask for both in the same query (and both are now embedded in `execution` as siblings) but it facilitates the implementation to separate that which needs to read the entire sequence from that which needs only the first few steps.

```
query {
  observation(observationId: "o-100") {
    execution {
      digest { ... }
      config { ... }
    }
  }
}
```

PR #771, opened 2023-08-01
Remove seqexec only enums

cquiroz:
These enums are only used in the seqexec (probably observe too) but it is hard to maintain with changes happening often
They are now internalized in the seqexec code


rpiaggio (approved):
LGTM

PR #792, opened 2023-08-25
Add method to calculate central wavelength for a dynamic config

cquiroz:
This is basically taken from GMOSSupport at the wdba:
https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.wdba.xmlrpc.server/src/main/java/edu/gemini/wdba/tcc/GMOSSupport.java#L41


tpolecat (approved):
:+1:

jluhrs (approved):
LGTM. Would it be worth to put it in a common interface for all instruments?

cquiroz:
I think so, as it returns an `Option[Wavelength]` in the worst case we'd return a `None`

At the moment there is no generic `DynamicConfig` in general, only for gmos longslit though

/cc @swalker2m

PR #798, opened 2023-09-01
Method to estimate what target is brightest in a collection

cquiroz:
We've been trying to answer the question of which target is brightest in an asterism and have different ways to do it in explore and lucuma-odb

Turns out the question isn't very simple to answer, in particular it is nonsensical if member of the asterism don't have comparable brightnesses profile and units. e.g. you cant really compare a target defined with emission lines and one defined with surface units

Even when the units are compatible the may refer to different bands. The implementation here uses a simplified approach where we find the nearest available band and compare their values. This is an approximation but could be refined in the future.

Note that eventually we may need to ask the user which one to use for ITC estimation if we can't deduce it

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
@@ -4,13 +4,17 @@
 package lucuma.core.model
 
 import cats.Eq
+import cats.Foldable
+import cats.Functor
 import cats.syntax.all.*
 import eu.timepit.refined.cats.given
 import lucuma.core.enums.Band
 import lucuma.core.math.Angle
 import lucuma.core.math.BrightnessUnits.*
+import lucuma.core.math.BrightnessValue
 import lucuma.core.math.Wavelength
 import lucuma.core.math.dimensional.*
+import lucuma.core.model.SpectralDefinition.BandNormalized
 import monocle.Focus
 import monocle.Lens
 import monocle.Optional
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
```diff
@@ -267,4 +273,74 @@ object SourceProfile {
...
+      (w.toPicometers.value.value - b.center.toPicometers.value.value).abs
+    }.map(x => (x._1, x._2.value, x._2.units))
+
+  extension(sp: SourceProfile)
+    /**
+     * Returns the band and brightness of a source profile for the band closest to the given wavelength.
+     * It is only valid for source profiles with integrated or surface brightnesses.
+     * (Emission lines not supported)
+     */
+    def nearestBand(wavelength: Wavelength): Option[(Band, BrightnessValue, Units)] =
```
rpiaggio, line 291: Why an extension instead of a trait method?

modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
```diff
@@ -267,4 +273,74 @@ object SourceProfile {
...
+     * Even with this limitations this method is useful for finding the brightest target in an asterism, though in general the user may need to override it.
+     *
+     * A more advanced version could also take into account the SED of the target, but this is not currently implemented.
+     */
+    def brightestAt(w: Wavelength): Option[SourceProfile] =
+      if (e.isEmpty || !canCompareBrightnesses) None
+      else e.minimumByOption(_.nearestBand(w).map(_._2.value)) }
+
+  /**
+   * This method is used to find the brightest target in an colllection of something (e.g. asterism.)
```
rpiaggio, line 342: Typo in comment (extra `l`)

rpiaggio (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/SourceProfile.scala
cquiroz: Could be in the trait, just wanted to keep the trait simple

PR #800, opened 2023-09-06
Add ZipperCodec

cquiroz:
This is taken from `lucuma-odb` but it can be used in `lucuma-itc` as well


swalker2m (approved):
Looks great, I think I was hesitant to add the codec to core since the use of 'all', 'selected' and 'index' doesn't seem necessarily universal.

cquiroz:
Maybe we should expose it as a codec rather than the canonical codec?

swalker2m (approved):
Seems better that you have to explicitly import them.  :+1:

PR #810, opened 2023-09-28
GMOS binning calculation

swalker2m:
Updates to GMOS binning calculations based on [the latest specification](https://docs.google.com/document/d/1P8_pXLRVomUSvofyVkAniOyGROcAtiJ7EMYt9wWXB0o).  This will imply changes to the sequence generation code in the ODB, not the least of which will be to update the default Y bin to use the calculation instead of a constant.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
@@ -1,0 +1,137 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.model.sequence.gmos
+
+import cats.Order
+import cats.syntax.all.*
+import coulomb.Quantity
+import eu.timepit.refined.types.numeric.PosDouble
+import eu.timepit.refined.types.numeric.PosInt
+import lucuma.core.enums.GmosNorthDetector
+import lucuma.core.enums.GmosNorthFpu
+import lucuma.core.enums.GmosNorthGrating
+import lucuma.core.enums.GmosSouthDetector
+import lucuma.core.enums.GmosSouthFpu
+import lucuma.core.enums.GmosSouthGrating
+import lucuma.core.enums.GmosXBinning
+import lucuma.core.enums.GmosYBinning
+import lucuma.core.enums.ImageQuality
+import lucuma.core.math.Angle
+import lucuma.core.math.Wavelength
+import lucuma.core.math.units.NanometersPerPixel
+import lucuma.core.model.SourceProfile
+import spire.math.Rational
+
+import java.math.RoundingMode.HALF_UP
+
+/**
+ * Optimal GMOS spectral and spatial binning calculations.
+ *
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
```diff
@@ -0,0 +1,137 @@
...
+  val DefaultSampling: PosDouble =
+    PosDouble.unsafeFrom(2.5)
+
+  /**
+   * Object angular size estimate based on source profile and image quality.
+   */
+  def objectSize(p: SourceProfile, iq: ImageQuality): Angle =
+    p match {
+      case SourceProfile.Point(_)          => iq.toAngle
+      case SourceProfile.Uniform(_)        => Angle.Angle180
```
swalker2m, line 44: The fact that the spatial binning calculation ignores the slit width (is that intended?) means that a `Uniform` source profile will always result in a y-binning of 4.  Is that expected?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
```diff
@@ -0,0 +1,137 @@
...
+  /**
+   * Minimum of the slit and object size.
+   */
+  def effectiveWidth(
+    slitWidth:  Angle,
+    srcProfile: SourceProfile,
+    iq:         ImageQuality
+  ): Angle = {
+    given Order[Angle] = Angle.AngleOrder
+    slitWidth min objectSize(srcProfile, iq)
```
cquiroz, line 62: Ah I guess the 180 is intersected with slitWidth, right?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
swalker2m: "Any" large value is fine.  I think basically it is saying that a `Uniform` source does not have a meaningful limiting size.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
swalker2m: For the spectral binning calculation yes.  In this case, the slit width sets the effective size.

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gmos/binning.scala
```diff
@@ -0,0 +1,137 @@
...
+    sampling:   PosDouble = DefaultSampling
+  ): GmosYBinning = {
+
+    // Number of unbinned pixels to sample the slit width
+    val npix = objectSize(srcProfile, iq).arcsec / (pixelScale.arcsec * sampling.value)
+
+    // The maximum binning that gives the required sampling (<npix)
+    GmosYBinning.all.tail.reverse.find { bin =>
+      bin.count <= npix
+    }.getOrElse(GmosYBinning.One)
```
swalker2m, line 118: The python code ignores `bin - npix` values  under `-10.0`, which seemed rather arbitrary and I assumed it was just to make the maximum-finding search work out. Is there actually something significant about `-10.0` though? If so, `Uniform` source profile would probably result in y-binning of `One` instead of `Four`.

PR #815, opened 2023-10-02
Add scalar angle multiplication

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
```diff
@@ -300,6 +300,20 @@ object Angle extends AngleOptics {
...
+    * @group Operations
+    */
+    def *(a: Long): Angle =
+      Angle.fromMicroarcseconds(toMicroarcseconds * a)
+
+    /**
+    * Scalar multiplication of this angle and double precission number `a`. Approximate
+    * @group Operations
+    */
+    def *?(a: Double): Angle =
```
cquiroz, line 314: I made it `*?` to denote that is approximate. not sure if it is a good idea, WDYT?

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
```diff
@@ -300,6 +300,20 @@ object Angle extends AngleOptics {
     def -(a: Angle): Angle =
       Angle.fromMicroarcseconds(toMicroarcseconds - a.toMicroarcseconds)
 
+    /**
+    * Scalar multiplication of this angle and and natural number `a`. Exact
+    * @group Operations
+    */
+    def *(a: Long): Angle =
+      Angle.fromMicroarcseconds(toMicroarcseconds * a)
```
tpolecat, line 308: Unfortunately this overflows when `a` gets large. You may need to go to `BigDecimal` and renormalize it back into 0..360° before constructing the return value.

```
scala> Angle.Angle90 * 4000000
val res14: lucuma.core.math.Angle = 0
                                                                                                                                                                            
scala> Angle.Angle90 * 40000000
val res15: lucuma.core.math.Angle = 230290448384
```

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
tpolecat: We just document it when it's inexact. Pretty much anything involving `Double` is inexact by definition.

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
cquiroz: Thanks I pushed a new commit doing as you suggest and some tests

modules/core/shared/src/main/scala/lucuma/core/math/Angle.scala
cquiroz: I reverted to `*`

tpolecat (approved):

PR #817, opened 2023-10-06
Add duration parameter to average parallactic angle calculation method

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/math/skycalc/package.scala
```diff
@@ -28,8 +29,9 @@ def parallacticAngle(site: Site, tracking: ObjectTracking, vizTime: Instant): An
   skycalc.calculate(c, vizTime, false).parallacticAngle
 }
 
-def averageParallacticAngle(site: Site, tracking: ObjectTracking, vizTime: Instant): Option[Angle] =
-  val defined: TimeRange = (vizTime.toEpochMilli(), vizTime.toEpochMilli() + 1.hour.toMillis)
+def averageParallacticAngle(site: Site, tracking: ObjectTracking, vizTime: Instant, duration: Duration): Option[Angle] =
+  val end = vizTime.toEpochMilli() + duration.toMillis()
```
swalker2m, line 33: I think this could also be `vizTime.plus(duration)`.

swalker2m (approved):
Looks good.  :+1:

modules/core/shared/src/main/scala/lucuma/core/math/skycalc/package.scala
toddburnside: Yeah. That looks cleaner. Then it's short enough to inline.

cquiroz (approved):
LGTM

PR #819, opened 2023-10-10
Dataset Id

swalker2m:
Switches `Dataset.Id` to a GID and fixes an oversight in which there were conflicting definitions for the dataset id.  I'd originally defined the dataset id as a combination step + index but this was quite challenging in Grackle / GraphQL and I'm no longer sure what I was hoping to buy with that anyway.  A simple GID seems to work fine.


toddburnside (approved):
LGTM

PR #826, opened 2023-11-06
Add ObjectTracking.constant

toddburnside:
Other methods, such as the calculation of the average parallactic angle, use the `ObjectTracking` for the base. However, if the user specifies explicit base coordinates, we don't want that to change over time. This allows us to model that behavior.


modules/core/shared/src/main/scala/lucuma/core/model/ObjectTracking.scala
```diff
@@ -37,6 +37,11 @@ object ObjectTracking:
       trackings.centerOfAt(i).map(CoordinatesAtVizTime(_))
     def baseCoordinates: Coordinates                 = trackings.centerOf
 
+  case class ConstantTracking(coordinates: Coordinates) extends ObjectTracking derives Eq:
```
cquiroz, line 40: Seems good, I thought we could do the same  with a constant `SiderealTracking` but this may be more generic

cquiroz (approved):
LGTM

PR #827, opened 2023-11-06
KeyDecoder/KeyEncoder for WithUid.Id/WithGid.Id

rpiaggio:
Adds `KeyDecoder`/`KeyEncoder` to classes that extend `WithGuid`/`WithUid`.

It doesn't seem possible to make `Gid`/`Uid` extend `KeyDecoder`/`KeyEncoder` since their `apply` signatures clash with `Decoder`/`Encoder`. (Maybe they should be named `decode`/`encode` in circe...)


hugo-vrijswijk (approved):
LGTM

PR #830, opened 2023-11-09
`TimestampInterval` and `TimestampUnion`

swalker2m:
Time accounting calculations in the `ocs` project make extensive use of its `Interval` and `Union` classes.  Here in `lucuma` we have the `BoundedInterval` and Spire's `IntervalSeq`, which appear to address similar concerns.  I will need something like this for time accounting in GPP. 

For time accounting even the `BoundedInterval` simplification of Spire's `Interval` is more general than we need thanks to `Point` instances and open vs closed bounds on each end of the interval.  We'll always want `[start, end)`.  The `IntervalSeq` in turn works with raw `Interval`s, including unbounded and empty intervals etc.

This PR would add `TimestampInterval` and `TimestampUnion` to provide a simple solution for time accounting.  I'll offer it here but I can move it to `lucuma-odb` if we think it would be confusing or not useful in other contexts.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
@@ -12,10 +12,6 @@ import org.typelevel.cats.time.*
 
 import java.time.Instant
 
-// Open/closed bounds complication.  Always want [start, end).
-// Type parameter instead of int value, allows floating point, rational etc.
-// 'Point' unnecessary.
-
 /**
  * A TimestampInterval represents a period of time with a fixed starting point
  * 'start' (inclusive) and an fixed ending point 'end' (exclusive).  In other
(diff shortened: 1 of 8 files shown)


cquiroz (approved):
Beautiful

tpolecat (approved):
This looks good. I wrote something like this for QPT a long time ago and it was really hard to get right. Generative testing probably helps a lot.

I think the set/monoid behavior of `TimestampUnion` makes it a `BoundedSemilattice`, so you might think about defining an instance.

toddburnside (approved):
Nice. I want this for the guide star availability, too. I had the same problem with `BoundedInterval` being too general.

modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
```diff
@@ -0,0 +1,236 @@
...
+/**
+ * A TimestampInterval represents a period of time with a fixed starting point
+ * 'start' (inclusive) and an fixed ending point 'end' (exclusive).  In other
+ * words, [start, end).  Whereas a TimeSpan focuses solely on an amount of
+ * time, the TimestampInterval tracks specific start and end times.
+ *
+ * @param start time at which the interval begins (inclusive)
+ * @param end time at which the interval ends (exclusive)
+ */
+sealed class TimestampInterval private (val start: Timestamp, val end: Timestamp) {
```
rpiaggio, line 24: Nice.

My first impression was that this could `extend BoundedInterval[Timestamp]` but upon closer inspection not all methods convey the same information, especially information about left-closed, right-open resulting intervals.

But there are number of methods that seems the same: `abuts`, `contains`, `intersects` at least. 

Maybe this class can wrap a `BoundedInterval[Timestamp]` and just export the methods that are the same? Maybe it can rely on other logic from `BoundedInterval` for other operations too.

rpiaggio (approved):
LGTM

swalker2m:
> I wrote something like this for QPT a long time ago

Yes this is a copy, translated to Scala, with a few tweaks. I used the old version in `ocs` to simplify (tremendously) the time accounting implementation there.

> I think the set/monoid behavior of TimestampUnion makes it a BoundedSemilattice, so you might think about defining an instance.

Thanks for pointing that out!  I added an instance and law checks.

swalker2m:
> Maybe this class can wrap a BoundedInterval[Timestamp] and just delegate the methods that are the same? 

I wanted to give this a try, but with `TimestampInterval` as an `opaque type` defined as a `BoundedInterval[Timestamp]`.  Unfortunately I wasn't able to get an `extension` object with methods like `abuts` that are also defined in `BoundedInterval` to work -- the compiler insists on interpreting them as recursive calls.

Perhaps defining it as an `Interval[Timestamp]` instead (so it isn't an `opaque type` of an `opaque type`) would help.  I decided that the win though would be just avoiding the definition of `abuts` and `contains` which are pretty trivial anyway so I'll just go with this as it is.  Definitely not opposed to switching it over if someone wants to continue investigating.

PR #833, opened 2023-11-13
Add `intersection` and `timeBetween` to TimestampInterval

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
```diff
@@ -127,6 +128,35 @@ sealed class TimestampInterval private (val start: Timestamp, val end: Timestamp
   def intersects(other: TimestampInterval): Boolean =
     overlap(other).intersects
 
+  /**
+   * Determine the intersection between 2 intervals, or None if they don't
+   * intersect.
+   */
+  def intersection(other: TimestampInterval): Option[TimestampInterval] =
+    overlap(other) match
+      case Overlap.None => None
```
cquiroz, line 137: Non aligned arrows?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
toddburnside: Oh, that's right. I need to manually format... Thanks.

PR #835, opened 2023-11-14
Change `TimestampInterval.minus` wrt "empty" intervals.

toddburnside:
There was some edge cases in `minus` related to intervals where `start === end`. 

In 2 cases, the resulting List of intervals would produce adjacent intervals that would "sum" to what I expect. Such as (using the test notation):
> List(interval(0, 3), interval(3, 10))

In 2 other cases, the result contained a "empty" interval disjoint from what I would expect. Such as:
> List(interval(0, 3), interval(10, 10))

I will put comments in the test denoting the exact behavior that changed.

@swalker2m - You'll have to let me know if this matches your expectations for your use in the `TimestampUnion` or not.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
@@ -85,12 +85,15 @@ sealed class TimestampInterval private (val start: Timestamp, val end: Timestamp
    */
   def minus(other: TimestampInterval): List[TimestampInterval] =
     overlap(other) match {
-      case Overlap.LowerPartial   => List(between(start, other.start))
-      case Overlap.UpperPartial   => List(between(other.end, end))
-      case Overlap.ProperSuperset => List(between(start, other.start), between(other.end, end))
-      case Overlap.None           => List(this)
+      case Overlap.LowerPartial                                  => List(between(start, other.start))
+      case Overlap.UpperPartial                                  => List(between(other.end, end))
+      case Overlap.ProperSuperset if (other.start === other.end) => List(this)
+      case Overlap.ProperSuperset if (start === other.start)     => List(between(other.end, end))
+      case Overlap.ProperSuperset if (end === other.end)         => List(between(start, other.start))
+      case Overlap.ProperSuperset                                => List(between(start, other.start), between(other.end, end))
+      case Overlap.None                                          => List(this)
       case Overlap.ProperSubset |
-           Overlap.Equal          => Nil
+           Overlap.Equal                                         => Nil
     }
 
   /**
--- a/modules/tests/shared/src/test/scala/lucuma/core/util/TimestampIntervalSuite.scala
+++ b/modules/tests/shared/src/test/scala/lucuma/core/util/TimestampIntervalSuite.scala
@@ -125,16 +125,29 @@ class TimestampIntervalSuite extends DisciplineSuite {
     assertEquals(interval(1, 10).minus(interval(9, 11)), List(interval(1, 9)))
 
     // Proper Superset
-    assertEquals(interval(0, 10).minus(interval(3,  6)), List(interval(0, 3), interval(6, 10)))
+    assertEquals(interval(0, 10).minus(interval(3, 6)), List(interval(0, 3), interval(6, 10)))
+    assertEquals(interval(0, 10).minus(interval(0, 6)), List(interval(6, 10)))
+    assertEquals(interval(0, 10).minus(interval(3, 10)), List(interval(0, 3)))
+    // with empty intervals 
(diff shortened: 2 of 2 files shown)


modules/tests/shared/src/test/scala/lucuma/core/util/TimestampIntervalSuite.scala
```diff
@@ -125,16 +125,29 @@ class TimestampIntervalSuite extends DisciplineSuite {
     assertEquals(interval(1, 10).minus(interval(9, 11)), List(interval(1, 9)))
 
     // Proper Superset
-    assertEquals(interval(0, 10).minus(interval(3,  6)), List(interval(0, 3), interval(6, 10)))
+    assertEquals(interval(0, 10).minus(interval(3, 6)), List(interval(0, 3), interval(6, 10)))
+    assertEquals(interval(0, 10).minus(interval(0, 6)), List(interval(6, 10)))
```
toddburnside, line 129: This used to produce: `List(interval(0, 0), interval(6, 10))`

modules/tests/shared/src/test/scala/lucuma/core/util/TimestampIntervalSuite.scala
```diff
@@ -125,16 +125,29 @@ class TimestampIntervalSuite extends DisciplineSuite {
...
+    assertEquals(interval(0, 10).minus(interval(3, 10)), List(interval(0, 3)))
```
toddburnside, line 130: This used to produce: List(interval(0, 3), interval(10, 10))

modules/tests/shared/src/test/scala/lucuma/core/util/TimestampIntervalSuite.scala
```diff
@@ -125,16 +125,29 @@ class TimestampIntervalSuite extends DisciplineSuite {
...
+    // with empty intervals 
+    assertEquals(interval(0, 10).minus(interval(0, 0)), List(interval(0, 10)))
```
toddburnside, line 132: This used to produce: List(interval(0, 0), interval(0, 10))

modules/tests/shared/src/test/scala/lucuma/core/util/TimestampIntervalSuite.scala
```diff
@@ -125,16 +125,29 @@ class TimestampIntervalSuite extends DisciplineSuite {
...
+    assertEquals(interval(0, 10).minus(interval(3, 3)), List(interval(0, 10)))
```
toddburnside, line 133: This used to produce: List(interval(0, 3), interval(3, 10))

modules/core/shared/src/main/scala/lucuma/core/util/TimestampInterval.scala
```diff
@@ -85,12 +85,15 @@ sealed class TimestampInterval private (val start: Timestamp, val end: Timestamp
...
+      case Overlap.LowerPartial                                  => List(between(start, other.start))
+      case Overlap.UpperPartial                                  => List(between(other.end, end))
+      case Overlap.ProperSuperset if (other.start === other.end) => List(this)
+      case Overlap.ProperSuperset if (start === other.start)     => List(between(other.end, end))
+      case Overlap.ProperSuperset if (end === other.end)         => List(between(start, other.start))
+      case Overlap.ProperSuperset                                => List(between(start, other.start), between(other.end, end))
+      case Overlap.None                                          => List(this)
       case Overlap.ProperSubset |
-           Overlap.Equal          => Nil
+           Overlap.Equal                                         => Nil
```
swalker2m, line 96: Hmm, I wonder if this should produce a `TimestampUnion` instead of a `List[TimestampInterval]`.  In the original code when the `List` is added to the union it gets merged in as expected.

toddburnside: So the disjoint "empty" intervals didn't cause a problem? It seems like that would create a Union with 2 pieces. But, maybe the empty piece is expected?

toddburnside: Expected by your code, that is. Mine doesn't expect it. But, I could filter it out.

swalker2m: An empty interval would add nothing to the sum, so in that sense it doesn't matter for time accounting.

swalker2m (approved):
Thanks, this seems fine. The number of cases is annoying but your result is much more intuitive. :+1:

toddburnside:
> Thanks, this seems fine. The number of cases is annoying but your result is much more intuitive. 👍

OK. Fortunately, it only seems to be odd wrt `minus`.

PR #836, opened 2023-11-14
Downgrade cats-parse to 0.3.10

toddburnside:
Shane and I discussed this and decided it would be better to downgrade cats-parse until http4s is on the new version than to propagate the sbt `libraryDependencySchemes` hack throughout our projects.

Any disagreements?


swalker2m (approved):

toddburnside:
Http4s just dropped the new version.

PR #837, opened 2023-11-15
Night Syntax

swalker2m:
Adds `toTimestampInterval` syntax to `Night`.  `Night`s are `BoundedInterval[Instant]` paired with a `Site`.  I'm kind of wishing it were instead a `TimestampInterval` with a `Site`, but it would likely cause hardship to refactor it now?


cquiroz:
No problem refactoring downstream

cquiroz (approved):
LGTM

toddburnside:
I don't think switching to TimestampInterval would be difficult in Explore. If it is switched, other classes like TwilightBoundedNight should probably also be changed.

PR #839, opened 2023-11-23
CategorizedTime

swalker2m:
Only belatedly did I realize that the breakdown of time into program, partner and non-charged categories for the planned time calculation is the same information I need for time accounting.  Yet it is unseemly to talk about `PlannedTime` for cost that has already been incurred.

This PR adds `CategorizedTime` which is basically a copy of `PlannedTime` with a new name. It will be employed for time charges after the fact as well as planned time before.  Similarly, there is now `CategorizedTimeRange` for `PlannedTimeRange`.  In the API, `plannedTime` fields have been marked deprecated and replaced with `timeEstimate`s returning `CategorizedTime`.

I've left `PlannedTime` in the model for now though because it still appears in the API (albeit deprecated).  Once we can remove it from the API it can also be removed from `core`.  I don't like the name `CategorizedTime` but nothing better occurred to me.  I started with `TimeCharge` but that seemed inappropriate for a planned time estimate.  Suggestions for improvement are most welcome.

Over in the API, `CategorizedTime` will differ slightly from `PlannedTime`.  It will remove the list of charges (which could have duplicate entries for any category) and instead explicitly list the categories:

```
"A time amount broken into charge class categories."
type CategorizedTime {

  "Time charged to the program / PI."
  program: TimeSpan!

  "Time charged to the partner."
  partner: TimeSpan!

  "Execution time that is not charged."
  nonCharged: TimeSpan!

  "Total of program, partner and uncharged times."
  total: TimeSpan!

}
```

Finally it adds a `TimeChargeCorrection` which is used to add manual corrections to time charges.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-ThisBuild / tlBaseVersion                         := "0.89"
+ThisBuild / tlBaseVersion                         := "0.90"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/Atom.scala
@@ -28,8 +28,8 @@ case class Atom[D](
   lazy val observeClass: ObserveClass =
     steps.foldMap(_.observeClass)
 
-  lazy val plannedTime: PlannedTime =
-    steps.foldMap(_.plannedTime)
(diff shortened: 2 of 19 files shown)


swalker2m:
I'm vying for an award for most verbose description of trivial changes.  Wish me luck.

swalker2m:
Corresponding ODB updates: https://github.com/gemini-hlsw/lucuma-odb/pull/841

toddburnside (approved):
LGTM

swalker2m:
I'm going to piggyback a couple of additive (i.e., not breaking anything already out there) updates to `TimestampInterval` on this, if you will allow.  They're all golden, I promise...

PR #844, opened 2023-12-07
`CategorizedTime.equals` vs. `Order[CategorizedTime]`

swalker2m:
`CategorizedTime` is an `opaque type` that is a `SortedMap` at heart. When a `ChargeClass` has `Zero` time associated with it, the map might or might not have an entry for that charge class.  This makes comparison with `.equals`, which just compares the two underlying `SortedMap`s,  inconsistent with the `Order` implementation and frustrates testing.  For example `SortedMap.empty != SortedMap(ChargeClass.Program -> TimeSpan.Zero)` but `Order` considers the two `CategorizedTime`s to be equal.

I've "fixed" this by ensuring that a charge class with a  `TimeCharge.Zero` is never present in the map.  This may be considered a bit fragile.  An alternative would be to just convert `CategorizedTime` into a straightforward `case class`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,4 +1,4 @@
-ThisBuild / tlBaseVersion                         := "0.90"
+ThisBuild / tlBaseVersion                         := "0.91"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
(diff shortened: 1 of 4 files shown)


swalker2m:
> Would changing to a case class end up complicating things elsewhere?

No, I don't think so.  Nobody else uses this yet, I think, and I'm inclined to do it as well.  I'll switch since it would be so easy to mess up in the future.

tpolecat (approved):
:+1: looks great!

PR #854, opened 2024-01-31
DatasetStage enum value rename

swalker2m:
Discussion with Javier and Carlos revealed a preference to rename `DatasetStage` enum values from `StartObserve` and `EndObserve` to `StartExpose` and `EndExpose` respectively to avoid confusion with `StepStage` start/end observe.


cquiroz (approved):
LGTM

PR #863, opened 2024-02-26
Reference updates

swalker2m:
Updates core in support of program and proposal references in the ODB.  Most of these changes were added to the ODB but would be removed in favor of core.

* Changes `ProgramType` to enumerate only the broad program categories.

* Adds `ScienceSubtype` enumeration of the types of science programs.

* Adds `ProgramReference` and `ProposalReference`, which define the new science-mandated program ids.

* Refactors `Semester` to limit the semester years to those supported in the ODB.  Previously the semester definition allowed any `java.time.Year` but the parser only supported 4 digit years.


toddburnside (approved):
LGTM

tpolecat (approved):

PR #871, opened 2024-03-14
DatasetReference

swalker2m:
Adds the `DatasetReference` here in `lucuma-core` in order to, ultimately, take the place of the one in `lucuma-odb`.


toddburnside (approved):
LGTM

PR #872, opened 2024-03-14
Execution Event Updates

swalker2m:
After discussion about execution events, I believe the consensus is:

* `Abort`, `Continue`, `Pause`, `Start` have an interpretation at the sequence level and will remain there
* They also need to be present at the step level and are added in this PR
* We're missing slew events

but please correct me if I'm wrong.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/SlewStage.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/SlewStage.scala
@@ -7,7 +7,7 @@ import lucuma.core.util.Enumerated
 
 enum SlewStage(val tag: String, val name: String, val description: String) derives Enumerated {
 
-  case Start extends SlewStage("start", "Slew Start", "Slew started.")
-  case End   extends SlewStage("stop",  "Stop Stop",  "Slew stopped.")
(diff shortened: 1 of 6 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/SlewStage.scala
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+enum SlewStage(val tag: String, val name: String, val description: String) derives Enumerated {
+
+  case Start extends SlewStage("start_slew", "Start Slew", "Slew started.")
```
cquiroz, line 10: Should this be called `StartSlew/EndSlew`?

swalker2m: Yes, that would be better.

cquiroz (approved):
LGTM

PR #874, opened 2024-03-18
Move guide config model from observe/seqexec

cquiroz:
We want navigate and observe to communicate with each other., sharing the state of telescope guiding

At the moment they have different models to do so but we can unify them and move them here to `lucuma-core`

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
@@ -8,8 +8,8 @@ import lucuma.core.util.Enumerated
 /**
  * Enumerated type for guide probe
  */
-enum GuideProbe(val tag: String) extends Product with Serializable derives Enumerated {
-  case PWFS1     extends GuideProbe("Pwfs2")
-  case PWFS2     extends GuideProbe("Pwfs1")
+enum GuideProbe(val tag: String) derives Enumerated {
+  case PWFS1     extends GuideProbe("Pwfs1")
+  case PWFS2     extends GuideProbe("Pwfs2")
   case GmosOIWFS extends GuideProbe("GmosOiwfs")
 }
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
@@ -1,0 +1,15 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for guide probe
+ */
+enum GuideProbe(val tag: String) extends Product with Serializable derives Enumerated {
+  case PWFS1     extends GuideProbe("Pwfs2")
+  case PWFS2     extends GuideProbe("Pwfs1")
+  case GmosOIWFS extends GuideProbe("GmosOiwfs")
+}
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/StepGuideState.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/StepGuideState.scala
@@ -8,9 +8,9 @@ import lucuma.core.util.Enumerated
 /**
  * An enumeration whose value determines whether guiding is enabled for a particular step.
  */
-enum GuideState(val tag: String, val name: String, val description: String) derives Enumerated {
+enum StepGuideState(val tag: String, val name: String, val description: String) derives Enumerated {
 
-  case Enabled  extends GuideState("enabled",  "Enabled",  "Guiding is enabled.")
-  case Disabled extends GuideState("disabled", "Disabled", "Guiding is disabled.")
+  case Enabled  extends StepGuideState("enabled",  "Enabled",  "Guiding is enabled.")
+  case Disabled extends StepGuideState("disabled", "Disabled", "Guiding is disabled.")
 
 }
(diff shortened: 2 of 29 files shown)


toddburnside (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/StepGuideState.scala
```diff
@@ -8,9 +8,9 @@ import lucuma.core.util.Enumerated
 /**
  * An enumeration whose value determines whether guiding is enabled for a particular step.
  */
-enum GuideState(val tag: String, val name: String, val description: String) derives Enumerated {
+enum StepGuideState(val tag: String, val name: String, val description: String) derives Enumerated {
```
cquiroz, line 11: I renamed this as it really is a property for a `Step`, but now I see it changes `lucuma-schemas`. Should I revert?

cquiroz: And there is also a `GuideState` on `navigate` which refers to the guiding state in the telescope

modules/core/shared/src/main/scala/lucuma/core/model/GuideConfig.scala
```diff
@@ -0,0 +1,263 @@
...
+import monocle.Focus
+import monocle.Lens
+import monocle.Optional
+import monocle.std.either
+import monocle.std.option
+
+case class GuideConfig(
+  tcsGuide:      TelescopeGuideConfig,
+  gaosGuide:     Option[Either[AltairConfig, GemsConfig]],
+  gemsSkyPaused: Boolean
```
swalker2m, line 44: I know nothing whatsoever about this, but I wonder why `gemsSkyPaused` isn't just part of `GemsConfig`.

modules/core/shared/src/main/scala/lucuma/core/enums/StepGuideState.scala
toddburnside: I think I should leave this question to someone who is actually using this stuff...

modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
```diff
@@ -0,0 +1,15 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for guide probe
+ */
+enum GuideProbe(val tag: String) extends Product with Serializable derives Enumerated {
```
rpiaggio, line 11: Do we need `Product` and `Serializable` on an enum?

modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
```diff
@@ -0,0 +1,15 @@
...
+  case PWFS1     extends GuideProbe("Pwfs2")
+  case PWFS2     extends GuideProbe("Pwfs1")
```
swalker2m, line 13: Oops, the tag doesn't match the name.

modules/core/shared/src/main/scala/lucuma/core/model/GuideConfig.scala
jluhrs: Ooh, someone is exposing my dirty secrets. `gemsSkyPaused` is just an internal flag used when dealing with GeMS, it is not related to the GeMS configuration applied in TCC. I just needed a place to store it, and `GuideConfig`  seemed convenient at the time.

modules/core/shared/src/main/scala/lucuma/core/enums/StepGuideState.scala
rpiaggio: Does the `GuideState` on navigate take other values? If it takes the same values, maybe it can be converted to a `NewType` of this one? (And leave this one unchanged)

modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
cquiroz: Should they match?

modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
cquiroz: Nope, actually this was a test because the `Codec` test was failing. it seems circe tries to test serialization as well which we don't care.
I'll remove this

modules/core/shared/src/main/scala/lucuma/core/enums/StepGuideState.scala
cquiroz:
```scala
case class GuideState(
  mountOffload: MountGuideOption,
  m1Guide:      M1GuideConfig,
  m2Guide:      M2GuideConfig
) derives Eq
```

modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
toddburnside: It `seems` like PWFS1 should extend `GuideProbe("Pwfs1")`...

modules/core/shared/src/main/scala/lucuma/core/model/GuideConfig.scala
swalker2m: Seemed a bit odd that it'd be there whether or not working with GeMS.

modules/core/shared/src/main/scala/lucuma/core/enums/GuideProbe.scala
cquiroz: Uh I see

swalker2m (approved):
Looks good. :+1:

PR #875, opened 2024-03-18
Convert implicits to givens

cquiroz:
Since we are going to  update the universe, why not covert some implicits to givens


swalker2m (approved):
This one's going to take a bite!  Looks good.  :+1:

PR #881, opened 2024-04-02
Claim 'c' for Call for Proposals

swalker2m:
We define GID type `c` for `Configuration` but configurations are no-longer top-level objects and this GID type was never used.  I'd like to repurpose it for Call for Proposals objects.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/ids.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/ids.scala
@@ -6,9 +6,9 @@ package lucuma.core.model
 import lucuma.core.util.WithGid
 import lucuma.refined.*
 
-object Configuration  extends WithGid('c'.refined)
-object Group          extends WithGid('g'.refined)
-object ObsAttachment  extends WithGid('a'.refined)
-object Observation    extends WithGid('o'.refined)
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/ids.scala
```diff
@@ -6,9 +6,9 @@ package lucuma.core.model
 import lucuma.core.util.WithGid
 import lucuma.refined.*
 
-object Configuration  extends WithGid('c'.refined)
```
cquiroz, line 9: No `Configuration`?

swalker2m: I couldn't find any reference to it and believe it is a relic from the past when we had top-level objects (identified with a GID) for configurations and constraints.

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/ids.scala
swalker2m: I'll leave it open until tomorrow or so in case anybody wants to cry foul.

tpolecat (approved):
Looks fine. 👍🏻

PR #883, opened 2024-04-03
Move some types and functions from explore that should be in core

cquiroz:
Some of these should be common across explore and lucuma-odb


rpiaggio (approved):
LGTM

PR #884, opened 2024-04-04
Add prism for refined types

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala
```diff
@@ -10,6 +10,7 @@ import eu.timepit.refined.api.*
 import io.circe.Decoder
 import io.circe.Encoder
 import monocle.Iso
```
rpiaggio, line 12:
```suggestion
import lucuma.core.optics.refinedPrism
import monocle.Iso
```

modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala
```diff
@@ -40,5 +41,6 @@ trait NewType[Wrapped]:
 
 trait RefinedNewType[T, P](using RefinedType.AuxT[T Refined P, T]) extends NewType[T Refined P]:
   private val TypeOps = new RefinedTypeOps[T Refined P, T]
+  def from: Prism[T, Type] = Prism[T, Type](from(_).toOption)(_.value.value)
```
rpiaggio, line 44:
```suggestion
  def from(using Validate[T, P]): Prism[T, Type] = refinedPrism.andThen(value.reverse)```

rpiaggio (approved):
LGTM. The suggestion is a bit of nitpicking, but composing existing optics instead of building a new one guarantees lawfulness in the general case.

cquiroz:
Great idea

PR #885, opened 2024-04-11
ShortCut 2772

swalker2m:
Fixes the `ImageQuality.toAngle` calculation, which corrects ShortCut 2772.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/ImageQuality.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/ImageQuality.scala
@@ -21,7 +21,7 @@ sealed abstract class ImageQuality(val tag: String, val toDeciArcSeconds: Quanti
   def label: String        = f"""< ${toArcSeconds.value.toDouble}%.1f\""""
 
   def toAngle: Angle =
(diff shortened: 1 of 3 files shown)


cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/ImageQuality.scala
```diff
@@ -21,7 +21,7 @@ sealed abstract class ImageQuality(val tag: String, val toDeciArcSeconds: Quanti
   def label: String        = f"""< ${toArcSeconds.value.toDouble}%.1f\""""
 
   def toAngle: Angle =
-    Angle.fromMicroarcseconds(toDeciArcSeconds.tToUnit[MicroArcSecond].value.value.toLong)
+    Angle.fromMicroarcseconds(toDeciArcSeconds.value.value * 100_000L)
```
swalker2m, line 24: If anybody understands coulomb and feels sufficient anxiety about bypassing it, I'd welcome input on how the conversion stuff is supposed to work.  I'm happy with just making it work though.

cquiroz: @armanbilge Can you enlighten us?

PR #886, opened 2024-04-23
Fix objectSize calculation for Gaussian profile

swalker2m:
From discussion in Slack, the Gaussian profile contribution to the object size needs an adjustment.


cquiroz (approved):
LGTM

PR #890, opened 2024-04-26
Atom Events

swalker2m:
This PR prepares the way for adding atom events to the API.  Atom events will allow Explore to display the sequence in a more intuitive way.  It only occurred to me to add start / end since it doesn't seem like abort, etc. work at the atom level?


rpiaggio (approved):
LGTM

PR #891, opened 2024-05-05
Add poor weather (P) science subtype

swalker2m:
This is the first step in updating program references to support explicit poor weather subtypes (e.g., `G-2025A-0001-P`).


cquiroz (approved):
LGTM

PR #893, opened 2024-05-27
Move `Partner` to `enums` package

swalker2m:
@toddburnside  pointed me to the existence of the Scala `Partner` enum, which will now agree with the database `t_partner` table once the time accounting categories are removed.  To avoid duplication and make it easier to work with, I removed the dynamic `Partner` definition in the ODB in favor of the Scala `Partner`.  This PR moves the `Partner` enumeration from `model` to the `enums` package and beefs it up a bit with the `sites` `Set`.


toddburnside (approved):
LGTM. It's nice to have Partner refer to the same thing in both the ODB and Explore.

PR #905, opened 2024-06-10
Make dayTimeMode optional

cquiroz:
This is to make observe compatible with the TCC

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/GuideConfig.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/GuideConfig.scala
@@ -59,7 +59,7 @@ object GuideConfig {
         MountGuideOption.MountGuideOff,
         M1GuideOff,
         M2GuideOff,
-        false,
+        None,
         None),
       None,
     )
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/GuideConfig.scala
```diff
@@ -59,7 +59,7 @@ object GuideConfig {
         MountGuideOption.MountGuideOff,
         M1GuideOff,
         M2GuideOff,
-        false,
+        None,
```
swalker2m, line 62: How is a `None` value interpreted?

cquiroz: false, but it is not really used on observe

swalker2m (approved):
Seems like it shouldn't be an option in Navigate / TCC ?  At any rate :+1:

cquiroz:
In navigate as yyou said it shouldn't be an option but we don't wantn to modify TCC
I think it should return to `Option` once we kill TCC

PR #906, opened 2024-06-12
Move a few things from the ODB

toddburnside:


modules/core/shared/src/main/scala/lucuma/core/enums/CallForProposalsType.scala
```diff
@@ -0,0 +1,26 @@
...
+enum CallForProposalsType(val tag: String, val title: String) derives Enumerated:
+  case DemoScience        extends CallForProposalsType("demo_science", "Demo Science")
+  case DirectorsTime      extends CallForProposalsType("directors_time", "Director's Time")
+  case FastTurnaround     extends CallForProposalsType("fast_turnaround", "Fast Turnaround")
+  case LargeProgram       extends CallForProposalsType("large_program", "Large Program")
+  case PoorWeather        extends CallForProposalsType("poor_weather", "Poor Weather")
+  case RegularSemester    extends CallForProposalsType("regular_semester", "Regular Semester")
+  case SystemVerification extends CallForProposalsType("system_verification", "System Verification")
+
+  def subTypes: NonEmptyList[ScienceSubtype] =
```
toddburnside, line 18: This was included in the Explore version of this enum, so I ported it here.

cquiroz: I think it is generally useful

PR #915, opened 2024-07-04
Moving a couple of classes from ODB

swalker2m:
Adds a couple of (unrelated) things that were introduced in the ODB for convenience, but which belong here in core.  `DateInterval` is a bit sparse, but has everything needed in the ODB for now.


cquiroz (approved):
LGTM

PR #923, opened 2024-07-30
Add degree hms/dms utils to npm package

hugo-vrijswijk:
Useful for navigate. I've made the interface the same as the utility functions currently in navigate-(ui|configs), to make it slightly easier to use from the js side.

### DIFF
--- a/modules/npm/src/main/scala/lucuma/core/exports.scala
+++ b/modules/npm/src/main/scala/lucuma/core/exports.scala
@@ -3,91 +3,12 @@
 
 package lucuma.core
 
-import cats.syntax.all.*
-import lucuma.core.enums.Site
-import lucuma.core.enums.TwilightType
 import lucuma.core.math.Angle
-import lucuma.core.math.Coordinates
 import lucuma.core.math.HourAngle
 import lucuma.core.math.parser.AngleParsers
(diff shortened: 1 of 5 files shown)


modules/npm/src/main/scala/lucuma/core/exports.scala
```diff
@@ -3,27 +3,31 @@
...
-import lucuma.core.enums.TwilightType
+import scala.scalajs.js
 import scala.scalajs.js.JSConverters.*
+import scala.scalajs.js.annotation.JSExportTopLevel
 
 /**
  * JS facade to plot points calculated with the method `nightPlot`.
  */
 @JSExportTopLevel("PlotPoint")
 class PlotPoint(val instant: js.Date, val airmass: Double, val altitude: Double) extends js.Object {
```
hugo-vrijswijk, line 29: @cquiroz Is this used anywhere? If it can be removed I'd like to. It adds about 350kB to the bundle size for navigate that cannot be tree-shaked away

cquiroz: @stroncod Are you using this?

stroncod: We are not currently using it in the Scheduler so you can go ahead and remove it if you want.

PR #925, opened 2024-08-04
Allow the average parallactic angle to have a fallback airmass parameter

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/math/skycalc/package.scala
```diff
@@ -78,15 +94,14 @@ def averageParallacticAngle(site: Site, tracking: ObjectTracking, vizTime: Insta
...
+              if (dec - place.latitude.toAngle.toSignedDoubleDegrees < -10) 0
+              else if (dec - place.latitude.toAngle.toSignedDoubleDegrees < 10) 180
               else 360
             }
             angle + normalizingFactor
           } else angle
 
-        // val weight = if (airmass <= 1.0) 0.0 else 1.6 * math.pow(airmass - 1.0, 0.6)
-        val weight = if (airmass <= 1.0) 0.0 else math.pow(airmass - 1.0, 1.3)
+        val weight = if (airmass <= 0.0) fallbackAirmass else math.pow(airmass - 1.0, 1.3)
```
swalker2m, line 104: If `airmass` is less than 1, this will produce a negative `weight`?  Was that intended?

cquiroz: Good point. Do you think we should add a refinement? It probably should be 1 or higher @andrewwstephens

andrewwstephens: Some airmass algorithms may give a negative value when the target is below the horizon, i.e. the zenith angle is > 90 deg.     1.0 / math.cos(91 / 180 * math.pi) = -57.29

andrewwstephens: It seems like the original weight formula was okay?  I'm trying to remember where the 1.6 came from...

cquiroz: this was mostly copied from the ocs. the 1.6 is also commented there
https://github.com/gemini-hlsw/ocs/blob/312c0c3baca369ba00cb271a705212fad15804c9/bundle/edu.gemini.util.skycalc/src/main/scala/edu/gemini/util/skycalc/calc/TargetCalculator.scala#L88

swalker2m (approved):
Looks great. :+1:

PR #926, opened 2024-08-19
+ `PartnerLink`,  `ProgramUserRole`

swalker2m:
Adds a couple of new types from the ODB, from which they will be removed.


modules/core/shared/src/main/scala/lucuma/core/enums/PartnerLinkType.scala
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+enum PartnerLinkType(val tag: String) derives Enumerated:
```
cquiroz, line 8: embracing whitespace?

cquiroz (approved):
LGTM

tpolecat (approved):

Issue #933, opened 2024-08-26
Negative LST 


swalker2m:
Add this test case at the end of the ODB's `createCallForProposals`:

```scala
  test("bug") {
    expect(
      user = staff,
      query = """
        mutation {
          createCallForProposals(
            input: {
              SET: {
                type: FAST_TURNAROUND
                semester: "2024B"
                activeStart: "2024-10-01"
                activeEnd: "2024-12-31"
                submissionDeadlineDefault: "2024-08-31T22:00:00Z"
                partners: [{ partner: US }]
                instruments: [GMOS_NORTH, GMOS_SOUTH]
              }
            }
          ) {
            callForProposals {
              id
            }
          }
        }
      """,
      expected = json"""
        {
          "createCallForProposals": {
            "callForProposals": {
              "id": "c-108"
            }
          }
        }
      """.asRight
    )
  }
```

It will produce a stack trace:

```
java.time.DateTimeException: Invalid value for HourOfDay (valid values 0 - 23): -4
	at java.base/java.time.temporal.ValueRange.checkValidValue(ValueRange.java:319)
	at java.base/java.time.temporal.ChronoField.checkValidValue(ChronoField.java:718)
	at java.base/java.time.LocalTime.of(LocalTime.java:341)
	at lucuma.core.math.skycalc.ImprovedSkyCalcMethods.setHours(ImprovedSkyCalcMethods.scala:66)
	at lucuma.core.math.skycalc.ImprovedSkyCalcMethods.getLst(ImprovedSkyCalcMethods.scala:47)
	at lucuma.core.math.skycalc.ImprovedSkyCalcMethods.getLst$(ImprovedSkyCalcMethods.scala:17)
	at lucuma.core.math.skycalc.ImprovedSkyCalc.getLst(ImprovedSkyCalc.scala:18)
	at lucuma.core.math.skycalc.ImprovedSkyCalc.getLst(ImprovedSkyCalc.scala:199)
	at lucuma.odb.graphql.input.CoordinateLimitsInput$.lstAt$1(CoordinateLimitsInput.scala:61)
	at lucuma.odb.graphql.input.CoordinateLimitsInput$.defaultRaLimits(CoordinateLimitsInput.scala:63)
	at lucuma.odb.graphql.input.CoordinateLimitsInput$Create$.default(CoordinateLimitsInput.scala:115)
[...]

PR #934, opened 2024-08-26
Handle negative LST again

swalker2m:
Addresses #933 via the same [fix applied to the OCS codebase](https://github.com/gemini-hlsw/ocs/pull/2051).


tpolecat (approved):
👍🏻

PR #955, opened 2024-10-15
Change ConfigurationRequestStatus.tag to lowercase to match DB

toddburnside:
Because we have lots of enums of various case schemes, or some reason I thought the Codecs handled the switch to all lowercase for the DB. It turns out I was wrong. The `tag` has to match the DB enum exactly.

Against all odds, this does not seem to break bincompat. 😌


tpolecat (approved):
oops 👍🏻

PR #959, opened 2024-10-21
Add arbitraries for ConfigurationRequest and friends.

toddburnside:
I also threw in another Eq instance as a bonus.


rpiaggio (approved):
LGTM

PR #963, opened 2024-10-23
fix keyboard flub

tpolecat:
I failed to see this in #961, oops.


toddburnside (approved):
Thanks for finding and fixing this. :)

PR #972, opened 2024-11-11
Convenience methods to build GmosCcdMode instances

rpiaggio:
This code was in explore but I think it should be moved here to be shared with the ODB so that we can consolidate the logic.


cquiroz (approved):
LGTM

PR #973, opened 2024-11-15
Add execution state to sequence digest

swalker2m:
Adds the observation execution state to the sequence digest, so that it will be cached in the ODB.

__Note:__ It's a little misleading to put an `ObservationExecutionState` in the `SequenceDigest`, since at that level it is really referring to the sequence execution state. In reality, the acquisition execution state is never "completed" and the _science_ sequence execution state is synonymous with the observation execution state as a whole.  I could rename things to make this clearer if necessary.


rpiaggio (approved):
LGTM

You have my vote for renaming things and/or adding a comment so that it's clearer for a maintainer what's being reflected.

PR #976, opened 2024-11-27
Changes to support emission lines throughout, especially ITC

rpiaggio:
- Adds a method to find the `nearestLine` in the distribution, also relocate `nearestBand`.
- Adds a standard `Display[BigDecimal]` and forwards `Display` instances for refined types.
- Explicit display and serialization for velocity units, since the derived ones from coulomb are not ITC friendly.
- Move `Quantity` -> `Measure` extension methods to a `syntax` package.


modules/core/shared/src/main/scala/lucuma/core/math/units.scala
```diff
@@ -43,8 +43,16 @@ trait units {
   type PicometersPerPixel = Picometer / Pixels
 
   type MetersPerSecond      = Meter / Second
+  given DerivedUnit[MetersPerSecond, Meter / Second, "meters per second", "m/s"] = DerivedUnit()
+  given TypeString[MetersPerSecond] = TypeString("M_S")
```
cquiroz, line 47: What is `TypeString`

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/units.scala
rpiaggio: > What is `TypeString`

It's a very simple typeclass that specifies a serialization string for units.

PR #979, opened 2024-12-02
`UserProfile`

swalker2m:
This is a straightforward, if irksome, change to `OrcidProfile`.  It splits the user profile part (names+email) into `UserProfile` and makes `OrcidProfile` a tuple of `(OrcidId, UserProfile)`.  The reason for this is a pending update to the ODB schema that will add a "fallback" `UserProfile` to the `ProgramUser` type which can be used to hold name and email information when not publicly shared via ORCiD.


cquiroz (approved):
LGTM

PR #984, opened 2024-12-09
Add calculations to find if RA/Dec are within limits

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/model/SiteCoordinatesLimits.scala
```diff
@@ -0,0 +1,32 @@
...
+  raStart:  RightAscension,
+  raEnd:    RightAscension,
+  decStart: Declination,
+  decEnd:   Declination
+) derives Eq {
+  def inRaLimits(ra: RightAscension): Boolean =
+    if (raStart < raEnd)
+      raStart <= ra && ra <= raEnd
+    else
+      raStart <= ra || (ra.toHourAngle.toDoubleDegrees >= 0 && ra <= raEnd)
```
swalker2m, line 23: Is the `0` check necessary?  Seems like an RA should always be `[0, 24)`.

cquiroz: I took this from the phase1 model code but I was wondering the same as I typed.
I can take it away

swalker2m: It seems like we should include an assertion in the `HourAngle` "constructor", but all ways to make an `HourAngle` appear to produce a result that is in range.

swalker2m (approved):
Looks great. :+1:

PR #990, opened 2024-12-18
Only provide Eq for NewType if there is no Order instance

toddburnside:
This is one small step in handling the implicit resolution changes in scala 3.6

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala
@@ -13,13 +13,15 @@ import lucuma.core.optics.refinedPrism
 import monocle.Iso
 import monocle.Prism
 
+import scala.util.NotGiven
+
 /**
-  * Usage:
-  * ```
-  * object Name extends NewType[String]
-  * type Name = Name.Type
-  * ```
-  */
+ * Usage:
+ * ```
+ * object Name extends NewType[String]
+ * type Name = Name.Type
+ * ```
+ */
 trait NewType[Wrapped]:
   opaque type Type = Wrapped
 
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala
```diff
@@ -28,20 +30,21 @@ trait NewType[Wrapped]:
...
-  given (using eq: Eq[Wrapped]): Eq[Type]                        = eq
-  given (using enc: Encoder[Wrapped]): Encoder[Type]             = enc
-  given (using dec: Decoder[Wrapped]): Decoder[Type]             = dec
-  given (using disp: Display[Wrapped]): Display[Type]            = disp
-  given (using m: Monoid[Wrapped]): Monoid[Type]                 = m
-  given (using ord: Order[Wrapped]): Order[Type]                 = ord
-  given (using ord: Ordering[Wrapped]): Ordering[Type]           = ord
+  given (using CanEqual[Wrapped, Wrapped]): CanEqual[Type, Type]               = CanEqual.derived
+  // Only provide an Eq instance if it doesn't also provide Order
+  given (using eq:   Eq[Wrapped], noOrder: NotGiven[Order[Wrapped]]): Eq[Type] = eq
```
toddburnside, line 38: This one line is the actual change. The rest is just formatting.

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala
armanbilge: This works, but the recommended way to fix this is by explicit prioritization.

https://www.scala-lang.org/2024/08/19/given-priority-change-3.7.html#explicit-prioritization-by-owner

toddburnside: Hmm. I see. It seems like a lot more work since we're just forwarding the instances. I can change if it seems best, though.

armanbilge: I just tried and it's not nice, because the `opaque type Type` doesn't work across inheritance boundaries ... this is fine 🙈

hugo-vrijswijk:
Not directly related to this PR, but shouldn't this lib be on 3.3? As a library

PR #991, opened 2024-12-18
ProgramUser

swalker2m:
I feel sheepish about this, but a pending update to the ODB would make the `ProgramUser` an object with an id that is referred to in the API.  In particular, the goal is to make it possible to add new users with all the attributes of a `ProgramUser` (fallback name, partner, gender, thesis, etc.) before there is even an actual `User` associated with it.  For example, it would then be possible to add a CoI from Canada in order to request time from Canada even if no actual `User` is ever invited.

`ProgramUser`s will be added and removed independently of linking and unlinking actual `User`s.  Invitations will refer to specific `ProgramUser`s.


modules/core/shared/src/main/scala/lucuma/core/model/ids.scala
```diff
@@ -11,4 +11,5 @@ object CallForProposals extends WithGid('c'.refined)
 object Group            extends WithGid('g'.refined)
 object Observation      extends WithGid('o'.refined)
 object Program          extends WithGid('p'.refined)
+object ProgramUser      extends WithGid('r'.refined)
```
swalker2m, line 14: Hopefully I'm in the running for longest PR description for a one-line change.

toddburnside: A quick search showed we're already using `r` for 'StandardRole` in Role.scala. Is that a problem?

tpolecat: Yeah we're running out of letters but it would be best to choose another one.

swalker2m: Alright, how do we feel about `m`?  The only other letter in `ProgramUser` we haven't used is `e`, I think, unless that is used for events or something.

toddburnside: `m` is good. `e` is used for `ExecutionEvent`. The letters are starting to be non-mnemonic.

toddburnside (approved):
LGTM

PR #1001, opened 2025-01-02
add justification field

tpolecat:
This adds a `justification` field to `ConfigurationRequest`, which will break things I imagine.


swalker2m (approved):
Looks good. :+1:

toddburnside:
This won't break explore because I created a parallel case class that has the observation ids.

PR #1002, opened 2025-01-04
Define F2 LongSlit/Imaging geometry

cquiroz:
Here is a port of the ocs geometry to gpp. It includes geometry for:

* science area for imaging and longslit
* Probe arm geometry
* Patrol field geometry

Still missing is MOS support and the candidates area. Some examples comparing OT and gpp

## LongSlit posAngle 145, offset (-60, 60), side port
f2ot1
f2gpp1

## LongSlit posAngle 145, offset (-60, 60), bottom port
f2ot4
f2gpp4

## Imaging posAngle 145, offset (-60, 60), bottom port
f2ot3
f2gpp3

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/f2/F2ScienceAreaGeometry.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/f2/F2ScienceAreaGeometry.scala
@@ -1,0 +1,113 @@
+// Copyright (c) 2016-2023 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.geom.f2
+
+import coulomb.*
+import coulomb.policy.spire.standard.given
+import coulomb.syntax.*
+import coulomb.units.accepted.*
+import coulomb.units.si.prefixes.*
+import lucuma.core.enums.F2Fpu
+import lucuma.core.enums.F2LyotWheel
+import lucuma.core.geom.*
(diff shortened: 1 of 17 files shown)


modules/core/shared/src/main/scala/lucuma/core/geom/f2/F2ScienceAreaGeometry.scala
```diff
@@ -0,0 +1,102 @@
...
+  //   https://github.com/gemini-hlsw/ocs/blob/develop/bundle/edu.gemini.pot/src/main/scala/edu/gemini/spModel/gemini/flamingos2/F2ScienceAreaGeometry.scala#L67
+  //   val width  = MOSFOVWidth * plateScale
+  //   val height = ImagingFOVSize * plateScale
+  //   val radius = height / 2.0
+  //
+  //   // The FOV is the intersection of a rectangle and a circle.
+  //   val circle = ShapeExpression.centeredEllipse(-radius.toAngle, height.toAngle)
+  //   val rectangle = ShapeExpression.rectangleAt((-width.toAngle.bisect.p, -radius.toAngle.q), (width.toAngle.p, height.toAngle.q))
+  //   circle ∩ rectangle
+  // }
```
swalker2m, line 84: Saving some fun for later?

cquiroz: Under some definition of fun. There is no MOS on the enums, maybe it is better to add it now and get the geometry working while it is fresh

I'll do that in a next PR

PR #1007, opened 2025-01-07
Add F2 dynamic config model

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/model/sequence/f2/F2DynamicConfig.scala
```diff
@@ -0,0 +1,63 @@
...
+import monocle.Lens
+
+case class F2DynamicConfig(
+  exposure:       TimeSpan,
+  disperser:      Option[F2Disperser],
+  filter:         F2Filter,
+  readMode:       F2ReadMode,
+  lyot:           F2LyotWheel,
+  fpu:            F2FpuMask,
+  readoutMode:    Option[F2ReadoutMode], // engineering
```
cquiroz, line 20: Do we need engineering params?

swalker2m (approved):
Looks great.  :+1:

PR #1014, opened 2025-01-21
Add methods to calculate the likelihood of conditions

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
```diff
@@ -0,0 +1,121 @@
...
+import scala.math.abs
+import scala.math.pow
+import scala.math.sin
+
+// Calculations of likelihood of occurrence of observing conditions
+// Taken from:
+// https://github.com/andrewwstephens/pyexplore/blob/3edd50f6c41509752cda6ad493ccaadd5eb5ad82/test/percentile.py
+//
+// Store a percentage with two decimals of precision for a 0-100% range
+type CentiPercent    = RefinedInterval.Closed[0, 10000]
```
cquiroz, line 33: this add safety but complicates the code. I could maybe instead just use double or big decmial. Let me know what you think

modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  /**
+    * Approximate a double to a IntCentiPercent
+    */
+  def fromDouble(d: Double): IntCentiPercent = IntCentiPercent.unsafeFrom((d * 100).toInt)
+
+  /**
+    * Approximate a BigDecimal to a IntCentiPercent
+    */
+  def fromBigDecimal(d: BigDecimal): IntCentiPercent = IntCentiPercent.unsafeFrom((d * 100).toInt)
```
swalker2m, line 46: Maybe a `Format[BigDecimal, IntCentiPercent]` would be useful.

modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
```diff
@@ -0,0 +1,121 @@
...
+import scala.math.pow
+import scala.math.sin
+
+// Calculations of likelihood of occurrence of observing conditions
+// Taken from:
+// https://github.com/andrewwstephens/pyexplore/blob/3edd50f6c41509752cda6ad493ccaadd5eb5ad82/test/percentile.py
+//
+// Store a percentage with two decimals of precision for a 0-100% range
+type CentiPercent    = RefinedInterval.Closed[0, 10000]
+object IntCentiPercent extends RefinedNewType[Int, CentiPercent]:
```
swalker2m, line 34: Seems like it could go with `IntPercent` in `lucuma.core.model`?

modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
```diff
@@ -0,0 +1,121 @@
...
+import lucuma.core.math.erf
+import lucuma.core.util.RefinedNewType
+import lucuma.refined.*
+
+import scala.math.Pi
+import scala.math.abs
+import scala.math.pow
+import scala.math.sin
+
+// Calculations of likelihood of occurrence of observing conditions
```
swalker2m, line 28: Likelihood over a semester?  On a random night?

modules/core/shared/src/main/scala/lucuma/core/math/package.scala
```diff
@@ -40,3 +40,29 @@ type LineFluxValue = LineFluxValue.Type
 type FluxDensityContinuumValueRefinement = Interval.Closed[0, 1]
 object FluxDensityContinuumValue extends RefinedNewType[BigDecimal, FluxDensityContinuumValueRefinement]
 type FluxDensityContinuumValue = FluxDensityContinuumValue.Type
+
+// Copied from:
+// https://github.com/ghewgill/picomath/blob/master/scala/src/Erf.scala
+object Erf:
```
swalker2m, line 46: `Erf` is new to me.

modules/tests/shared/src/test/scala/lucuma/core/conditions/ConditionsLikelihoodSuite.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  test("minimum airmass"):
+    assertEqualsDouble(minimumAirmass(Declination.fromDoubleDegrees(20).get, Site.GN).value.toDouble, 1.000003, 0.000001)
+    assertEqualsDouble(minimumAirmass(Declination.fromDoubleDegrees(80).get, Site.GN).value.toDouble, 2.003746, 0.000001)
+    assertEqualsDouble(minimumAirmass(Declination.fromDoubleDegrees(-30).get, Site.GS).value.toDouble, 1.000006, 0.000001)
+    assertEqualsDouble(minimumAirmass(Declination.fromDoubleDegrees(-90).get, Site.GS).value.toDouble, 1.978889, 0.000001)
+
+  test("percentile sky background"):
+    assertEquals(percentileSkyBackground(SkyBackground.Bright).toPercent, 100.0)
+    assertEquals(percentileSkyBackground(SkyBackground.Dark).toPercent, 50.0)
```
swalker2m, line 29: Maybe there should be some `Int` syntax to create `CentiPercent`s?  I was wondering why these weren't centi-percents but guess it's probably because they are more awkward to create.

swalker2m (approved):
I think this looks good. :+1:

modules/tests/shared/src/test/scala/lucuma/core/conditions/ConditionsLikelihoodSuite.scala
cquiroz: Right, this matches user expectation I think, and the original python code

modules/core/shared/src/main/scala/lucuma/core/math/package.scala
cquiroz: Me too, luckily somebody had done it before

modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
cquiroz: In many cases it is just a condition has attached a likelihood, however for image quality it also depends on the observing wavelength and airmass, which in turns depends on when and where you're observing

modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
cquiroz: Right I could move it there

PR #1019, opened 2025-01-26
Fix edge case

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
```diff
@@ -85,8 +85,8 @@ def percentileImageQuality(fwhm: Quantity[BigDecimal, ArcSecond], wavelength: Wa
 
   // model fit to QAP from 2004-2024:  (the extra +0.5 is to force 100% in the worst IQ)
   val c = Array(50.10221383, 0.87712202, 0.78467697, 16.10928544, 0.13778389, -15.8255612, 49.37405633 + 0.5)
-  // The equation gives a number between 0 and 100
-  IntCentiPercent.fromBigDecimal.getOption(c(0) * erf(c(1) * pow(wavelength.toMicrometers.value.value.toDouble, c(2)) + c(3) * pow(zenithFwhm, c(4)) + c(5)) + c(6)).get
+  // The equation should give a number between 0 and 100 but on ocassions it gives a number slightly below 1
```
swalker2m, line 88: You mean slightly below 0?  This is rounding error I suppose?  Why do we assume `Min`?  That is, should it check whether the value is over 100.0 and use `Max` in that case?

cquiroz: Yeah it is a rounding error. I guess we should check the max case. I'll update it

swalker2m (approved):
> The equation should give a number between 0 and 100 but on ocassions it gives a number slightly below or obove 1

I don't follow the comment, but the change looks good.  :+1:

PR #1022, opened 2025-01-29
Make F2 Fpu tag match Gmos format

cquiroz:


swalker2m (approved):
This will require an update to the ODB schema and smart gcal tables, I think.  :+1:

PR #1024, opened 2025-02-04
fix div by zero

tpolecat:
This fixes `averageParallacticAngle` to work correctly when the observation has zero duration.


toddburnside (approved):
LGTM

PR #1029, opened 2025-02-28
Shortcut 4779: Filter obsolete acquisition filters

swalker2m:
I guess we need `Resource` to do this correctly, but in the meantime ....


cquiroz (approved):
LGTM

cquiroz:
WDYM by `Resource` in this case?

swalker2m:
> WDYM by `Resource` in this case?

I just mean we won't always be using `obsolete` flags in enums in core.  That is, `Resource` as a concept (probably just part of the ODB service).

PR #1032, opened 2025-03-01
ITC now supports more SEDs

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/enums/StellarLibrarySpectrum.scala
```diff
@@ -5,118 +5,173 @@ package lucuma.core.enums
...
-      M6V,
-      M6III,
-      M9III
-    ).withTag(_.tag)
-
-}
+enum StellarLibrarySpectrum(val tag: String, val label: String, val sedSpectrum: String) derives Enumerated:
+  /** Known library stars. */
+  // O
+  case O5V        extends StellarLibrarySpectrum("O5V", "O5V (300nm-6.0um)", "O5V")
```
cquiroz, line 11: @swalker2m Will this change require an update on enums on the odb

rpiaggio (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/StellarLibrarySpectrum.scala
rpiaggio: > @swalker2m Will this change require an update on enums on the odb

I *think* this is stored as a Json in the ODB, so it shouldn't make a huge difference there. It will need this version of core in order to be able to serialize.

cquiroz: good, the exsiting tags haven't changed so it should be able to deserialize existing data 🤞

swalker2m: They do need to be added to the schema though, right?

https://github.com/gemini-hlsw/lucuma-odb/blob/main/modules/schema/src/main/resources/lucuma/odb/graphql/OdbSchema.graphql#L10716

cquiroz: Right, I'll add them

PR #1033, opened 2025-03-01
Add CustomSedAttachment

rpiaggio:
This allows us to store a reference to an attachment in the ODB and model it in Explore. ITC will ultimately receive this and resolve it by invoking the ODB.

We want to keep the case with explicit values since we want to support 3rd party invocations to the ITC with custom SEDs.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/UnnormalizedSED.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/UnnormalizedSED.scala
@@ -149,18 +149,18 @@ object UnnormalizedSED {
 
   given Eq[UnnormalizedSED] =
     Eq.instance {
-      case (a @ StellarLibrary(_), b @ StellarLibrary(_))   => a === b
-      case (a @ CoolStarModel(_), b @ CoolStarModel(_))     => a === b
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/UnnormalizedSED.scala
```diff
@@ -148,6 +159,7 @@ object UnnormalizedSED {
       case (a @ PowerLaw(_), b @ PowerLaw(_))               => a === b
       case (a @ BlackBody(_), b @ BlackBody(_))             => a === b
       case (a @ UserDefined(_), b @ UserDefined(_))         => a === b
+      case (a @ UserDefinedAttachment(_), b @ UserDefinedAttachment(_)) => a === b
```
cquiroz, line 162: missalignment?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/UnnormalizedSED.scala
rpiaggio: forgot we don't use scalafmt here

PR #1035, opened 2025-03-03
Improve screaming case to handle dots and dashes

cquiroz:
Some of the SEDs contain dots and dashes or their tags


swalker2m (approved):
Seems fine, as long as we never have two tags that differ only in `.` vs `-` vs `_`.

PR #1036, opened 2025-03-03
Add dateToLocalObservingNight JS export

hugo-vrijswijk:
afbeelding

### DIFF
--- a/modules/npm/src/main/typescript/main.d.ts
+++ b/modules/npm/src/main/typescript/main.d.ts
@@ -2,3 +2,5 @@ export declare function deg2hms(deg: number): string;
 export declare function deg2dms(deg: number): string;
 export declare function hms2deg(hms: string): number;
 export declare function dms2deg(dms: string): number;
+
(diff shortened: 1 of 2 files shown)


rpiaggio (approved):
LGTM

modules/npm/src/main/typescript/main.d.ts
```diff
@@ -2,3 +2,5 @@ export declare function deg2hms(deg: number): string;
 export declare function deg2dms(deg: number): string;
 export declare function hms2deg(hms: string): number;
 export declare function dms2deg(dms: string): number;
+
+export declare function dateToLocalObservingNight(date: Date): string;
```
cquiroz, line 6: Nice, we are sharing code!

cquiroz (approved):
LGTM

modules/npm/src/main/typescript/main.d.ts
hugo-vrijswijk: Yes, I think this is the most KB's anyone has ever added to check if it is before or after 14:00!

PR #1037, opened 2025-03-03
Shortcut 4428: Remove (most) obsolete flags

swalker2m:
[Shortcut 4428](https://app.shortcut.com/lucuma/story/4428/remove-gmos-n-r150-g5306-from-api?vc_group_by=week&ct_workflow=all&cf_workflow=500000071) requests the removal from the API of a specific obsolete GMOS North grating.  I think by extension it is requesting the removal of all obsolete items.  So far, I've just removed GMOS and GCal obsolete items since we only support GMOS Longslit at the moment.

We can do this now because there are no old observations that use any of the obsolete items and we've been absolved of the need to import old programs.

The "obsolete" flag was a very poor man's "Resource" workaround in `ocs`.  In GPP this is tied up with dynamic enums and needs to be resolved one way or another once we get the initial public release out.  For now though, it isn't serving any purpose?  We were doing essentially nothing with it.

Thoughts?

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 import org.scalajs.linker.interface.ESVersion
 
-ThisBuild / tlBaseVersion                         := "0.116"
+ThisBuild / tlBaseVersion                         := "0.117"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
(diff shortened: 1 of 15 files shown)


swalker2m:
Obviously this doesn't update the ODB GraphQL schema to match.  I'd do that in a followup.

swalker2m:
I suppose I'd need to check for any unfortunate usages in a database migration and, um, delete them or substitute them for valid items.

toddburnside (approved):
Seems OK to me if resource is going to be used to determine availability. 

I will note that currently Explore uses the obsolete flags on at least Gmos*Filter, Gmos*Grating and GmosRoi to filter out items from the dropdown. So, explore will need updated for this.

PR #1046, opened 2025-03-11
Add site to instrument and remove obsolete instruments

cquiroz:
Removed Bhros, Michell, Nici, Phoneix, Trecs
We have also `AcqCamGS`, `AcqCamGN` and `VisitorGS`, `VisitorGN`


modules/core/shared/src/main/scala/lucuma/core/enums/Instrument.scala
```diff
@@ -13,24 +13,21 @@ import lucuma.refined.*
...
-  /** @group Constructors */ case Niri       extends Instrument("Niri", "NIRI", "NIRI", "NIRI".refined[NonEmpty], false)
-  /** @group Constructors */ case Phoenix    extends Instrument("Phoenix", "Phoenix", "Phoenix", "PHOENIX".refined[NonEmpty], false)
-  /** @group Constructors */ case Trecs      extends Instrument("Trecs", "TReCS", "TReCS", "TRECS".refined[NonEmpty], false)
-  /** @group Constructors */ case Visitor    extends Instrument("Visitor", "Visitor Instrument", "Visitor Instrument", "VISITOR".refined[NonEmpty], false)
-  /** @group Constructors */ case Scorpio    extends Instrument("Scorpio", "SCORPIO", "Scorpio", "SCORPIO".refined[NonEmpty], false)
-  /** @group Constructors */ case Alopeke    extends Instrument("Alopeke", "ALOPEKE", "Alopeke", "ALOPEKE".refined[NonEmpty], false)
-  /** @group Constructors */ case Zorro      extends Instrument("Zorro", "ZORRO", "Zorro", "ZORRO".refined[NonEmpty], false)
+enum Instrument(val tag: String, val shortName: String, val longName: String, val referenceName: NonEmptyString, val site: Site, val obsolete: Boolean) derives Enumerated:
```
swalker2m, line 16: Let's remove the `obsolete` flag while we're at it?

swalker2m: Hmm, I don't know.  I was arguing for removing the `obsolete` concept and working on `Resource`.  Come to think of it, instruments have changed site before so maybe the `site` doesn't belong there after all. 🤔

cquiroz: True, GPI is moving north

swalker2m: And wasn't GNIRS at GS before?  Site is maybe something that Resource should track.

cquiroz:
Let's resource handle this

PR #1047, opened 2025-03-11
Remove some obsolete f2 filters

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/enums/F2Filter.scala
```diff
@@ -11,25 +11,15 @@ import lucuma.core.util.Enumerated
...
-  val tag: String,
-  val shortName: String,
-  val longName: String,
-  val wavelength: Option[Wavelength],
-  val obsolete: Boolean
-) derives Enumerated:
+enum F2Filter(val tag: String, val shortName: String, val longName: String, val wavelength: Option[Wavelength], val obsolete: Boolean) derives Enumerated:
 
   case Y      extends F2Filter("Y",      "Y",       "Y (1.02 um)",        Wavelength.fromIntPicometers(1020000), false)
-  case F1056  extends F2Filter("F1056",  "F1056",   "F1056 (1.056 um)",   Wavelength.fromIntPicometers(1056000), false)
```
cquiroz, line 23: These are the important changes, the rest is fluff

swalker2m: I got rid of the `obsolete` flags when I removed the obsolete GMOS stuff on the theory that it isn't sufficient to manage resources with a boolean flag.  Maybe we need it for something though?

modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
```diff
@@ -90,6 +88,9 @@ object Enumerated {
...
 
-  extension [A](a: A)(using ev: Enumerated[A])
-    def tag: String = ev.tag(a)
-}
\ No newline at end of file
+
+/** @group Typeclasses */
+trait Obsoletable[A]:
+  def isActive(a: A): Boolean
+  final def isObsolete(a: A): Boolean = !isActive(a)
```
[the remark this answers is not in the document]
cquiroz: Right, I restored some bits that I deleted about syntax and this got back along. I'll remove it

PR #1053, opened 2025-03-17
[Shortcut 4973] Add `DeclaredComplete` execution state

swalker2m:
Adds a new `ExecutionState` element, `DeclaredComplete`.  It is employed when a user has determined that the observation should be considered complete even though there may be remaining steps prescribed by the generator.


tpolecat (approved):
bueno

PR #1055, opened 2025-03-18
Shortcut 4994: remove partner time

swalker2m:
The concept of "partner time" from `ocs` is [being removed in GPP](https://app.shortcut.com/lucuma/story/4994/remove-partner-time?vc_group_by=week&ct_workflow=all&cf_workflow=500000071).  That implies the removal of the `PartnerCal` and `AcquisitionCal` observe classes which existed to distinguish the charge class from `Program` time.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 import org.scalajs.linker.interface.ESVersion
 
-ThisBuild / tlBaseVersion                         := "0.118"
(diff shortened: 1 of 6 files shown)


tpolecat (approved):
I'm amazed that time accounting is getting simpler, but if they say so … 👍🏻

PR #1056, opened 2025-03-22
Merge Catalog

cquiroz:
We are spending too much time doing releases. I propose we merge some of the repos

In this PR I'm bringing catalog and ags to core

The code is exactly as in the catalog repo. Except for the large version number change, it shouldn't produce any problems downstream.

Code formatting is off for the old modules.


rpiaggio (approved):
LGTM

toddburnside (approved):
LGTM. Thanks for taking the time to do this.

PR #1062, opened 2025-03-25
Add some more `InputValidSplitEpi` instances

toddburnside:
A couple of these were already in explore with `// TODO: Move to core`. I'll remove those from explore in my next explore PR.


cquiroz (approved):
LGTM

PR #1063, opened 2025-03-25
Remove Bhros, Michelle, Nici, Nifs, Phoenix, Trecs

toddburnside:
Carlos says all of these instruments can be removed. I thought I'd do it now since we already have a breaking change with the removal of the Bhros FPU.


cquiroz (approved):
LGTM

swalker2m (approved):
This seems in keeping with removing deprecated items and the decision to not support importing old programs.  :+1:

PR #1068, opened 2025-04-01
Migrate ImageQuality and CloudExtinction to values

rpiaggio:
Turns out that `ImageQuality` and `CloudExtinction` are actually arbitrary values in some cases, like when we refer to the actual current conditions.

This PR remodels the `enums` we had for selecting certain values for constraints and configuration. These `enums` are still kept and they wrap specific values.

Although a breaking change, I already have all the downstream changes lined up and tested.

No serialized instances were harmed in the production of this PR.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
@@ -54,15 +54,15 @@ def percentileSkyBackground(bg: SkyBackground): IntCentiPercent =
 /**
   * Return the percentile of a measured extinction.
   */
-def percentileCloudCoverage(extinction: CloudExtinction.Point): IntCentiPercent =
+def percentileCloudCoverage(extinction: CloudExtinction.Preset): IntCentiPercent =
   extinction match
-    case CloudExtinction.Point.PointOne       => IntCentiPercent(5000.refined)
-    case CloudExtinction.Point.PointThree     => IntCentiPercent(7000.refined)
-    case CloudExtinction.Point.PointFive      => IntCentiPercent(7500.refined)
-    case CloudExtinction.Point.OnePointZero   => IntCentiPercent(8000.refined)
-    case CloudExtinction.Point.OnePointFive   => IntCentiPercent(9000.refined)
-    case CloudExtinction.Point.TwoPointZero   => IntCentiPercent(9500.refined)
-    case CloudExtinction.Point.ThreePointZero => IntCentiPercent(10000.refined)
+    case CloudExtinction.Preset.PointOne       => IntCentiPercent(5000.refined)
+    case CloudExtinction.Preset.PointThree     => IntCentiPercent(7000.refined)
+    case CloudExtinction.Preset.PointFive      => IntCentiPercent(7500.refined)
+    case CloudExtinction.Preset.OnePointZero   => IntCentiPercent(8000.refined)
+    case CloudExtinction.Preset.OnePointFive   => IntCentiPercent(9000.refined)
+    case CloudExtinction.Preset.TwoPointZero   => IntCentiPercent(9500.refined)
+    case CloudExtinction.Preset.ThreePointZero => IntCentiPercent(10000.refined)
 
 /**
   * Return the percentile of the water vapor.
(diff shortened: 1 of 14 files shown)


rpiaggio:
I hadn't noticed there is already an `Extinction` type. I'll think a bit how to merge both things.

cquiroz:
> No serialized instances were harmed in the production of this PR.
🤣

cquiroz:
In general, it seems we'll allow any value for IQ and CE (within restrictions) and have an enum for the most used values too.

I can see this will cascade everywhere, but you mentioned that you had updated all the projects downstream. 
I assume this change was requested by science.

How would it affect the odb? Are we storing enums there or numbers?

swalker2m:
> How would it affect the odb? Are we storing enums there or numbers?

Enums.  The GraphQL Schema will need to be updated along with the database schema.

rpiaggio:
In the ODB these are only stored as part of `ConstraintSet`s, where we want to store the enum values. I haven't changed the names of the enums exposed by the schema, but maybe it's a good idea to do so.

rpiaggio:
I have reworked `Extinction` to be a `NewType` and `CloudExtinction` is a refinement.

For this, I created implicit `Validator`s that allow refining any `Refined`, `Quantity` or `NewType`. These may be useful in other places.

modules/core/shared/src/main/scala/lucuma/core/model/ConstraintSet.scala
```diff
@@ -17,8 +15,8 @@ import monocle.Optional
 
 /** Constraints for an observation. */
 case class ConstraintSet(
-  imageQuality:    ImageQuality,
-  cloudExtinction: CloudExtinction,
+  imageQuality:    ImageQuality.Point,
+  cloudExtinction: CloudExtinction.Point,
```
swalker2m, line 19: I wonder if something like `Constraint` would make sense if we only plan on using the enums for the constraint set. Then it would read `ImageQuality.Constraint.PointOne` instead of `ImageQuality.Point.PointOne`.

swalker2m (approved):
This looks great to me. :+1:

modules/core/shared/src/main/scala/lucuma/core/model/ConstraintSet.scala
rpiaggio: I've struggled with this name and I couldn't find anything that conveys the meaning of a "singled-out instance".

`Constraint` seems to describe more our usage than the general "singled-out instance".

Other names I've considered, in case you think any of them are a better fit: `Stop`, `Instance`, `Mark`, `Peg`, `Hint`. IMHO, none of those looked too good in an API.

swalker2m: `Preset` maybe?  Idk, it's fine however if you want to leave it as is.

rpiaggio: I like `Preset`, let's go with that one.

modules/core/shared/src/main/scala/lucuma/core/conditions/package.scala
```diff
@@ -53,17 +53,16 @@ def percentileSkyBackground(bg: SkyBackground): IntCentiPercent =
...
+def percentileCloudCoverage(extinction: CloudExtinction.Preset): IntCentiPercent =
   extinction match
-    case CloudExtinction.PointOne       => IntCentiPercent(5000.refined)
-    case CloudExtinction.PointThree     => IntCentiPercent(7000.refined)
-    case CloudExtinction.PointFive      => IntCentiPercent(7500.refined)
-    case CloudExtinction.OnePointZero   => IntCentiPercent(8000.refined)
-    case CloudExtinction.OnePointFive   => IntCentiPercent(9000.refined)
-    case CloudExtinction.TwoPointZero   => IntCentiPercent(9500.refined)
-    case CloudExtinction.ThreePointZero => IntCentiPercent(10000.refined)
+    case CloudExtinction.Preset.PointOne       => IntCentiPercent(5000.refined)
```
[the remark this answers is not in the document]
rpiaggio: That's a good question. The same applies for the extensions for `WaterVapor` and `SkyBackground` here.

Let's ask the author: @cquiroz ?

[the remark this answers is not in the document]
cquiroz: Ah do you mean the method?, no reason, we should probably put it in the enum

PR #1072, opened 2025-04-08
Provide method for finding anglesToTestAt with known APA

toddburnside:
The existing `anglesToTestAt` always calculated the average parallactic angle for `AverageParallacticAngle` constraints. However, in some cases, the APA is already known.

### DIFF
--- a/modules/ags/src/main/scala/lucuma/ags/package.scala
+++ b/modules/ags/src/main/scala/lucuma/ags/package.scala
@@ -151,15 +151,21 @@ extension (posAngleConstraint: PosAngleConstraint)
     vizTime:  Instant,
     duration: Duration
   ): Option[NonEmptyList[Angle]] =
-    TimeSpan
-      .fromDuration(duration)
-      .filter(_ > TimeSpan.Zero)
-      .flatMap: ts =>
-        posAngleConstraint match
-          case PosAngleConstraint.Fixed(a)               => NonEmptyList.of(a).some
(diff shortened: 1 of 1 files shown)


modules/ags/src/main/scala/lucuma/ags/package.scala
```diff
@@ -151,15 +151,21 @@ extension (posAngleConstraint: PosAngleConstraint)
...
-        posAngleConstraint match
-          case PosAngleConstraint.Fixed(a)               => NonEmptyList.of(a).some
-          case PosAngleConstraint.AllowFlip(a)           => NonEmptyList.of(a, a.flip).some
-          case PosAngleConstraint.ParallacticOverride(a) => NonEmptyList.of(a).some
-          case PosAngleConstraint.AverageParallactic     =>
-            averageParallacticAngle(site.place, tracking, vizTime, ts)
-              .map(a => NonEmptyList.of(a, a.flip))
-          case PosAngleConstraint.Unbounded              => UnconstrainedAngles
+    anglesToTestAt(
+      TimeSpan
```
toddburnside, line 155: The vizTime is no longer validated unless we need to calculate the APA. I don't think this is a problem.

toddburnside: We were making sure it was a valid TimeSpan and not Zero at the beginning of the method before. Now we only do it if we need it.

cquiroz: It seems we validate `duration` but not `vizTime`

cquiroz (approved):
LGTM

PR #1080, opened 2025-04-12
Add toSentenceCase

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/syntax/String.scala
```diff
@@ -60,6 +60,13 @@ trait ToStringOps {
     def toKebabCase: String =
       kebabCaseTransformation(self)
 
+    /**
+    * Converts the `String` to "sentence case" (eg "Foo-bar").
+    */
+    def toSentenceCase: String =
+      if (self.length == 0) self
+      else if (self.length == 1) self.toUpperCase
+      else s"${self.charAt(0).toUpper}${self.substring(1)}"
```
swalker2m, line 69: Is this the same as `capitalize`?
```
@ "hello world".capitalize 
res0: String = "Hello world"
```

cquiroz: I didn't know about it, thxs

PR #1088, opened 2025-04-23
initial metadata

tpolecat:
This adds a `Metadata` structure that will contain lookup tables and other dynamic but slowly-changing information from the ODB, which will eventually include some things that are now `enum`s, but for now it only contains instrument availability. The intent is that this information will be loaded at startup and passed around implicitly, but for now there is just a hardcoded instance containing instance availability (also hardcoded). This should be passed to applications from `main` and used implicitly thereafter, to minimize impact when the hardcoded instance goes away.

~~A utility class introduced here the functor `PerSite[+A](gn: A, gs: A)` which holds a pair of values, one per site.~~

~~The end result for this initial PR is that `given Metadata` you can say `Instrument.availability` and get back a `PerSite[Diet[Instant]]`; i.e., a time interval set (a **d**iscrete **i**nterval **e**ncoding **t**ree) for each site indicating when the instrument is available.~~

This adds the following utility classes:

- `DisjointIntervalMap[K,V]` which maps keys `K` to interval sets `Diet[V]` where the interval sets are disjoint and non-empty. So if the interval set for key `k` contains a value `v` then `v` will not exist in the interval set for any other key.
- `Availability`, which wraps a `DisjointIntervalMap[Site, Instant]`, representing the time intervals where some resource is available at each site, but never at both sites at the same time. This handles the case where an instrument moves from site to site. There are many accessors on `Availability` that allow selection by `DateInterval`, `Semester`, etc.

The end result for this initial PR is that `given Metadata` you can say `Instrument.availability` and get back an `Availability` that tells you where and when the instrument is available.

---

The `PerSite` data type introduced in an earlier version remains, in case it might be useful elsewhere.


modules/core/shared/src/main/scala/lucuma/core/instances/TimeInstances.scala
```diff
@@ -0,0 +1,13 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.instances
+
+import cats.kernel.Order
+
+import java.time.Instant
+
+trait TimeInstances:
+  given Order[Instant] = Order.by(_.toEpochMilli)
```
rpiaggio, line 11: This is in `cats-time`, which we already import in `core`.

modules/core/shared/src/main/scala/lucuma/core/data/Metadata.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import java.time.Instant 
+
+/** 
+ * A bundle of lookup tables and similar slow-changing but dynamic information
+ * from the ODB, intended to be loaded at application startup and passed 
+ * around implicitly. Rather than using this instance directly, methods on 
+ * data classes can provide better-placed methods that demand `Metadata`.
+ */
+trait Metadata:
```
rpiaggio, line 20: This will travel around, from ODB to UIs and servers. Can we add a `Codec` here? Or is a 2nd PR coming with the endpoint an the codec?

modules/core/shared/src/main/scala/lucuma/core/data/PerSite.scala
```diff
@@ -0,0 +1,89 @@
...
+import cats.kernel.Monoid
+import cats.syntax.monoid.*
+import io.circe.Decoder
+import io.circe.Encoder
+import io.circe.Json
+import io.circe.syntax.*
+import lucuma.core.enums.Site
+
+/** A total map from Site to A. */
+final case class PerSite[+A](gn: A, gs: A) extends (Site => A):
```
rpiaggio, line 17: Aren't `case class`es `final` by default in Scala 3?

modules/core/shared/src/main/scala/lucuma/core/data/PerSite.scala
```diff
@@ -0,0 +1,89 @@
...
+    def empty: PerSite[A] = PerSite(M.empty, M.empty)
+    def combine(x: PerSite[A], y: PerSite[A]): PerSite[A] =
+      PerSite(x.gn |+| y.gn, x.gs |+| y.gs)
+
+  /** PerSite is Eq if A is Eq. */
+  given [A: Eq]: Eq[PerSite[A]] =
+    Eq.by(_.toPair)
+
+  /** PerSite is encodable to Json if A is encodable. */      
+  implicit def encoderPerSite[A: Encoder]: Encoder[PerSite[A]] = ps =>
```
rpiaggio, line 75: `given` ?

modules/core/shared/src/main/scala/lucuma/core/data/Metadata.scala
```diff
@@ -0,0 +1,62 @@
...
+                 Instrument.Ghost      |
+                 Instrument.Gsaoi      |
+                 Instrument.Zorro      => PerSite(never, always)
+
+            // both
+            case Instrument.AcqCam     |
+                 Instrument.Visitor    => PerSite(always, always)
+
+            // future
+            case Instrument.Gpi        |
```
cquiroz, line 57: I'd put Gpi at GN already

modules/core/shared/src/main/scala/lucuma/core/data/PerSite.scala
```diff
@@ -0,0 +1,89 @@
...
+    site match
+      case Site.GN => gn
+      case Site.GS => gs    
+
+  /** PerSite is a functor. */
+  def map[B](f: A => B): PerSite[B] =
+    PerSite(f(gn), f(gs))
+
+  /** Examine each (Site, A) pair and fold into monoid B. */
+  def fold[B: Monoid](f: (Site, A) => B): B =
```
cquiroz, line 38: Maybe a `fold(gn: A => B, gs: A => B)` would be useful.

modules/core/shared/src/main/scala/lucuma/core/data/Metadata.scala
tpolecat: Yeah the wire structure will be determined by the schema so I'll put this on the ODB side in a second PR

modules/core/shared/src/main/scala/lucuma/core/data/PerSite.scala
tpolecat: No, Martin has some cases where he likes to extend them.

```scala
scala> case class Foo()
// defined case class Foo
                                                                                
scala> class Bar extends Foo
// defined class Bar
```

modules/core/shared/src/main/scala/lucuma/core/data/PerSite.scala
tpolecat: I thought about that signature but it's kind of like Boolean.fold ... there's nothing to prevent you from passing the args in the wrong order. I'll add it and change this one to `foldWithSite` or something.

swalker2m (dismissed):
I'm not clear on how `Metadata` becomes available everywhere it will be needed, but this looks great.  :+1:  As mentioned, a major goal of the instrument / site association is to know where an observation will happen.  As it stands this implementation wouldn't suffice for that purpose since an instrument could be at multiple sites simultaneously so we'll still need a solution for that.

modules/core/shared/src/main/scala/lucuma/core/enums/Instrument.scala
```diff
@@ -28,3 +35,15 @@ enum Instrument(val tag: String, val shortName: String, val longName: String, va
   /** @group Constructors */ case Scorpio    extends Instrument("Scorpio", "SCORPIO", "Scorpio", "SCORPIO".refined[NonEmpty], false)
   /** @group Constructors */ case Alopeke    extends Instrument("Alopeke", "ALOPEKE", "Alopeke", "ALOPEKE".refined[NonEmpty], false)
   /** @group Constructors */ case Zorro      extends Instrument("Zorro", "ZORRO", "Zorro", "ZORRO".refined[NonEmpty], false)
+
+  def availability(using md: Metadata): PerSite[Diet[Instant]] =
+    md.instrumentAvailability.getOrElse(this, PerSite.unfold(_ => Diet.empty))
+
+  def availabilityAt(when: Instant)(using Metadata): Set[Site] =
+    availability.filter(_.contains(when))
```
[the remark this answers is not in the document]
tpolecat: Yeah it seems like what we want is a struct where the interval sets always have to be disjoint, so you'll never get "both" as an answer. I will think about a good way to do that.

modules/core/shared/src/main/scala/lucuma/core/data/Availability.scala
```diff
@@ -0,0 +1,91 @@
...
+  def forDateInterval(dateInterval: DateInterval): Availability =
+    Availability:
+      DisjointIntervalMap.unsafeFromMap:
+        Site.all.fproduct(forSiteAndDateInterval(_, dateInterval)).toMap  
+  
+  /** The interval set for the given site and DateInterval. */
+  def forSiteAndDateInterval(site: Site, dateInterval: DateInterval): Diet[Instant] =
+    val start = dateInterval.start.atStartOfDay(site.timezone).toInstant() 
+    val end   = dateInterval.end.atStartOfDay(site.timezone).plusDays(1).toInstant()
+    forSiteAndRange(site, Range(start, end))
```
swalker2m, line 49: Since day boundaries occur in the middle of the night, I would if a `forObservingNightInterval` (`ObservingNight` includes a `Site`) would be useful at some point.

modules/core/shared/src/main/scala/lucuma/core/data/DisjointIntervalMap.scala
```diff
@@ -0,0 +1,146 @@
...
+import cats.collections.Discrete
+import cats.collections.Range
+import cats.kernel.CommutativeSemigroup
+import cats.kernel.Eq
+import cats.kernel.LowerBounded
+import cats.kernel.Order
+import cats.kernel.UpperBounded
+
+/** A map from keys to interval sets, where the interval sets are disjoint and non-empty. */
+sealed trait DisjointIntervalMap[K, V: Discrete: Order]:
```
swalker2m, line 16: This seems like it could maybe replace or be used as basis of `TimeAccountingState` in the ODB. It's a map keyed by `TimestampInterval`.

swalker2m (approved):
This looks outstanding.  Thanks for taking it on.  :+1:

PR #1089, opened 2025-04-23
Add units to F2 enums

cquiroz:
I'm also removing obsolete items

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/syntax/units.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/syntax/units.scala
@@ -1,0 +1,23 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math.syntax
+
+import coulomb.Quantity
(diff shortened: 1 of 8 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/syntax/units.scala
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.math.syntax
```
rpiaggio, line 4: I think we are concentranting all syntax extensions in `lucuma.core.syntax`.

rpiaggio (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/math/syntax/units.scala
cquiroz: We have e.g.
`import lucuma.core.math.syntax.int.*`

cquiroz: an exercise for a newcomer

PR #1090, opened 2025-04-24
F2 geom

cquiroz:


rpiaggio (approved):
LGTM, I'm assuming the logic is correct

PR #1091, opened 2025-04-25
Remove SlewEvent

cquiroz:


swalker2m (approved):
I think this will be fine but you might want to try updating `lucuma-odb` before committing it.  I'm not 100% sure whether there will be gotchas.

cquiroz:
I attempted to implement this on odb, However, it's a lot more involved, requiring special mappings. 
We will punt on this change as it's not very critical

PR #1099, opened 2025-05-02
CalculatedValue

swalker2m:
Adds a calculation state for background calculations and a generic calculated value to hold the state plus last known value.  This will be used by the background calculation update in the ODB.


modules/core/shared/src/main/scala/lucuma/core/util/CalculatedValue.scala
```diff
@@ -0,0 +1,77 @@
...
+  /**
+   * An empty, or zero, 'Ready' calculated value.
+   */
+  def empty[A](using Monoid[A]): CalculatedValue[A] =
+    CalculatedValue(CalculationState.Zero, Monoid[A].empty)
+
+  given [A: Eq]: Eq[CalculatedValue[A]] =
+    Eq.by(cv => (cv.state, cv.value))
+
+  given [A](using Monoid[A]): Monoid[CalculatedValue[A]] =
```
swalker2m, line 45: This could be a `CommutativeMonoid` if there is a `CommutativeMonoid` for `A`.  I didn't want to limit it to `CommutativeMonoid`s though.  Is there a better way to define this?

toddburnside: Is this the problem of defining both a Monoid and a CommunativeMonoid causing an error? You could only define the Monoid if A doesn't provide a CommunativeMonoid, like here https://github.com/gemini-hlsw/lucuma-core/blob/9fab14aeeba5448514d03483e84d66919b6c9737/modules/core/shared/src/main/scala/lucuma/core/util/NewType.scala#L54

modules/tests/shared/src/test/scala/lucuma/core/util/CalculatedValueSuite.scala
```diff
@@ -0,0 +1,24 @@
...
+class CalculatedValueSuite extends DisciplineSuite:
+  checkAll("CalculatedValue.Monoid", MonoidTests[CalculatedValue[Int]].monoid)
+
+  // Without explicitly defining which version to use, we get this error:
+  //    But both object given_Monad_CalculatedValue in object CalculationValue
+  //    and object given_Traverse_CalculatedValue in object CalculationValue
+  //    match type cats.Invariant[lucuma.core.util.CalculatedValue].
+  // How do you fix this in a less clumsy way?
+  given Invariant[CalculatedValue] = Monad[CalculatedValue]
+  checkAll("CalculatedValue.Monad",  MonadTests[CalculatedValue].monad[Int, String, Long])
```
tpolecat, line 22: That's odd. I don't know who is asking for `Invariant`. But what you have here works. The alternative would be passing the correct instance explicitly, if you can figure out where it's demanded.

tpolecat (approved):
This looks great :+1:

We may need `CalculatedValueT` at some point.

modules/core/shared/src/main/scala/lucuma/core/util/CalculatedValue.scala
```diff
@@ -0,0 +1,77 @@
...
+ * underlying value is associated with a calculation state. If the state is
+ * 'Ready' then the value may be considered up-to-date at the moment it was
+ * retrieved. Otherwise the value represents the last known value before an
+ * update (possibly) invalidated it and triggered a new computation.
+ *
+ * @param state current calculation state
+ * @param value last known underlying value
+ * @tparam A data type
+ */
+final case class CalculatedValue[A](
```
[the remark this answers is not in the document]
toddburnside: It has more potential states than a `Pot`.

modules/core/shared/src/main/scala/lucuma/core/util/CalculatedValue.scala
swalker2m: Something like this?  From more specialized to more general:

```
  given [A](using CommutativeMonoid[A]): CommutativeMonoid[CalculatedValue[A]] =
    CommutativeMonoid.instance[CalculatedValue[A]](
      empty,
      (a, b) => CalculatedValue(a.state |+| b.state, a.value |+| b.value)
    )

  given [A](using Monoid[A], NotGiven[CommutativeMonoid[A]]): Monoid[CalculatedValue[A]] =
    Monoid.instance[CalculatedValue[A]](
      empty,
      (a, b) => CalculatedValue(a.state |+| b.state, a.value |+| b.value)
    )

  given [A](using Semigroup[A], NotGiven[Monoid[A]], NotGiven[CommutativeMonoid[A]]): Semigroup[CalculatedValue[A]] =
    Semigroup.instance[CalculatedValue[A]](
      (a, b) => CalculatedValue(a.state |+| b.state, a.value |+| b.value)
    )
```

modules/tests/shared/src/test/scala/lucuma/core/util/CalculatedValueSuite.scala
swalker2m: Defining it as 

```scala
given Monad[CalculatedValue] with Traverse[CalculatedValue] with
  ...
```

(instead of separately) solves the problem.  Thanks @toddburnside for the tip.

tpolecat: Aha, duh. 👍🏻

modules/core/shared/src/main/scala/lucuma/core/util/CalculatedValue.scala
swalker2m: Didn't seem to be necessary?  I just defined them all and there doesn't seem to a problem.

modules/core/shared/src/main/scala/lucuma/core/util/CalculatedValue.scala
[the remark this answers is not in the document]
rpiaggio: It also holds a value always. `Pot` only does so when `Ready`.

PR #1104, opened 2025-05-09
Shortcut 4365: Require custom mask filename

swalker2m:
Changes the F2 custom mask to require a filename.  This matches GMOS and, I think, in reality there is a mask definition file which must be identified.

This isn't needed for F2 long slit, but it is needed to fully define the F2 dynamic configuration for steps and those must be recorded as steps are executed.


cquiroz (approved):
LGTM

PR #1106, opened 2025-05-09
Shortcut 4365: Remove custom slit width from static config

swalker2m:
Custom slit width is applicable to (and indeed already defined in) the `F2FpuMask`, which is part of the dynamic config.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/f2/F2StaticConfig.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/f2/F2StaticConfig.scala
@@ -5,19 +5,17 @@ package lucuma.core.model.sequence.f2
 
 import cats.Eq
 import cats.syntax.all.*
-import lucuma.core.enums.*
 import monocle.Focus
 import monocle.Lens
 
 case class F2StaticConfig(
   mosPreImaging:          Boolean,
-  useElectronicOffseting: Boolean,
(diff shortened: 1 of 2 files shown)


swalker2m:
Also, switch the `mosPreImaging` field from `Boolean` to the `MosPreImaging` enum to be consistent with GMOS.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/f2/F2StaticConfig.scala
```diff
@@ -5,29 +5,23 @@ package lucuma.core.model.sequence.f2
...
-import lucuma.core.enums.*
+import lucuma.core.enums.MosPreImaging
 import monocle.Focus
 import monocle.Lens
 
 case class F2StaticConfig(
-  mosPreImaging:          Boolean,
-  useElectronicOffseting: Boolean,
-  customSlitWidth:        Option[F2CustomSlitWidth]
+  mosPreImaging:          MosPreImaging,
```
cquiroz, line 13: I learned this is only used only on imaging

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/sequence/f2/F2StaticConfig.scala
swalker2m: That's right, I think it is used to design a mask for MOS.

PR #1108, opened 2025-05-12
Fix filter name typo

cquiroz:


toddburnside:
> I predict a month of effort to find everything that breaks

Well, what else do we have to do, anyway? 🤣

PR #1110, opened 2025-05-15
Shortcut 4365: `Flamingos2DynamicConfig` updates

swalker2m:
This PR makes a few adjustments discovered when working with the F2 model.  Comments inline.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/Flamingos2Filter.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/Flamingos2Filter.scala
@@ -3,6 +3,7 @@
 
 package lucuma.core.enums
 
+import cats.data.NonEmptyList
 import lucuma.core.math.Wavelength
 import lucuma.core.util.Enumerated
 
@@ -23,3 +24,9 @@ enum Flamingos2Filter(val tag: String, val shortName: String, val longName: Stri
   case KShort extends Flamingos2Filter("KShort", "K-short", "K-short (2.15 um)",  2150000.pm)
   case KBlue  extends Flamingos2Filter("KBlue",  "K-blue",  "K-blue (2.06 um)",   2060000.pm)
   case KRed   extends Flamingos2Filter("KRed",   "K-red",   "K-red (2.31 um)",    2310000.pm)
+
+object Flamingos2Filter:
+
+  /** Acquisition filter options. */
+  val acquisition: NonEmptyList[Flamingos2Filter] =
+    NonEmptyList.of(J, H, KShort)
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
@@ -27,7 +27,7 @@ case class Flamingos2DynamicConfig(
   val isLongSlit: Boolean = fpu.isLongSlit
 
   val centralWavelength: Wavelength =
-    disperser.fold(filter.wavelength)(_.wavelength)
+    filter.wavelength
 
 object Flamingos2DynamicConfig:
   given Eq[Flamingos2DynamicConfig] =
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
@@ -7,28 +7,31 @@ import cats.Eq
 import cats.syntax.all.*
 import lucuma.core.enums.*
 import lucuma.core.util.TimeSpan
+import lucuma.core.math.Wavelength
 import monocle.Focus
 import monocle.Lens
 
 case class Flamingos2DynamicConfig(
-  exposure:       TimeSpan,
-  disperser:      Option[Flamingos2Disperser],
-  filter:         Flamingos2Filter,
-  readMode:       Flamingos2ReadMode,
-  lyot:           Flamingos2LyotWheel,
-  fpu:            Flamingos2FpuMask,
-  readoutMode:    Option[Flamingos2ReadoutMode], // engineering
-  reads:          Option[Flamingos2Reads], // engineering
-) {
-  val isImaging: Boolean       = fpu.isImaging
-  val isMOS: Boolean           = fpu.isMOS
-  val isLongSlit: Boolean      = fpu.isLongSlit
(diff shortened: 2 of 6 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/Flamingos2Filter.scala
```diff
@@ -23,3 +24,9 @@ enum Flamingos2Filter(val tag: String, val shortName: String, val longName: Stri
   case KShort extends Flamingos2Filter("KShort", "K-short", "K-short (2.15 um)",  2150000.pm)
   case KBlue  extends Flamingos2Filter("KBlue",  "K-blue",  "K-blue (2.06 um)",   2060000.pm)
   case KRed   extends Flamingos2Filter("KRed",   "K-red",   "K-red (2.31 um)",    2310000.pm)
+
+object Flamingos2Filter:
+
+  /** Acquisition filter options. */
+  val acquisition: NonEmptyList[Flamingos2Filter] =
+    NonEmptyList.of(J, H, KShort)
```
swalker2m, line 32: I'll need to pick the filter closest to the observation's wavelength from this list of valid acquisition filters.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
```diff
@@ -6,29 +6,32 @@ package lucuma.core.model.sequence.flamingos2
...
-  val isLongSlit: Boolean      = fpu.isLongSlit
-  val decker: Option[Flamingos2Decker] = fpu.fold(none, _.value.decker.some, _ => none)
-}
+  exposure:    TimeSpan,
+  disperser:   Option[Flamingos2Disperser],
+  filter:      Flamingos2Filter,
+  readMode:    Flamingos2ReadMode,
+  lyotWheel:   Flamingos2LyotWheel,
+  fpu:         Flamingos2FpuMask,
+  decker:      Flamingos2Decker,
```
swalker2m, line 21: `decker` appears in the F2 long slit configuration as an option, so if it is selected to be anything other than the default we'll need to keep up with it.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
```diff
@@ -6,29 +6,32 @@ package lucuma.core.model.sequence.flamingos2
...
+  readoutMode: Flamingos2ReadoutMode,
+  reads:       Flamingos2Reads
```
swalker2m, line 23: I made these non-optional here since we need some value to apply to the instrument.  In the observing mode configuration itself, they need not be required since we have logical defaults for them.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
```diff
@@ -6,29 +6,32 @@ package lucuma.core.model.sequence.flamingos2
...
+):
+  val isImaging: Boolean  = fpu.isImaging
+  val isMOS: Boolean      = fpu.isMOS
+  val isLongSlit: Boolean = fpu.isLongSlit
+
+  val centralWavelength: Wavelength =
```
toddburnside, line 29: In Explore on the configuration panel we are just using the filter wavelength for the central wavelength. (We use it for calculating the λ interval.) I used this at Andy's suggestion. I wonder if it should use the same method as here instead?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/sequence/flamingos2/Flamingos2DynamicConfig.scala
swalker2m: I took this from the `ocs` project where we have:

```java
    /**
     * Determines the observing wavelength to use for the given combination of
     * disperser and filter.  Implements the algorithm specified in SCI-0206.
     */
    public static Option<Double> getObservingWavelength(Disperser d, Filter f) {
        // When in spectroscopy mode, if the filter is OPEN, use the grism
        // wavelength.
        if ((d != Disperser.NONE) && (f == Filter.OPEN)) return d.getWavelength();

        // Otherwise, use the filter wavelength.
        return f.getWavelength();
    }
```

Note `Filter.OPEN` is obsolete.  Now we get into the difference between "central wavelength" and "observing wavelength" which admittedly baffles me.

toddburnside: Yes. There are too many wavelengths.

swalker2m: Maybe since the filter is never `OPEN` now (since it doesn't exist), I should just use the filter wavelength.  Is this actually the "observing wavelength" or the "central wavelength" or neither or both?

toddburnside: I just asked Andy what to use for the central wavelength, and he said he "would be inclined to use the filter wavelength" - or something very similar. Maybe we should tag him here? You could merge as is and update if needed.

toddburnside: Or, I could update explore if needed.

swalker2m: > Or, I could update explore if needed.

I'll switch to filter since `OPEN` doesn't exist.

andrewwstephens: I think we should be using the filter wavelength.
imaging will *always* use a filter.
F2 will *always* use a filter (imaging and spectroscopy).

cquiroz: We support n filters right? Does this dynamic config apply to only one part of a sequence with multiple filters?

swalker2m: The dynamic config is how the instrument is configured for one F2 dataset.  If there are multiple filters in the observing mode, they would be used in different steps.

PR #1112, opened 2025-05-27
Scala 3.7

cquiroz:
There is a workaround that lets us use scala 3.7.0 with gemini locales
This change, if merged, means all downstream projects have to switch to 3.7.0 too


toddburnside (approved):
LGTM. As long as 3.7.0 is stable enough and we want to update the world to it.

PR #1113, opened 2025-05-29
SC-5657 TOO targets part 1

tpolecat:
TOO Targets have a region constraint (an area of the sky) that will be used to populate an associated configuration request. When the TOO target eventually becomes a real target with tracking information, there coordinates must lie on an approved region.

Anyway, this introduces `Arc`, representing a clockwise arc; and `Region`, representing orthogonal arcs in RA and Dec.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/math/Region.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/math/Region.scala
@@ -6,23 +6,39 @@ package lucuma.core.math
 import cats.Eq
 import monocle.Focus
 import monocle.Lens
+import monocle.Optional
 
 /** A quasirectangular region on a sphere, given by ranges in RA and Dec. */
 final case class Region(raArc: Arc[RightAscension], decArc: Arc[Declination]):
-
-  def containsAll(other: Region): Boolean =
-    raArc.containsAll(other.raArc) && 
-    decArc.containsAll(other.decArc)
 
   def contains(coords: Coordinates): Boolean =
     raArc.contains(coords.ra) &&
     decArc.contains(coords.dec)
 
+  def containsAll(other: Region): Boolean =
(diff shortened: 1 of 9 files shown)


modules/core/shared/src/main/scala/lucuma/core/math/Region.scala
```diff
@@ -0,0 +1,70 @@
...
+
+package lucuma.core.math
+
+import cats.Eq
+import monocle.Focus
+import monocle.Lens
+import monocle.Optional
+
+/** A quasirectangular region on a sphere, given by ranges in RA and Dec. */
+final case class Region(raArc: Arc[RightAscension], decArc: Arc[Declination]) {
```
cquiroz, line 12: Is that how ToOs are defined? you give a region where the target may be?

modules/tests/shared/src/test/scala/lucuma/core/math/ArcSuite.scala
```diff
@@ -0,0 +1,139 @@
...
+import org.scalacheck.Cogen
+import org.scalacheck.Prop.*
+
+final class ArcSuite extends munit.DisciplineSuite:
+  import ArbAngle.given
+  import ArbArc.given
+  import ArbRightAscension.given
+  import ArbDeclination.given
+
+  override def scalaCheckInitialSeed = "Gdvwgu9aVUkJ7kMUl5BHZvfg1c5QPMnJ4EGutEgXr6A="
```
cquiroz, line 23: Why the seed?

cquiroz (approved):
LGTM

modules/tests/shared/src/test/scala/lucuma/core/math/ArcSuite.scala
tpolecat: Good catch, thanks.

PR #1116, opened 2025-06-05
Gmos binnnig

cquiroz:
Imaging supports a single binning parameter, thus having X and Y is maybe unnecessary, though used in spectroscopy.

The enum is being unified. However I kept opaque types for X and Y meaning to simplify the transition.

Not sure if it is a good idea. I'd keep open this until I can show it can work up to the odb


cquiroz:
`lucuma-itc` works

swalker2m:
> However I kept opaque types for X and Y meaning to simplify the transition.

Seems like a good idea regardless, to avoid the possibility of mixing up X and Y binning values in the spectroscopy observing mode configuration.  If the ODB update goes fine then this seems good to me.

rpiaggio (approved):
LGTM

cquiroz:
> Nice. I like the idea.
> 
> You can put extensions in the newtypes exposing `count` (and maybe `shortName`, `longName`), so that we don't have to modify code everywhere.

Great idea

PR #1125, opened 2025-06-13
Implement GaiaClient

rpiaggio:
Implement a full-fledged blackbox Gaia client that can be invoked from odb and explore, instead of having duplicated code to make the HTTP invocation+parse at each call site.

I coalesced as much code as possible into the new `GaiaClientImpl`. The `CatalogAdapter.Gaia` is now an explicit parameter.

Added bonus: syntax for `F[A].raceToSuccess(F[B])` and `NonEmptyChain[F[A]].raceAllToSuccess`.

There was some commented code in the sample queries, I left it there in case it's useful at some other time.


modules/core/shared/src/main/scala/lucuma/core/syntax/effect.scala
```diff
@@ -0,0 +1,57 @@
...
+import cats.Functor
+import cats.data.NonEmptyChain
+import cats.effect.Outcome
+import cats.effect.Spawn
+import cats.syntax.all.*
+
+trait effect:
+  extension [F[_]](F: Spawn[F])
+    /**
+     * Similar to `F.race`, but in case of error in the winner fiber, keeps the other one. Note that
```
cquiroz, line 16: shoould we use this for the simbad query?

modules/core/shared/src/main/scala/lucuma/core/syntax/effect.scala
```diff
@@ -0,0 +1,57 @@
...
+package lucuma.core.syntax
+
+import cats.FlatMap
+import cats.Functor
+import cats.data.NonEmptyChain
+import cats.effect.Outcome
+import cats.effect.Spawn
+import cats.syntax.all.*
+
+trait effect:
```
toddburnside, line 13: Very nice

toddburnside (approved):
Look great. Thanks for doing this!

modules/core/shared/src/main/scala/lucuma/core/syntax/effect.scala
rpiaggio: Yes, I'm planning to use it in explore when train reaches there.

cquiroz (approved):
LGTM

PR #1129, opened 2025-06-19
raceToSuccess: don't cancel fiber on error/cancel of the other

rpiaggio:
I copied the `raceToSuccess` code from cats-effect's `race` and adapted it, but I don't think the cats-effect code is doing what its scaladoc says it's doing when one of the fibers cancels (which is that other one should keep going).

Anyway, when racingToSuccess, if one fiber errored, the other was canceled. That's precisely what we want to avoid here. This PR fixes the situation.


cquiroz (approved):
LGTM

PR #1132, opened 2025-07-03
add offset generator functions to core

cquiroz:
it also adds an ABBA class for the offset pattern

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/OffsetGenerator.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/OffsetGenerator.scala
@@ -21,24 +21,19 @@ object OffsetGenerator:
   /**
    * Generates a grid pattern of offsets.
    */
-  def grid(
-    nx: PosInt,
-    ny: PosInt,
-    dp: Angle,
-    dq: Angle,
-    center: Offset = Offset.Zero
-  ): NonEmptyList[Offset] =
+  def grid(nx: PosInt, ny: PosInt, dp: Angle, dq: Angle, center: Offset = Offset.Zero): NonEmptyList[Offset] =
     val pStart = -(nx.value - 1) / 2.0
     val qStart = (ny.value - 1) / 2.0
 
-    val offsets = for {
-      j <- 0 until ny.value
-      i <- 0 until nx.value
-    } yield {
-      val pOffset = dp * (pStart + i)
-      val qOffset = dq * (qStart - j)
-      center + Offset(pOffset.p, qOffset.q)
-    }
+    val offsets =
+      for
+        j <- 0 until ny.value
+        i <- 0 until nx.value
+      yield
+        val pOffset = dp * (pStart + i)
+        val qOffset = dq * (qStart - j)
+        center + Offset(pOffset.p, qOffset.q)
+
     // can't be empty as nx and ny are 1 or more
     NonEmptyList.fromListUnsafe(offsets.toList)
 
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/ABBA.scala
```diff
@@ -0,0 +1,52 @@
...
+import cats.derived.*
+import cats.syntax.eq.*
+import lucuma.core.math.Offset
+import monocle.Focus
+import monocle.Lens
+
+/**
+ * An ABBA pattern containing exactly 4 offsets with symmetric structure.
+ */
+case class ABBA(a: Offset, b: Offset) derives Eq:
```
swalker2m, line 16: I think this is usually right but sometimes the two a’s and/or b’s aren’t exactly the same. We have to leave open the possibility that they differ or it won’t be usable by F2 long slit at least. Maybe a constructor that takes an ‘a’ and a ‘b’ but in general I think we need four positions.

cquiroz: Got his from science:
```
sw
[5:20] are they always symmetric?
[5:21]-a, +a, +a, -a

science:
[5:21] yes, e.g. +3, -2, -2, +3

sw
[5:26] no odd astronomer will do -1, 3, -4, 5?

science:
[5:27 PM] they may, but that's not ABBA :slightly_smiling_face:
```

swalker2m: From the requirements spreadsheet examples:

Screenshot 2025-07-04 at 10 06 06 AM

The code looks great though, and it will probably be used in other modes so :+1:

swalker2m: No worries, maybe we eventually could use an `Abba-ish` class with 4 positions but for now F2 Long Slit can continue working just fine with a `List` + validation.

modules/core/shared/src/main/scala/lucuma/core/geom/OffsetGenerator.scala
```diff
@@ -0,0 +1,138 @@
...
+   */
+  def gridP[F[_]: Applicative](num: PosInt, step: Angle, center: Offset.P = Offset.P.Zero): F[NonEmptyList[Offset.P]] =
+    val start = -(num.value - 1) / 2.0
+
+    val components =
+      (0 until num.value).map: i =>
+        val pOffset = step * (start + i)
+        center + Offset.P(pOffset)
+
+    NonEmptyList.fromListUnsafe(components.toList).pure[F]
```
swalker2m, line 99: This one also could maybe just return a `NonEmptyList` directly?  Or maybe `grid` could be done in terms of `gridP` and `gridQ` or vice versa.  Like `gridP(nx, dp, center).zipWith(gridQ(ny, dq, center)((p, q) => Offset(p, q))` or something.  The same for `random` / `randomP` / `randomQ` perhaps.

swalker2m (approved):
Looks great to me.  Some minor comments, but nothing important. :+1:

modules/core/shared/src/main/scala/lucuma/core/geom/OffsetGenerator.scala
```diff
@@ -0,0 +1,138 @@
...
+      for
+        j <- 0 until ny.value
+        i <- 0 until nx.value
+      yield
+        val pOffset = dp * (pStart + i)
+        val qOffset = dq * (qStart - j)
+        center + Offset(pOffset.p, qOffset.q)
+
+    // can't be empty as nx and ny are 1 or more
+    NonEmptyList.fromListUnsafe(offsets.toList).pure[F]
```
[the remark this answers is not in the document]
cquiroz: right I thought it maybe better to make them all return on `F[_]`

PR #1139, opened 2025-07-11
Add `AtomDigest` type

swalker2m:
Adds a straightforward `AtomDigest` type with the per-atom information requested by the scheduler.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/AtomDigest.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/AtomDigest.scala
@@ -6,6 +6,7 @@ package lucuma.core.model.sequence
 import cats.Eq
 import lucuma.core.enums.GcalLampType
 import lucuma.core.enums.ObserveClass
+import lucuma.core.enums.StepType
 import monocle.Focus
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/SequenceDigest.scala
```diff
@@ -35,22 +35,14 @@ case class SequenceDigest(
...
-        SequenceDigest.atomCount.replace(ac)(this)
-      }
+  def add[D](a: Atom[D]): SequenceDigest =
+    SequenceDigest(
+      observeClass   = observeClass |+| a.observeClass,
+      timeEstimate   = timeEstimate |+| a.timeEstimate,
+      offsets        = a.steps.toList.foldLeft(offsets)((os, s) => os + s.telescopeConfig.offset),
+      atomCount      = NonNegInt.unsafeFrom(atomCount.value + 1),
+      executionState = executionState
+    )
```
swalker2m, line 45: This is more or less unrelated but fell out of an aborted previous version and strikes me as better than the individual `add`s that the ODB was previously using.

PR #1141, opened 2025-07-17
Add MultipleFiltersMode enum

toddburnside:
For imaging configurations with multiple filters selected. Will determine how the sequence is generated.

I'm open to other names.


toddburnside:
> do we need a manual mode?

AFAIK, that is not a requirement at this point. According to the story, `The options for "Multiple Filters" are either "Grouped" or "Interleaved".`
https://app.shortcut.com/lucuma/story/4029/add-configuration-ui

PR #1145, opened 2025-07-25
Shortcut 6121: Switch exposure time mode count to `PosInt`

swalker2m:
Per discussion, the requested exposure count must be positive.


swalker2m:
I don't understand why this change doesn't require a new `tlBaseVersion` number, but in MiMa we trust.

PR #1151, opened 2025-07-29
Shortcut 6168: GMOS R400 tag change

swalker2m:
Updates the tag for the GMOS North R400 grating to reflect the new value.

Obviously we have not implemented Resource and we are not yet using the database as the ultimate source of truth for enum values.  The plan to address the R400 swap before XT1 then is to simply rename the R400 grating and rewrite history (if any test steps have been written using the old grating).  Does anybody have a better idea?


toddburnside (commented):
Seems good for now.

PR #1152, opened 2025-07-30
Make AirMass.minumumFor return an Option

toddburnside:
This code is being used to determine the likelihood of conditions.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/package.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/package.scala
@@ -63,10 +63,11 @@ object AirMass extends NewRefined[BigDecimal, Not[Less[1]]]:
   /**
     * Minimum airmass that a given declination reaches from a given latitude.
     */
-  def minimumFor(dec: Declination, latitude: Lat): AirMass =
+  def minimumFor(dec: Declination, latitude: Lat): Option[AirMass] =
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/package.scala
```diff
@@ -63,10 +63,11 @@ object AirMass extends NewRefined[BigDecimal, Not[Less[1]]]:
   /**
     * Minimum airmass that a given declination reaches from a given latitude.
     */
-  def minimumFor(dec: Declination, latitude: Lat): AirMass =
+  def minimumFor(dec: Declination, latitude: Lat): Option[AirMass] =
     // Maximum elevation in degrees
     val elevation = 90.0 - abs(dec.toAngle.toSignedDoubleDegrees - latitude.toAngle.toSignedDoubleDegrees)
-    AirMass.from(BigDecimal(1.0 / sin((elevation + 244.0 / (165.0 + 47.0 * pow(elevation, 1.1))) * Pi / 180.0))).getOrElse(sys.error("Not possible"))
+    if elevation < 0 then None
```
toddburnside, line 69: When a target below the horizon was selected, `elevation` would be negative, causing `pow(elevation, 1.1)` to return an `NaN`, resulting in `BigDecimal` throwing an exception.

PR #1153, opened 2025-07-31
Add myself to the skip-sc workflow

toddburnside:
A PR in core seldom is all that is needed to fix a problem, so I'm adding myself to skip-sc workflow.


swalker2m (approved):
Come on in, the water's fine.

PR #1155, opened 2025-07-31
Shortcut 6292: Delete G5323

swalker2m:
Removes the GMOS South G5323 grating option.


toddburnside (approved):
LGTM after mima bump. 🚂

PR #1157, opened 2025-08-01
Shortcut 5099: Add client id to execution events

swalker2m:
Adds an optional client id to the execution events.  If observe specifies a client id when it adds an event, and if that client id has been previously used, the attempt would fail with a duplicated event error.  (Requires a corresponding update to the ODB of course.)  How does this sound?

Note, this reuses the `c` prefix, but in a UUID-based id (the Call For Proposals id is gid-based).  I hope this isn't too confusing. 

```
c-123                                  => call for proposals id
c-530c979f-de98-472f-9c23-a3442f2a9f7f => client id
```

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -1,6 +1,6 @@
 import org.scalajs.linker.interface.ESVersion
 
-ThisBuild / tlBaseVersion                         := "0.139"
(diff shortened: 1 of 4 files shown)


toddburnside (approved):
I'm not doing any work in this area, but it seems fine to me.

PR #1164, opened 2025-08-21
sc-4512 configuration updates

tpolecat:
Subsumes #1154 

This updates the `Configuration` model to support TOO targets, as well as GMOS imaging and F2 longslit configurations. This is a breaking change.


toddburnside (approved):
LGTM

PR #1202, opened 2025-10-03
Add model for Tellluric type

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/model/TelluricType.scala
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Possible telluric calibration type
+ */
+sealed trait TelluricType(val tag: String) derives Eq
+
+object TelluricType:
+  case object Hot                                     extends TelluricType("Hot")
+  case object A0V                                     extends TelluricType("A0V")
+  case object Solar                                   extends TelluricType("Solar")
+  case class  Manual(starTypes: NonEmptyList[String]) extends TelluricType("Manual") derives Eq
```
swalker2m, line 23: What do we expect for `starTypes`?

PR #1345, opened 2026-03-31
Downgrade to Scala 3 LTS

hugo-vrijswijk:


hugo-vrijswijk:
Downgrading seems to generate a few too many compile errors related to newer features that I don't want to back-fix

PR #1353, opened 2026-04-07
Shortcut 8293: GHOST Updates

swalker2m:
Some tweaks to core GHOST types:

* I inadvertently left `GhostIfu` out of the `Enumerated` instance definition for `ObservingModeType`, causing much grief and consternation.  Switched to derive `Enumerated` instead.
* Switched the fiber agitator business from a `NewBoolean` to a vanilla pair of Scala enums.  I got the other to work with the ODB but it was kind of a hassle.
* Cosmetic: now that we have scads of observing modes I'm trying to sort them alphabetically everywhere they appear.


modules/core/shared/src/main/scala/lucuma/core/enums/ObservingModeType.scala
```diff
@@ -7,23 +7,14 @@ import lucuma.core.math.Angle
 import lucuma.core.model.PosAngleConstraint
 import lucuma.core.util.Enumerated
 
-enum ObservingModeType(val dbTag: String, val instrument: Instrument, val defaultPosAngleConstraint: PosAngleConstraint):
-  case GmosNorthLongSlit  extends ObservingModeType("gmos_north_long_slit",  Instrument.GmosNorth,  PosAngleConstraint.AverageParallactic)
-  case GmosSouthLongSlit  extends ObservingModeType("gmos_south_long_slit",  Instrument.GmosSouth,  PosAngleConstraint.AverageParallactic)
+enum ObservingModeType(val dbTag: String, val instrument: Instrument, val defaultPosAngleConstraint: PosAngleConstraint) derives Enumerated:
```
cquiroz: you can get the enumeryrf for "free" if this field is renamed to `tag`

cquiroz (approved):
LGTM

PR #1357, opened 2026-04-12
Don't use z prime for acquisition

cquiroz:


arturog8m (approved):
Looks straightforward 👍

swalker2m:
You're probably already aware of this, but just in case a corresponding update to the database table will be required.  From a past migration:

```
UPDATE t_gmos_north_filter
  SET c_is_acquisition_filter = TRUE
  WHERE c_tag IN ('GPrime', 'RPrime', 'IPrime', 'ZPrime');
```

cquiroz:
Yes, we need to update odb before closing this task

PR #1360, opened 2026-04-14
blind offset candidate query failing for gea

cquiroz:
Blind offset search over gaia lite were failing due to the use of the coulmn `astrometric_excess_noise` which is not present on gaia lite. The fix is to switch to plain gaia.

Also tracing was added

### DIFF
--- a/modules/catalog/shared/src/main/scala/lucuma/catalog/clients/GaiaClient.scala
+++ b/modules/catalog/shared/src/main/scala/lucuma/catalog/clients/GaiaClient.scala
@@ -15,6 +15,7 @@ import lucuma.core.model.Target
 import org.http4s.Uri
 import org.http4s.client.Client
 import org.typelevel.log4cats.LoggerFactory
+import org.typelevel.otel4s.trace.Tracer
 
 trait GaiaClient[F[_]]:
   /**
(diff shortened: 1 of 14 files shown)


rpiaggio (approved):
LGTM

modules/catalog/shared/src/main/scala/lucuma/catalog/clients/GaiaClient.scala
```diff
@@ -69,5 +70,5 @@ object GaiaClient:
     NonEmptyChain.of(
       CatalogAdapter.Gaia3DataLab,
       CatalogAdapter.Gaia3LiteGavo,
-      CatalogAdapter.Gaia3LiteEsaProxy
+      CatalogAdapter.Gaia3Esa
```
toddburnside: It was using the proxy before. We don't need to use it now?

toddburnside (approved):
LGTM

modules/catalog/shared/src/main/scala/lucuma/catalog/clients/GaiaClient.scala
cquiroz: There is a mistake, I have to use the new `Gaia3EsaProxy`

PR #1365, opened 2026-04-17
Shortcut 8293: Minor convenience, rename.

swalker2m:
A bit of GHOST convenience.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostDetector.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostDetector.scala
@@ -6,7 +6,6 @@ package lucuma.core.model.sequence.ghost
 import cats.Eq
 import eu.timepit.refined.cats.given
 import eu.timepit.refined.types.numeric.PosInt
-import eu.timepit.refined.types.string.NonEmptyString
 import lucuma.core.enums.GhostBinning
 import lucuma.core.enums.GhostReadMode
 import lucuma.core.util.NewType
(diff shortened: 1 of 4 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostDetector.scala
```diff
@@ -16,7 +17,26 @@ final case class GhostDetector(
   exposureCount:  PosInt,
   binning:        GhostBinning,
   readMode:       GhostReadMode
-)
+):
+
+  /**
+   * Textual representation of exposure time and count.
+   */
+  def description: NonEmptyString =
```
rpiaggio: Would it make sense to use `Display` for this?

rpiaggio: Otherwise, `lazy val` instead of `def`?

rpiaggio (approved):
LGTM, with a couple of comments.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostDetector.scala
swalker2m: > Would it make sense to use `Display` for this?

I was going to use it in the atom description but now that I think about it, it's really a step-level thing.  I'll just delete it altogether.

> Otherwise, `lazy val` instead of `def`?

It could be a normal `val`, there's no need for laziness, but the computation is cheap and I suppose I thought I would avoid storing it.  I'll just switch to `val`.

PR #1366, opened 2026-04-20
Shortcut 8293: Add guide and slit viewing camera times

swalker2m:
Adds the optional guide and slit viewing camera exposure times to the GHOST static configuration.  These are engineering parameters, as I understand it, which are only set by Observe when specified.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostStaticConfig.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostStaticConfig.scala
@@ -4,14 +4,14 @@
 package lucuma.core.model.sequence.ghost
 
 import cats.Eq
+import cats.derived.*
 import lucuma.core.enums.GhostResolutionMode
+import lucuma.core.util.TimeSpan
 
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostStaticConfig.scala
```diff
@@ -4,14 +4,14 @@
...
 final case class GhostStaticConfig(
-  resolutionMode: GhostResolutionMode
-)
-
-object GhostStaticConfig:
-  given Eq[GhostStaticConfig] =
-    Eq.by(_.resolutionMode)
\ No newline at end of file
+  resolutionMode:                GhostResolutionMode,
+  guideCameraExposureTime:       Option[TimeSpan],
```
cquiroz: FWIW GuideCamera override is disabled in the seqexec

cquiroz: I think we don't need this field in gpp

PR #1367, opened 2026-04-21
Update `AtomDigest`

swalker2m:
The scheduler team requests the monotonically increasing start step index and step count in each `AtomDigest` entry.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/AtomDigest.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/AtomDigest.scala
@@ -4,6 +4,10 @@
 package lucuma.core.model.sequence
 
 import cats.Eq
+import cats.derived.*
+import eu.timepit.refined.cats.*
+import eu.timepit.refined.types.numeric.NonNegInt
+import eu.timepit.refined.types.numeric.PosInt
 import lucuma.core.enums.GcalLampType
 import lucuma.core.enums.ObserveClass
 import lucuma.core.enums.StepType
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/sequence/AtomDigest.scala
```diff
@@ -18,12 +22,14 @@ case class AtomDigest(
...
-)
+  lampTypes:    Set[GcalLampType],
+  stepIndex:    NonNegInt,
+  stepCount:    PosInt
+) derives Eq
 
 object AtomDigest:
 
-  def fromAtom[D](atom: Atom[D]): AtomDigest =
+  def fromAtom[D](stepIndex: NonNegInt)(atom: Atom[D]): AtomDigest =
```
cquiroz, line 32: why currying?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/sequence/AtomDigest.scala
swalker2m: Oh, over in the `ODB` where this is used I map a `Stream` of atoms to atom digest.  Now that I look at it though, I had to turn `atomStream.map(AtomDigest.fromAtom)` into a `mapAccumlate` in order to calculate the `stepIndex` and the curry didn't end up helping.

cquiroz: it's the only use I have found curry to be useful so far

PR #1369, opened 2026-04-22
Use the latest telluric catalog xhip simbad

cquiroz:
The tellurics catalog changed the format and now it returns a SED. This PR updates the core client for that server


swalker2m (approved):
:+1:

PR #1370, opened 2026-04-23
Gnirs configurations and other classes

rpiaggio:
- Defines `GnirsStaticConfig` and `GnirsDynamicConfig`.
- Adds Gnirs to `InstrumentExecutionConfig`.
- Relevant enums were simplified.

BONUS:
- `Display` instance derivation when there are `shortName` and `longName` members.
- `NewType` and friends now forward `Hash` instances of the base type. Tricky since both `Hash` and `Order` extend `Eq`, so having both would result in ambiguous `Eq`.


modules/core/shared/src/main/scala/lucuma/core/model/ExposureTimeMode.scala
```diff
@@ -16,32 +16,40 @@ import monocle.Prism
...
-    val at:    Lens[SignalToNoiseMode, Wavelength]    = Focus[SignalToNoiseMode](_.at)
+    val at: Lens[SignalToNoiseMode, Wavelength]       = Focus[SignalToNoiseMode](_.at)
     given Eq[SignalToNoiseMode]                       = Eq.by(a => (a.value, a.at))
 
-
-  final case class TimeAndCountMode(time: TimeSpan, count: PosInt, at: Wavelength) extends ExposureTimeMode
+  final case class TimeAndCountMode(
+    time:   TimeSpan,
+    count:  PosInt,
+    coadds: PosInt,
```
cquiroz, line 33: Should we default this to one to make transition easier

modules/core/shared/src/main/scala/lucuma/core/util/Display.scala
```diff
@@ -56,4 +58,18 @@ object Display {
...
+    import quotes.reflect.*
+    Select.unique(x.asTerm, "shortName").asExprOf[String]
+
+  private def longNameImpl[E](x: Expr[E])(using Quotes): Expr[String] =
+    import quotes.reflect.*
+    Select.unique(x.asTerm, "longName").asExprOf[String]
+
+  private def displayImpl[E: Type](using Quotes): Expr[Display[E]] =
+    '{
+      Display.by[E](x => ${ shortNameImpl[E]('x) }, x => ${ longNameImpl[E]('x) })
```
cquiroz, line 72: nice

cquiroz (approved):
LGTM

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/ExposureTimeMode.scala
rpiaggio: Yes, there's an `apply` method without `coadds` that defaults to 1.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDecker.scala
```diff
@@ -4,50 +4,25 @@
...
-    }
-
-}
+) derives Enumerated, Display:
+  case Acquisition extends GnirsDecker("Acquisition", "Acquisition", "Acquisition", false)
+  case PupilViewer extends GnirsDecker("PupilViewer", "Pupil", "Pupil viewer", false)
+  case ShortCamCrossDispersed extends GnirsDecker("ShortCamCrossDispersed", "Short camera XD", "Short camera cross dispersed", false)
+  case Ifu extends GnirsDecker("Ifu", "IFU", "Integral field unit", true)
+  case LongCamLongSlit extends GnirsDecker("LongCamLongSlit", "Long camera slit", "Long camera long slit", false)
+  case Wollaston extends GnirsDecker("Wollaston", "Wollaston", "Wollaston", true)
```
swalker2m, line 26: I think we can just delete the `obsolete` items like `Wollaston` and `Ifu`.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFpuSlit.scala
```diff
@@ -4,52 +4,26 @@
...
-}
+) derives Enumerated, Display:
+  case LongSlit_0_10 extends GnirsFpuSlit("LongSlit_0_10", "0.10\"", "0.10 arcsec", Angle.fromDoubleArcseconds(0.100), false)
+  case LongSlit_0_15 extends GnirsFpuSlit("LongSlit_0_15", "0.15\"", "0.15 arcsec", Angle.fromDoubleArcseconds(0.150), false)
+  case LongSlit_0_20 extends GnirsFpuSlit("LongSlit_0_20", "0.20\"", "0.20 arcsec", Angle.fromDoubleArcseconds(0.200), false)
+  case LongSlit_0_30 extends GnirsFpuSlit("LongSlit_0_30", "0.30\"", "0.30 arcsec", Angle.fromDoubleArcseconds(0.300), false)
+  case LongSlit_0_45 extends GnirsFpuSlit("LongSlit_0_45", "0.45\"", "0.45 arcsec", Angle.fromDoubleArcseconds(0.450), false)
+  case LongSlit_0_675 extends GnirsFpuSlit("LongSlit_0_675", "0.675\"", "0.675 arcsec", Angle.fromDoubleArcseconds(0.675), false)
+  case LongSlit_1_00 extends GnirsFpuSlit("LongSlit_1_00", "1.0\"", "1.0 arcsec", Angle.fromDoubleArcseconds(1.000), false)
+  case LongSlit_3_00 extends GnirsFpuSlit("LongSlit_3_00", "3.0\"", "3.0 arcsec", Angle.fromDoubleArcseconds(3.000), true)
```
swalker2m, line 29: Here too.  Since the 3" slit is no longer available and we have no existing GNIRS observations, it doesn't need to be kept around.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
```diff
@@ -0,0 +1,25 @@
...
+final case class GnirsDynamicConfig(
+  exposure:          TimeSpan,
+  filter:            GnirsFilter,
+  decker:            GnirsDecker,
+  fpu:               Either[GnirsFpuSlit, GnirsFpuOther],
+  acquisitionMirror: GnirsAcquisitionMirrorMode,
+  camera:            GnirsCamera,
+  focus:             GnirsFocus,
+  readMode:          GnirsReadMode
+) derives Eq
```
swalker2m, line 25: I often check with the Observing Tool to see what we have available there in the corresponding instrument iterator:

Image

* Are we missing "coadds"?
* Also what the OT calls "Cross-dispersed" but here we call "Prism"?
* The "Central Wavelength" is also configurable per step in the OT.
* Hartmann Mask doesn't exist at all, but maybe that is intentional I don't know.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentExecutionConfig.scala
```diff
@@ -84,20 +87,37 @@ object InstrumentExecutionConfig:
...
   case class Igrins2(
     executionConfig: ExecutionConfig[ig2.Igrins2StaticConfig, ig2.Igrins2DynamicConfig]
   ) extends InstrumentExecutionConfig derives Eq:
     override def instrument: Instrument = Instrument.Igrins2
     override def isComplete: Boolean    = executionConfig.isComplete
 
   object Igrins2:
-    val executionConfig: Lens[Igrins2, ExecutionConfig[ig2.Igrins2StaticConfig, ig2.Igrins2DynamicConfig]] =
+    val executionConfig
+      : Lens[Igrins2, ExecutionConfig[ig2.Igrins2StaticConfig, ig2.Igrins2DynamicConfig]] =
```
swalker2m, line 120: Auto formatting is bad.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentExecutionConfig.scala
```diff
@@ -106,8 +126,8 @@ object InstrumentExecutionConfig:
...
     Eq.instance:
       case (a @ Flamingos2(_), b @ Flamingos2(_)) => a === b
-      case (a @ Ghost(_),      b @ Ghost(_))      => a === b
-      case (a @ GmosNorth(_),  b @ GmosNorth(_))  => a === b
-      case (a @ GmosSouth(_),  b @ GmosSouth(_))  => a === b
-      case (a @ Igrins2(_),    b @ Igrins2(_))    => a === b
+      case (a @ Ghost(_), b @ Ghost(_))           => a === b
+      case (a @ GmosNorth(_), b @ GmosNorth(_))   => a === b
+      case (a @ GmosSouth(_), b @ GmosSouth(_))   => a === b
+      case (a @ Igrins2(_), b @ Igrins2(_))       => a === b
```
swalker2m, line 132: Did you want to add GNIRS here too?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsStaticConfig.scala
```diff
@@ -0,0 +1,12 @@
...
+
+package lucuma.core.model.sequence.gnirs
+
+import cats.Eq
+import cats.derived.*
+import lucuma.core.enums.GnirsWellDepth
+
+final case class GnirsStaticConfig(
+  wellDepth: GnirsWellDepth
+) derives Eq
```
swalker2m, line 12: Should it also have Pixel Scale?

swalker2m (commented):
I think it all looks great.  The questions about what to include or not in the dynamic and static configs are actual questions in that I don't know the answers.  I'm just basing them on what I see in the OT.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDecker.scala
rpiaggio: I'll remove them, as well as the `obsolete` flag.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFpuSlit.scala
rpiaggio: Same

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
rpiaggio: > Are we missing "coadds"?

That's been added to `ExposureTimeMode.TimeAndCountMode`.

> Also what the OT calls "Cross-dispersed" but here we call "Prism"?

Yes, that seems to be the case. It has the same 3 options.

> The "Central Wavelength" is also configurable per step in the OT.

Does that mean we should add it to the dynamic config?

> Hartmann Mask doesn't exist at all, but maybe that is intentional I don't know.

Yeah, Andy said it's been removed.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentExecutionConfig.scala
rpiaggio: Yup, thanks for the catch.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsStaticConfig.scala
rpiaggio: Pixel Scale is actually a property of `GnirsCamera`. Maybe the OT UI provided as a way to filter the cameras?

modules/core/shared/src/main/scala/lucuma/core/model/sequence/InstrumentExecutionConfig.scala
rpiaggio: I reverted this one. I agree that it's bad in a few pathological cases. The time and attention that it saves in all the other cases are still worth it for me.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
swalker2m: >> Are we missing "coadds"?

> That's been added to ExposureTimeMode.TimeAndCountMode.

Uff, I missed that.  That'll ripple through the ODB.  Not all instruments even have coadds?

At any rate, the dynamic config is applied to the instrument and should have a coadds configuration if the instrument supports it.  The exposure time mode will appear, I assume, in the observing mode and be used to obtain the exposure time (and coadds now?).

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsStaticConfig.scala
swalker2m: The OT, which is likely wrong, allows the user to select the Pixel Scale on the instrument component. 

Image

Then in the instrument "iterator" component they can independently select a different camera in each step.  Nothing enforces any correspondence.

Image

It sounds like that wasn't done correctly.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
swalker2m: >> The "Central Wavelength" is also configurable per step in the OT.

> Does that mean we should add it to the dynamic config?

I think so, each GMOS step for example has a central wavelength (though it is derived from the grating and filter).

PR #1372, opened 2026-04-23
Address previous PR comments for gnirs configs

rpiaggio:
Removed obsolete options, added missing case to `Eq[InstrumentExecutionConfig]`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import org.scalajs.linker.interface.ESVersion
 import org.typelevel.sbt.gha.PermissionValue
 import org.typelevel.sbt.gha.Permissions
 
-ThisBuild / tlBaseVersion                         := "0.185"
+ThisBuild / tlBaseVersion                         := "0.186"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDecker.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDecker.scala
@@ -16,13 +16,10 @@ enum GnirsDecker(
   val tag: String,
   val shortName: String,
   val longName: String,
-  val obsolete: Boolean
 ) derives Enumerated, Display:
-  case Acquisition extends GnirsDecker("Acquisition", "Acquisition", "Acquisition", false)
-  case PupilViewer extends GnirsDecker("PupilViewer", "Pupil", "Pupil viewer", false)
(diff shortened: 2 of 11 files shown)


swalker2m (approved):
Looks great.  I do think the dynamic config needs coadds if it is a setting that should be applied to the instrument.

rpiaggio:
> Looks great. I do think the dynamic config needs coadds if it is a setting that should be applied to the instrument.

I imagined observe sending it to the instrument from the ETM. That may be unclear though.

On the other hand, I don't like the idea of having it twice (ETM and config).

Alternatively, we could keep it only in the config and take it into account when we invoke ITC.

Not clear which option is better, kind of leaning towards the last one. WDYT?

swalker2m:
> I imagined observe sending it to the instrument from the ETM. That may be unclear though.
> On the other hand, I don't like the idea of having it twice (ETM and config).
> Alternatively, we could keep it only in the config and take it into account when we invoke ITC.
> Not clear which option is better, kind of leaning towards the last one. WDYT?

The exposure time mode and the instrument configuration are different concepts, but there is overlap.  The exposure time mode is an ITC input.  If your exposure time mode is S/N, then you are asking the ITC for the time and count.  If it is `TimeAndCount`, then you are asking the ITC for S/N.  On the other hand you cannot configure the instrument itself with a S/N exposure time mode.

The sequence is essentially a list of instrument configurations, which include filters, FPUs, and the exposure time and count.  Also, _the exposure time ~and count~ will vary over the course of the sequence_.  Calibrations in particular have different exposure times ~and counts~.

I think of the observing mode (i.e., `GnirsLongSlit`) as a high-level description of the observation that is used to produce the sequence.  The sequence itself as a  low-level `List[GnirsDynamicConfig]`.  So I think we need both the `ExposureTimeMode` in the `GnirsLongSlit` observing mode, and then the exposure time, ~count~, and coadds in the `GnirsDynamicConfig`.

swalker2m:
Also some instruments, like GMOS, have multiple observing modes (Long Slit, Imaging, IFU, MOS) and multiple ETM requirements.  GMOS imaging has an ETM per filter for example.  The GMOS sequence itself is the same (with time ~and count~) regardless of the mode that was used to produce it.

rpiaggio:
Thank you for the explanation. It's more clear to me now how observing mode and configs interact. I'll add coadds to the config.

But I thought `count` (in T+C mode) got translated into the number of steps. Unless I'm missing something, I don't see why it would get to be a property of the dynamic config.

swalker2m:
That's right, the count in all the instruments except Ghost is the step count and doesn't belong in the DynamicConfig. Coadds, as I understand it, are images that are added together in a single step so that would be part of the DynamicConfig.

swalker2m:
I updated my comments above to strike out "count" which I mistakenly included. Sorry for the confusion.

PR #1374, opened 2026-04-26
Initial GHOST configuration

swalker2m:
Adds at least a placeholder for GHOST to the configuration request model.  I don't know if there is anything instrument-specific in particular that needs to be approved when changed (resolution mode?) but this will enable an ODB update that will make the configuration request work for conditions and targets for GHOST observations.


cquiroz (approved):
LGTM

PR #1376, opened 2026-04-28
Add missing centralWavelength to GnirsDynamicConfig; filter resolution for spectroscopy

rpiaggio:
- It seems that in Gnirs the central wavelength can be specified instead of using the filter's.

- In old ITC, the spectroscopy filter is resolved depending on the wavelength. In GPP we will deduce it using seqexec's logic and pass it to ITC.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
@@ -4,10 +4,13 @@
 package lucuma
 package core
 package enums
+
+import cats.syntax.all.*
 import lucuma.core.math.Wavelength
 import lucuma.core.util.Display
 import lucuma.core.util.Enumerated
 
+import ConvenienceOps.*
 
 /**
  * Enumerated type for GNIRS Filter.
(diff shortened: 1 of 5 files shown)


cquiroz:
>     In GPP we will deduce it using seqexec's logic and pass it to ITC.
In general we want to avoid having this kind of logic in the seqexec itself. Could we put it instead either as a part of core or on the odb?
Ideally observe should just pass the values between the odb and instrument rather than manipulating them

rpiaggio:
> > ```
> > In GPP we will deduce it using seqexec's logic and pass it to ITC.
> > ```
> 
> In general we want to avoid having this kind of logic in the seqexec itself. Could we put it instead either as a part of core or on the odb? Ideally observe should just pass the values between the odb and instrument rather than manipulating them

Right, this is what we are trying to achieve here. The logic is here in core and will be used by the ODB to query ITC and also put the filter in GnirsDynamicConfig so that observe can pass it directly to the instrument.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
```diff
@@ -17,19 +20,29 @@ enum GnirsFilter(
...
+  case HNd100x extends GnirsFilter("HNd100x", "H+ND100X", "H + ND100X", 1_650_000.pm.some, none)
+  case H2Nd100x extends GnirsFilter("H2Nd100x", "H2+ND100X", "H2 + ND100X", 2_120_000.pm.some, none)
+  case PAH extends GnirsFilter("PAH", "PAH", "PAH: 3.3µm", 3_300_000.pm.some, none)
+  case Y extends GnirsFilter("Y", "Y", "Y: 1.03µm", 1_030_000.pm.some, none)
+  case J extends GnirsFilter("J", "J", "J: 1.25µm", 1_250_000.pm.some, none)
+  case K extends GnirsFilter("K", "K", "K: 2.20µm", 2_200_000.pm.some, none)
+
+object GnirsFilter:
+  private val SpectroscopyFilterTable: List[(GnirsFilter, Wavelength)] = 
+    values.map(f => f.spectroscopyCutoffWavelength.map(w => (f, w))).toList.flattenOption
```
cquiroz, line 43: `tupleLeft`?

cquiroz (approved):
LGTM

toddburnside:
> > > ```
> > > In GPP we will deduce it using seqexec's logic and pass it to ITC.
> > > ```
> > 
> > 
> > In general we want to avoid having this kind of logic in the seqexec itself. Could we put it instead either as a part of core or on the odb? Ideally observe should just pass the values between the odb and instrument rather than manipulating them
> 
> Right, this is what we are trying to achieve here. The logic is here in core and will be used by the ODB to query ITC and also put the filter in GnirsDynamicConfig so that observe can pass it directly to the instrument.

The `using seqexec's logic` that Raul is referring to here is a reference to the fact that the logic of using the central wavelength to deduce the filter was implemented in several places in the old ITC and seqexec, each of which used slightly different wavelength cutoffs. In a chat with Raul and Andy yesterday, we decided it was best to use the same cutoff values as the seqexec as our initial implementation. They can be refined later if needed.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
```diff
@@ -17,19 +20,29 @@ enum GnirsFilter(
...
+
+  // Adapted from seqexec
+  def fromSpectroscopyWavelength(wavelength: Wavelength): GnirsFilter = 
+    SpectroscopyFilterTable.foldRight[GnirsFilter](GnirsFilter.CrossDispersed):  (entry, selected) =>
```
toddburnside: I'm curious about the use of `CrossDispersed` as the initial element here. CrossDispersed is only valid for cases where cross dispersion is being used, right? So, if a central wavelength that is too large is specified, will we end up with `CrossDispersed` as the filter? It seems like we would want that to be an error of some sort, wouldn't we?

rpiaggio: I agree.

@jluhrs Do you remember why the decision to use XD as the default instead of erroring?

@andrewwstephens Is it correct to default to XD for longer wavelengths?

andrewwstephens: While the XD filter transmission is 0.9 - 3.6um, the longest valid XD central wavelength is 2.5um.

rpiaggio: I changed the code so that it returns an error for unsupported wavelenghts.

toddburnside: That seems like a better idea to me. 😄

toddburnside (approved):
LGTM

PR #1377, opened 2026-04-30
SC-8459 visitor mode support in odb

tpolecat:
This adds support for visitor observing modes. The only notable change is that `ObservingMode` now has subtypes `FacilityObservingMode` and `VisitorObservingMode` so we can pick off visitor modes at the type level. All three have `Enumerated` instances so existing code continues to work.


modules/core/shared/src/main/scala/lucuma/core/data/Metadata.scala
```diff
@@ -23,20 +23,21 @@ object Metadata:
...
-          case Instrument.Gnirs     => Availability.always(Site.GN)
-          case Instrument.Gpi       => Availability.always(Site.GN)
-          case Instrument.Igrins2   => Availability.always(Site.GN)
-          case Instrument.Niri      => Availability.always(Site.GN)
-          case Instrument.Alopeke   => Availability.always(Site.GN)        
+          case Instrument.AcqCamNorth  => Availability.always(Site.GN)
+          case Instrument.Alopeke      => Availability.always(Site.GN)        
+          case Instrument.GmosNorth    => Availability.always(Site.GN)
+          case Instrument.Gnirs        => Availability.always(Site.GN)
+          case Instrument.Gpi          => Availability.always(Site.GN)
```
cquiroz, line 30: Gpi is going to be a visitor too, but it doesn't matter for this PR

modules/core/shared/src/main/scala/lucuma/core/enums/ObservingModeType.scala
```diff
@@ -7,12 +7,38 @@ import lucuma.core.math.Angle
...
+  case GmosNorthLongSlit  extends FacilityObservingModeType("gmos_north_long_slit",  Instrument.GmosNorth,  PosAngleConstraint.AverageParallactic)
+  case GmosSouthImaging   extends FacilityObservingModeType("gmos_south_imaging",    Instrument.GmosSouth,  PosAngleConstraint.Unbounded)
+  case GmosSouthLongSlit  extends FacilityObservingModeType("gmos_south_long_slit",  Instrument.GmosSouth,  PosAngleConstraint.AverageParallactic)
+  case GnirsLongSlit      extends FacilityObservingModeType("gnirs_long_slit",       Instrument.Gnirs,      PosAngleConstraint.AverageParallactic)
+  case Igrins2LongSlit    extends FacilityObservingModeType("igrins_2_long_slit",    Instrument.Igrins2,    PosAngleConstraint.AverageParallactic)
+
+enum VisitorObservingModeType(
+  val tag: String, 
+  val instrument: Instrument
+) extends ObservingModeType derives Enumerated:
```
cquiroz, line 37: Why visitors don't have a pos angle constraint?

tpolecat: Thus far it hasn't been necessary, but Andy did suggest a default so I'll add it back.

cquiroz (approved):
LGTM

PR #1378, opened 2026-04-30
Gnirs filters with range instead of cutoff

rpiaggio:
This model reflects the fact that there are gaps in the spectrum not covered by any filter. We will need downstream to validate when users configure custom wavelengths, and we can still choose the right filter with these values.


swalker2m (approved):
:+1:

PR #1382, opened 2026-05-03
Update ags magnitude limits

cquiroz:
limits are now probe dependent

### DIFF
--- a/modules/ags/src/main/scala/lucuma/ags/package.scala
+++ b/modules/ags/src/main/scala/lucuma/ags/package.scala
@@ -12,6 +12,7 @@ import lucuma.catalog.BandsList
 import lucuma.catalog.BrightnessConstraints
 import lucuma.catalog.FaintnessConstraint
 import lucuma.catalog.SaturationConstraint
+import lucuma.core.enums.GuideProbe
 import lucuma.core.enums.GuideSpeed
 import lucuma.core.enums.SkyBackground
 import lucuma.core.geom.offsets.GeometryType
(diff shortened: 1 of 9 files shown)


modules/ags/src/main/scala/lucuma/ags/package.scala
```diff
@@ -42,88 +43,149 @@ def wfsFwhm(sciFwhm: ImageQuality, wavelength: Wavelength): Double = {
...
+          }
+        case SkyBackground.Bright                       =>
+          guideSpeed match {
+            case GuideSpeed.Fast   =>
+              14.4 - 0.8 * wfsFwhm(iq, wavelength) - ce.toVegaMagnitude.toDouble
+            case GuideSpeed.Medium =>
+              14.7 - 0.8 * wfsFwhm(iq, wavelength) - ce.toVegaMagnitude.toDouble
+            case GuideSpeed.Slow   =>
+              16.0 - 0.8 * wfsFwhm(iq, wavelength) - ce.toVegaMagnitude.toDouble
+          }
```
swalker2m, line 174: It looks like you could get away with just looking up the fast, medium and slow values for the dark background and everything else can be computed in the same way from there.  Maybe it won't always be like that though.

swalker2m (approved):
Looks great.

modules/ags/src/main/scala/lucuma/ags/package.scala
cquiroz: Right, we have a document that indicates the coefficients like in the code, it could change eventually

PR #1384, opened 2026-05-06
GHOST package object

swalker2m:
Adds a GHOST package object to store GHOST-related constants, with the `CentralWavelength` being the only example for now.  Upon request, this will replace a similar value in the `odb` source.


cquiroz (approved):
LGTM

PR #1387, opened 2026-05-07
Gnirs enums adjustments

rpiaggio:
- Adjust Gnirs filters wavelengths (result of requirements refinement with Andy).
- Adjust some shortnames.
- Reenable invoking `sortUsablePositions` in AGS results.

### DIFF
--- a/modules/ags/src/main/scala/lucuma/ags/AgsAnalysis.scala
+++ b/modules/ags/src/main/scala/lucuma/ags/AgsAnalysis.scala
@@ -191,8 +191,4 @@ object AgsAnalysis {
       usablePerTarget.sorted(using Usable.rankOrdering)
     }
 
-  extension (results: AgsAnalysisResult)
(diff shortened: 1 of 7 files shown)


modules/ags/src/main/scala/lucuma/ags/AgsAnalysis.scala
```diff
@@ -191,4 +191,8 @@ object AgsAnalysis {
       usablePerTarget.sorted(using Usable.rankOrdering)
     }
 
+  extension (results: AgsAnalysisResult)
```
cquiroz: can this be just a methods in `AgsAnalysisResult`?

rpiaggio: Yes, good idea

cquiroz (approved):
LGTM

PR #1391, opened 2026-05-08
Gnirs grating names

rpiaggio:
Revert the GnirsGrating shortName since it's used when decoding the table. Adjust the longName instead.


cquiroz (approved):
LGTM

PR #1393, opened 2026-05-08
Add geometry for NS FPUs

cquiroz:
We have some observations in the database using Node and Shuffle and they break when calling AGS. In this pull request we add the geometry for N&S and allow ags to use them


rpiaggio (approved):
LGTM

PR #1394, opened 2026-05-08
Gnirs enum-inferring logic

rpiaggio:
Add logic to infer GnirsWellDepth from camera and GnirsReadMode from exposure time.

Also adjusted some errors en GnirsReadMode, changed naming to reflect current one in ocs, and converted time values to TimeSpans.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsReadMode.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsReadMode.scala
@@ -4,8 +4,10 @@
 package lucuma.core
 package enums
 
+import lucuma.core.syntax.timespan.*
 import lucuma.core.util.Display
 import lucuma.core.util.Enumerated
+import lucuma.core.util.TimeSpan
 
 /**
  * Enumerated type for GNIRS Read Mode.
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/GnirsReadMode.scala
```diff
@@ -15,13 +17,21 @@ enum GnirsReadMode(
...
+    case VeryBright extends GnirsReadMode("VeryBright", "Very bright", "Very Bright Acquisition or High Background", 200.msTimeSpan, 190.msTimeSpan, 155, 1)
+    case Bright extends GnirsReadMode("Bright", "Bright", "Bright objects", 600.msTimeSpan, 690.msTimeSpan, 30, 1)
+    case Faint extends GnirsReadMode("Faint", "Faint", "Faint objects", 9000.msTimeSpan, 11400.msTimeSpan, 10, 16)
+    case VeryFaint extends GnirsReadMode("VeryFaint", "Very faint", "Very faint objects", 18000.msTimeSpan, 22310.msTimeSpan, 7, 32)
+
+object GnirsReadMode:
+  // https://app.shortcut.com/lucuma/story/8557/set-gnirs-read-mode-based-on-exposure-time
+  def forExposureTime(t: TimeSpan): GnirsReadMode =
+    val exposureSeconds: BigDecimal = t.toSeconds
+    if (exposureSeconds < 0.6) GnirsReadMode.VeryBright
```
cquiroz, line 34: can you compare timespans directly?

rpiaggio: I did a quick attempt and it probably requires some magic import. What would we gain with that?

cquiroz: Mostly unit safety, it has hit us several time when we compare directly double/bigdecimal values.
But nevermind, it is internal to this method

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsReadMode.scala
rpiaggio: Whenever I use a non-specific type for a measure I put the units in the name. Still, I agree that in general it's better to use types.

PR #1399, opened 2026-05-12
Sc 8464 update workflow for visitor modes

tpolecat:
This adds a `DeclaredOngoing` exec state to support visitor modes.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/ExecutionState.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/ExecutionState.scala
@@ -40,6 +40,12 @@ enum ExecutionState(val tag: String) derives Enumerated:
    */
   case DeclaredOngoing extends ExecutionState("declared_ongoing")
 
-  /** Union type of declared (i.e., non-computed) states. */
-  type DeclaredExecutionState = DeclaredComplete.type | DeclaredOngoing.type
 
+/** Union type of declared (i.e., non-computed) states. */
+type DeclaredExecutionState = 
+  ExecutionState.DeclaredComplete.type | 
+  ExecutionState.DeclaredOngoing.type
+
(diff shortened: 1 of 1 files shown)


swalker2m (approved):
:+1:

modules/core/shared/src/main/scala/lucuma/core/enums/ExecutionState.scala
```diff
@@ -32,4 +32,20 @@ enum ExecutionState(val tag: String) derives Enumerated:
    * The observation has been explicitly declared complete by a user even
    * though more atoms and/or steps remain.
    */
-  case DeclaredComplete extends ExecutionState("declared_complete")
\ No newline at end of file
+  case DeclaredComplete extends ExecutionState("declared_complete")
+
+  /**
+   * The observation has been explicitly declared ongoing, though no atoms
```
cquiroz, line 38: Will the user declare an obs as ongoing?

tpolecat: Staff users can move the observation between `Ready` and `Ongoing` ("declared ongoing" internally), so we can do this to indicate that we're working on it outside of GPP. Then the PI or staff can set it to `Complete` ("declared complete" internally), as with other observations.

cquiroz: @andrewwstephens i wonder if we could auto complete looking at the archive? are we sending images to the archive?

andrewwstephens: Data from resident visitors go to the archive.  They use the VII to get the current active program ID from the TCS and write this to the FITS header.  Unfortunately the VII only provides the program ID, not the observation ID.

PR #1400, opened 2026-05-12
Shortcut 8595: GHOST IFU mapping

swalker2m:
This PR represents the current state of the IFU mapping problem.  It will likely be updated.  I would like to seek feedback from those who know GHOST and Observe.

Here in core, `Target` is basically `Sidereal | Nonsidereal | Opportunity` whereas over in the GraphQL API it means that `+` an ID `+` a backlink to the program `+` some other attributes. For that reason, when this is added to the API I think I'll need a new type, say `TargetLocation` that maps to what we mean by `Target` in `core`.   It's confusing and unfortunate.

At any rate, the plan is to _derive_ the IFU mapping from the asterism, an optional sky position, the resolution mode (standard vs high) and the position angle ... when possible.  If not possible the observation should have a validation error and not be ready to run.  In other words, the IFU mapping is an output and not directly specified in an input.

At some point, we can add mutations that allow the user to automatically obtain a sky position from a catalog lookup (in reverse) and have the position angle set automatically as well.  Initially though I'll just derive it from the other properties.

**The question at hand is this (@cquiroz and @jluhrs particularly): if you had this information suitably encoded in the ODB GraphQL schema would it be sufficient?**


modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostIfuMapping.scala
```diff
@@ -0,0 +1,74 @@
...
+  def mappingType: GhostIfuMappingType
+
+object GhostIfuMapping:
+
+  /**
+   * Only IFU1 is used for single target mode, and it is not restricted to
+   * sidereal targets.
+   */
+  case class SingleTarget(
+    ifu1: Target
```
cquiroz: I guess this will support non-sidereals

modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostStaticConfig.scala
```diff
@@ -8,9 +8,8 @@ import cats.derived.*
 import lucuma.core.enums.GhostResolutionMode
 import lucuma.core.util.TimeSpan
 
-// Will the target mode (single vs dual, which if any is sky, etc.) go here?
-
 final case class GhostStaticConfig(
   resolutionMode:                GhostResolutionMode,
+  ifuMapping:                    Option[GhostIfuMapping],
```
cquiroz: Why is it optional? don't we need always a mapping?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/sequence/ghost/GhostStaticConfig.scala
swalker2m: You're right, if there's no mapping then there's no static config and no steps or dynamic config for that matter.  I'll change it.

swalker2m:
I reworked this PR because I ran into problems with the corresponding GraphQL API.  Before the first visit the `ExecutionConfig` is computed and returned as a JSON object for Grackle to match against the query.  That won't work for the schema `Target` because it has a backlink to its `Program`.  In fact, the GraphQL target API doesn't match the `core` classes very well in general.  The best I could come up with was to add a `Nonsidereal` mapping which is understood to be a single target (IFU1) case that Observe will handle specially.  The others now have a `SiderealTracking` in each place where a `Target.Sidereal` appeared before.  `SiderealTracking` corresponds to the GraphQL `Sidereal` type, roughly (it is missing the catalog reference).

I believe this can appear in the API as:

```graphql
enum GhostIfuMappingType {
  NONSIDEREAL,
  SINGLE_TARGET,
  TARGET_PLUS_SKY,
  SKY_PLUS_TARGET,
  DUAL_TARGET
}

interface GhostIfuMapping {
  mappingType: GhostIfuMappingType!
}

type GhostNonsidereal extends GhostIfuMapping {
  mappingType: GhostIfuMappingType!
}

type GhostSingleTarget extends GhostIfuMapping {
  mappingType: GhostIfuMappingType!
  ifu1:        Sidereal!
}

type GhostTargetPlusSky extends GhostIfuMapping {
  mappingType: GhostIfuMappingType!
  ifu1:        Sidereal!
  ifu2:        Coordinates!
}

type GhostSkyPlusTarget extends GhostIfuMapping {
  mappingType: GhostIfuMappingType!
  ifu1:        Coordinates!
  ifu2:        Sidereal!
}

type GhostDualTarget extends GhostIfuMapping {
  mappingType: GhostIfuMappingType!
  ifu1:        Sidereal!
  ifu2:        Sidereal!
}

"""
GHOST configuration that applies across all steps.
"""
type GhostStatic {
  resolutionMode: GhostResolutionMode!
  ifuMapping: GhostIfuMapping!
  slitViewingCameraExposureTime: TimeSpan
}
```

cquiroz:
Will we represent that HR can only work with some of the ifu mappings?

swalker2m:
> Will we represent that HR can only work with some of the ifu mappings?

The idea is that the IFU mapping will be derived.  If it is an HR observation then we'd never derive `DualTarget`.  Which IFU mappings do apply to HR mode?

cquiroz:
I think HR can only do single/single+sky/nonsidereal

toddburnside:
> I think HR can only do single/single+sky/nonsidereal

The GHOST in GPP document says HR can only be single+sky. Not sure if it is correct or not.

cquiroz:
> > I think HR can only do single/single+sky/nonsidereal
> 
> The GHOST in GPP document says HR can only be single+sky. Not sure if it is correct or not.

Looking at the seqexec (may not correspond to reality) we have for HR target+sky/nonsidereal

jluhrs:
At the end Observe just need the resolution mode, targets mode, and the coordinates when needed (or the data to calculate them) All of that seems to be fulfilled.
Regarding HR mode, I already had planned to make Observe reject invalid combinations of resolution mode and targets modes. The observer will get an error message when trying to load an offending observation.

PR #1401, opened 2026-05-12
SC-8470 AGS params for visitor instruments

tpolecat:
I *think* this is all that's needed here. @cquiroz ?

### DIFF
--- a/modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
+++ b/modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
@@ -317,8 +317,8 @@ object AgsParams:
 
   case class Visitor private (
     scienceRadius: Angle,
-    port: PortDisposition,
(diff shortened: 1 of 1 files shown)


modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
```diff
@@ -314,3 +314,25 @@ object AgsParams:
 
     // Is this correct? is it the same for either ifu?
     val GhostScienceRadius = 20.arcseconds
+
+  case class Visitor private (
+    scienceRadius: Angle,
```
cquiroz: in the odb we are using scienceFov understood as the diameter

tpolecat: Aha, ok. That makes things easier.

cquiroz (approved):
LGTM

PR #1403, opened 2026-05-12
Enums for gnirs offsets

rpiaggio:


cquiroz:
Why is this needed? in other instruments we use plain offsets for offsets along the slit

rpiaggio:
> Why is this needed? in other instruments we use plain offsets for offsets along the slit

Gnirs supports both offsets along the slit, where the user can specify a list of Qs (like in GMOS); and also out-of-slit offsets (like F2), where the user can specify a list of (P, Q)s

rpiaggio:
image

> The Spatial Offsets UI allows pre-populating the offsets that are input into the sequence generator.  The options in this menu are different depending on the mode:
Slit Spectroscopy:  ABBA on-slit, ABBA off-slit
IFU Spectroscopy:  Grid-4 Small, Grid-4 Large, ABBA off-source
Imaging:   Grid, Spiral, Random, Import

from the spec at: https://app.shortcut.com/lucuma/epic/5735

cquiroz:
We did the same for igrins2 without an extra set of specific types.
I guess the question is whether this should be generalized rather than being Gnirs specific and we apply it to every case

rpiaggio:
> We did the same for igrins2 without an extra set of specific types. I guess the question is whether this should be generalized rather than being Gnirs specific and we apply it to every case

Good question. This model seems a bit tighter. In Igrins2 nothing prevents you from having a `NodAlongSlit` with `P`s <> 0. Also in Igrins2 (at least in the UI) `Nod along slit` seems to be restricted to 4 offsets and `Not to sky` to 3 offsets. In Gnirs, although those are the usual, according to Andy we want to leave it unrestricted.

cquiroz (approved):
LGTM

rpiaggio:
Ok the answer is that yes, we should unify:

image

toddburnside:
> Ok the answer is that yes, we should unify:

I guess for GMOS, at least, adding the off-slit positions would change the sequence generation. I'm not sure if IGRINS-2 sequence generation is depending on the number of offsets being limited.

rpiaggio:
Let's generalize. Closing in favor of https://github.com/gemini-hlsw/lucuma-core/pull/1405.

PR #1405, opened 2026-05-13
Generalize offset mode, more general Display derivation

rpiaggio:
I'm keeping the `tag` values so that it's easy to migrate downstream: it's just a type rename. PRs for odb and apps are already ready.


modules/core/shared/src/main/scala/lucuma/core/model/sequence/TelescopeConfigAlongSlit.scala
```diff
@@ -0,0 +1,32 @@
...
+
+import cats.Order
+import lucuma.core.enums.StepGuideState
+import lucuma.core.math.Offset
+import monocle.Focus
+import monocle.Lens
+
+case class TelescopeConfigAlongSlit(
+  offset:  Offset.Q,
+  guiding: StepGuideState
```
cquiroz, line 14: 👍

cquiroz (approved):
LGTM

PR #1407, opened 2026-05-15
Add missing GnirsLongSlit case in Configuration

rpiaggio:


modules/core/shared/src/main/scala/lucuma/core/model/Configuration.scala
```diff
@@ -90,11 +97,17 @@ object Configuration:
       val GmosImaging        = gmos.scienceArea.imaging.eval.radius
       val Igrins2LongSlit    = igrins2.scienceArea.scienceSlitFOV.eval.radius
 
+      def gnirsLongSlit(camera: GnirsCamera, prism: GnirsPrism): Angle = 
+        gnirs.all.slitLength(camera, prism).bisect
+
+        Angle.fromMicroarcseconds(gnirs.SlitLengthLongCamLxd.toMicroarcseconds / 2L) // SlitLengthLongCamLxd is the most conservative
```
cquiroz: Do we have to pick the largest or the shortest slit?

modules/core/shared/src/main/scala/lucuma/core/model/Configuration.scala
```diff
@@ -90,11 +97,17 @@ object Configuration:
       val GmosImaging        = gmos.scienceArea.imaging.eval.radius
       val Igrins2LongSlit    = igrins2.scienceArea.scienceSlitFOV.eval.radius
 
+      def gnirsLongSlit(camera: GnirsCamera, prism: GnirsPrism): Angle = 
+        gnirs.all.slitLength(camera, prism).bisect
```
cquiroz, line 101: Is this dead code?

modules/core/shared/src/main/scala/lucuma/core/model/Configuration.scala
rpiaggio: I forgot to remove a previous attempt (which I now removed), the actual implementation is line 101. With that implementation, we are actually choosing the slit length for the chosen configuration.

modules/core/shared/src/main/scala/lucuma/core/model/Configuration.scala
rpiaggio: Actually, the dead code was the following line. Removed it now.

cquiroz:
Why for the other modes we have a fixed geometry regardless of the parameters?

rpiaggio:
> Why for the other modes we have a fixed geometry regardless of the parameters?

That's a very good question. I'm not 100% clear on how this is used downstream. @tpolecat told me that:

> This is used to allow the user to move the base position around without needing approval. We limit this to the radius of the FOV, so you can move the base position anywhere that would be visible from the approved position. (edited)

and then in conversation with @andrewwstephens he told me that it would be ideal if we could derive this from `camera` and `prism`.

cquiroz:
> > Why for the other modes we have a fixed geometry regardless of the parameters?
> 
> That's a very good question. I'm not 100% clear on how this is used downstream. @tpolecat told me that:
> 
> > This is used to allow the user to move the base position around without needing approval. We limit this to the radius of the FOV, so you can move the base position anywhere that would be visible from the approved position. (edited)
> 
> and then in conversation with @andrewwstephens he told me that it would be ideal if we could derive this from `camera` and `prism`.

I think deriving is fine but it seems that configuration is picking one particular fov to represent the mode. I'd expect it to be either the largest or the smallest FoV but I'm not sure which

rpiaggio:
Yeah neither am I, so I went for the exact one. Maybe @tpolecat or  @andrewwstephens want to chime in?

cquiroz (approved):
LGTM

rpiaggio:
@tpolecat also said that:

> In principle the selected one should be part of the configuration and that should be the radius, and I suppose you shouldn't be allowed to switch to a longer slit without permission. But it may be okay to just use the radius of the longest one.
> Maybe ask Andy.

And Andy suggested we derive from `camera` and `prism`.

cquiroz:
> @tpolecat also said that:
> 
> > In principle the selected one should be part of the configuration and that should be the radius, and I suppose you shouldn't be allowed to switch to a longer slit without permission. But it may be okay to just use the radius of the longest one.
> > Maybe ask Andy.
> 
> And Andy suggested we derive from `camera` and `prism`.

Right even though derived from camera and prism we could pick the largest possible fov

rpiaggio:
I'll merge for now and we can modify the logic later if need be.

PR #1416, opened 2026-05-22
Add maroonx ags params

cquiroz:
MaroonX requires an extended area where we consider vignetting beyond the actual science fiber


toddburnside (approved):
LGTM. Happy version 200! 🎉

PR #1422, opened 2026-05-28
Remove GnirsObsReadMode enum

rpiaggio:
Instead of having a dedicated enum that just adds the `Automatic` option, let's follow the same pattern that Flamingos2 has where a `None` means automatic. This unifies the logic downstream.


cquiroz (approved):
LGTM

swalker2m (approved):

PR #1424, opened 2026-05-29
Correctly model Gnirs centralwavelength

rpiaggio:
It's not a configured value but rather the grating wavelength in spectroscopy or the filter one in imaging, just like for Gmos.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
@@ -21,7 +21,6 @@ import monocle.Lens
 final case class GnirsDynamicConfig(
   exposure:          TimeSpan,
   coadds:            PosInt,
-  centralWavelength: Wavelength,
   filter:            GnirsFilter,
   decker:            GnirsDecker,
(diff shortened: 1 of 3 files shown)


rpiaggio:
I'm merging to be able to continue the train. Post-reviews welcome.

modules/core/shared/src/main/scala/lucuma/core/model/sequence/gnirs/GnirsDynamicConfig.scala
```diff
@@ -21,15 +21,18 @@ import monocle.Lens
...
   filter:            GnirsFilter,
   decker:            GnirsDecker,
   fpu:               Either[GnirsFpuSlit, GnirsFpuOther],
   acquisitionMirror: GnirsAcquisitionMirrorMode,
   camera:            GnirsCamera,
   focus:             GnirsFocus,
   readMode:          GnirsReadMode
-) derives Eq
+) derives Eq:
+  val centralWavelength: Wavelength =
```
cquiroz, line 32: Don't we have a `CentralWavelength` newtype?

rpiaggio: It doesn't look like it.

PR #1426, opened 2026-05-29
Cache intersection shapes per posAngle

cquiroz:
Doing some tests, we can avoid recomputing some of the intersection shapes precalculating them per `posAngle`
This is more evident for obs with many offests as the shapes are recalculated over and over

### DIFF
--- a/modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
+++ b/modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
@@ -60,14 +60,31 @@ trait SingleProbeAgsParams:
   def posCalculations(
     positions: NonEmptyList[OffsetPosition]
   ): NonEmptyMap[OffsetPosition, AgsGeomCalc] =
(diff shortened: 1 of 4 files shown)


modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
```diff
@@ -60,14 +60,31 @@ trait SingleProbeAgsParams:
   def posCalculations(
     positions: NonEmptyList[OffsetPosition]
   ): NonEmptyMap[OffsetPosition, AgsGeomCalc] =
+    val distinctOffsets: NonEmptyList[(Offset, Offset)] =
+      positions.map(pos => (pos.offsetPos, pos.pivot)).distinct
+
+    // We can cache the intersection shapes for each tested pos angle.
+    val intersectionByPA: Map[Angle, (ShapeExpression, Shape, BoundingOffsets)] =
```
rpiaggio, line 67: Nice

rpiaggio (approved):
LGTM

PR #1434, opened 2026-06-03
Shortcut 8595: GHOST IFU Mapping (Again)

swalker2m:
The [initial battle plan](https://github.com/gemini-hlsw/lucuma-core/pull/1400) did not survive contact with the enemy.    We ended up needing more information than just the sidereal tracking (name, source profile, ...) so we decided to punt and just map to target ids.  Observe will read the IFU mapping along with the observation asterism and get the target details it needs from there.

I removed the explicit `Nonsidereal` mapping option.  The nonsidereal mapping will always be `SingleTarget` with a target id that refers to a nonsidereal target.  This unfortunately doesn't preclude, based on the API alone, the possibility of having dual nonsidereal or nonsidereal + sky mappings.  The ODB won't produce these though and Observe should just error out if one is nevertheless somehow encountered.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GhostIfuMappingType.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GhostIfuMappingType.scala
@@ -6,7 +6,6 @@ package lucuma.core.enums
 import lucuma.core.util.Enumerated
 
 enum GhostIfuMappingType(val tag: String, val name: String) derives Enumerated:
-  case Nonsidereal   extends GhostIfuMappingType("nonsidereal",     "Nonsidereal")
   case SingleTarget  extends GhostIfuMappingType("single_target",   "Single Target")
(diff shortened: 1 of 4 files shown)


toddburnside (approved):
Seems fine to me.

modules/core/shared/src/main/scala/lucuma/core/enums/GhostIfuMappingType.scala
```diff
@@ -6,7 +6,6 @@ package lucuma.core.enums
 import lucuma.core.util.Enumerated
 
 enum GhostIfuMappingType(val tag: String, val name: String) derives Enumerated:
-  case Nonsidereal   extends GhostIfuMappingType("nonsidereal",     "Nonsidereal")
```
cquiroz, line 9: How would observe know to do the non-sidereal setup?

toddburnside: It will have a non-sidereal target.

PR #1442, opened 2026-06-08
Add ImagingVariantType enum

cquiroz:


swalker2m (approved):

rpiaggio:
Is the idea for this to replace `GmosImagingVariantType`? Maybe we should remove or deprecate that one?

cquiroz:
Yes, I though the gmos one was on odb only but you're right it is in core too. I'll send another PR to remove it

PR #1444, opened 2026-06-08
Add digitalAverages to GnirsReadMode

rpiaggio:
The GnirsReadMode.digitalAverages is not in the OCS model, it's hard-coded in seqexec; but let's bring it here where I think it belongs besides the other values that depend on the ReadMode.


toddburnside (approved):
LGTM

PR #1446, opened 2026-06-08
Fix readouttime (typo?)

rpiaggio:
There was a slight discrepancy with https://github.com/gemini-hlsw/seqexec/blob/main/modules/server/src/main/scala/seqexec/server/gnirs/GnirsController.scala#L158, probably a typo.

Faint readout time per coadd is 11.14s, not 11.4.


cquiroz (approved):
LGTM

PR #1449, opened 2026-06-09
Add enum for cass rotator modes

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/enums/CassRotator.scala
```diff
@@ -0,0 +1,14 @@
...
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+/**
+ * Enumerated type for possible cass rotator modes.
+ */
+enum CassRotator(val tag: String) derives Enumerated:
```
rpiaggio: Will this be an option that users can select? If so, I suggest a `Display` instance.

rpiaggio (approved):
LGTM

PR #1452, opened 2026-06-11
Add parser and angle-related exports for JS

hugo-vrijswijk:
To be used in navigate-configs so it can reuse types from ODB


rpiaggio (approved):
LGTM

PR #1459, opened 2026-06-17
Fix automatic GNIRS acquisition filter

rpiaggio:
Although imaging, acquisition should use the spectroscopy J and K flters.

Also, rearranged the filter so that H2 is considered as candidate and added ranges for H2 and PAH filters.

Finally, made more precise types.

Companion PRs are sent to ODB and Explore to deal with breaking change.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
@@ -5,16 +5,12 @@ package lucuma
 package core
 package enums
 
-import cats.Order.given
-import cats.data.NonEmptyMap
-import cats.data.NonEmptySet
+import cats.data.NonEmptyList
 import cats.syntax.all.*
 import lucuma.core.math.BoundedInterval
 import lucuma.core.math.Wavelength
 import lucuma.core.util.Display
 import lucuma.core.util.Enumerated
-
-import scala.collection.immutable.SortedMap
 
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFilter.scala
```diff
@@ -30,13 +42,14 @@ enum GnirsFilter(
   case Order6 extends GnirsFilter("Order6", "X", "Order 6 (X)", 1_100_000.pm.some, (1_030_000, 1_175_400).pmRange.some)
   case Order5 extends GnirsFilter("Order5", "J", "Order 5 (J)", 1_270_000.pm.some, (1_175_400, 1_370_000).pmRange.some)
   case Order4 extends GnirsFilter("Order4", "H", "Order 4 (H: 1.645µm)", 1_645_000.pm.some, (1_490_000, 1_800_000).pmRange.some)
+  // Order matters for range match. Since H2 is completely contained in Order3, it needs to come before or it will never be selected.
```
cquiroz: Order based on what?

rpiaggio: For the declaration of the cases

cquiroz: I don't think we should rely on declaration order for logic

cquiroz (approved):
LGTM

PR #1460, opened 2026-06-17
Flaky tests on ScalaJS

rpiaggio:
Avoiding test suite failure on flaky tests depends on `munitFlakyOK` being set in the suite.

MUnit has a `munitFlakyOK` implementation that turns the flag on depending on the env variable `MUNIT_FLAKY_OK`. However, in Scala.js, there's no env variables and this mechanism always fails, resulting in failed flaky tests ruining the test run.

Previously, we had a custom tag (`ScalaCheckFlaky`). Tests with this tags were outright excluded in CI runs.

This PR seeks to provide a way that always honors flaky test declaration without the need for us to remember to use a custom tag.

Now, there's no need for a custom tag and the behaviour is:
- JVM: flaky tests are run but tolerated (as intended).
- JS: they are excluded.

The alternative, to run and tolerate JS tests, is to define our own base suite trait but then we have to remember to implement all tests suites extending that particular trait, which as a general mechanism sounds... well... flaky.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -113,8 +113,14 @@ lazy val testkit = crossProject(JVMPlatform, JSPlatform)
   .jvmConfigure(_.enablePlugins(AutomateHeaderPlugin))
 
 val MUnitFramework = new TestFramework("munit.Framework")
+// MUnit's built-in `.flaky` handling reads MUNIT_FLAKY_OK via System.getenv at test runtime.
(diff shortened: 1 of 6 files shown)


rpiaggio:
I'll actually add this into our automatically-applied `LucumaScalaJSPlugin`. We want this everywhere.

PR #1464, opened 2026-06-17
Bring ghost ifu mapping code from the odb

cquiroz:
copied directly from the odb, with just a rename of the package name


cquiroz:
We want to put this logic in core to reuse it in explore to calculate the ifu assignments locally rather than calling the odb

rpiaggio (approved):
LGTM

PR #1467, opened 2026-06-18
GnirsFpu enum, DaytimePinhole calibration role

rpiaggio:
Having an `enum GnirsFpu` makes downstream code much easier to understand than an `Either`. This is a breaking change and PRs for ODB and apps are lined up and will be sent right after this one.

Also, add `DaytimePinhole` calibration role.


modules/core/shared/src/main/scala/lucuma/core/enums/CalibrationRole.scala
```diff
@@ -10,3 +10,4 @@ enum CalibrationRole(val tag: String) derives Enumerated:
   case Photometric        extends CalibrationRole("photometric")
   case SpectroPhotometric extends CalibrationRole("spectrophotometric")
   case Telluric           extends CalibrationRole("telluric")
+  case DaytimePinhole     extends CalibrationRole("daytime_pinhole")
```
cquiroz, line 13: will these be automatically generated?

cquiroz:
This looks good, we may use the same logic for other similar fpus. I think gmos also uses an either but I could be confused

cquiroz (approved):
LGTM

PR #1469, opened 2026-06-19
AGS Benchmarks

cquiroz:
This PR adds a few benchmarks using jmh to test a few scenarios related to AGS.

It allows us to identify that the new JTS version gives a substantial performance improvement in some cases. For larger cases with e.g. 100 offsets we are having less of an improvement as performance scales with `n` (amount of offsets)


toddburnside (approved):
Seems fine to me.

PR #1480, opened 2026-06-25
Shortcut 9121: exchange observations

swalker2m:
Adds an `ExchangeObservingModeType` for observations destined for Keck and Subaru.  Not only are these observations not given an execution sequence, they don't participate in AGS or configuration requests.  The new `ExchangeObservingModeType` is patterned after the `VisitorObservingModeType` but I needed to make the `Instrument` optional since there is no Gemini instrument for these.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/probes.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/probes.scala
@@ -4,6 +4,7 @@
 package lucuma.core.model
 
 import cats.syntax.all.*
+import lucuma.core.enums.ExchangeObservingModeType
 import lucuma.core.enums.GuideProbe
 import lucuma.core.enums.ObservingModeType
 import lucuma.core.enums.TrackType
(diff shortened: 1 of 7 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/probes.scala
```diff
@@ -12,6 +13,9 @@ import lucuma.core.enums.VisitorObservingModeType
 trait probes:
   def guideProbe(observingMode: ObservingModeType, trackType: TrackType): Option[GuideProbe] =
     (observingMode, trackType) match
+      // Exchange observations are not supported by AGS; there is no guide probe.
+      case (_: ExchangeObservingModeType, _) =>
+        none
```
swalker2m, line 18: @cquiroz  Is this okay or will another shoe drop at some point?

cquiroz: i certainly don't expect ags to run an obs for other observatories

cquiroz (approved):
LGTM

swalker2m:
I added a bit after @cquiroz approved.

* replace the `obsolete` flag on `Instrument` with an `isVisitor` flag instead
* add GNIRS, Visitor and Exchange instances to `ArbInstrumentExecutionConfig`

We had no obsolete instruments and that's what the `Availability` information is supposed to tell us anyway.  Ideally I think we should have separate enums for facility instruments and visitor instruments.  Right now, the `Visitor` `InstrumentExecutionConfig` accepts any `Instrument` in its constructor and yet it is meant to be limited to visitor instruments.  The `isVisitor` flag is a poor man's substitute but won't trigger a bunch of refactoring.  I need it over in the ODB for the `InstrumentExecutionConfig` JSON codec.

cquiroz:
Seems fine, I agree that `Visitor` maybe a separate enum like we do for `ObservingMode`

toddburnside (approved):
I don't want to see what this does to the ODB, but looks fine to me.

PR #1481, opened 2026-06-26
Gnirs IFU FPu and decker implementations

rpiaggio:
Upcoming companion PRs upcoming to ODB and apps.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFpuIfu.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFpuIfu.scala
@@ -1,0 +1,22 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma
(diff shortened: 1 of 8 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFpuIfu.scala
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Enumerated type for GNIRS IFU FPU.
+ * @group Enumerations (Generated)
+ */
+enum GnirsFpuIfu(
+  val tag: String,
+  val shortName: String,
+  val longName: String,
+  val slitWidth: Angle,
+) derives Enumerated, Display:
```
cquiroz, line 20: is `Display` derived from `longName`?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsFpuIfu.scala
rpiaggio: From both `shortName` and `longName`.

PR #1483, opened 2026-06-29
AGS fixes for multi-targets

cquiroz:
It was noted that in GHOST with multiple target sometimes AGS would pick targets that would vignettes targets out of the base position.

The bug was we missed offsetting the targets out of the base, which was ok when they match but not for dual targets.


modules/catalog-tests/jvm/src/main/scala/lucuma/ags/ShortCut9024App.scala
```diff
@@ -0,0 +1,163 @@
...
+import lucuma.core.model.ImageQuality
+import lucuma.core.model.SiderealTracking
+import org.http4s.jdkhttpclient.JdkHttpClient
+import org.typelevel.log4cats.LoggerFactory
+import org.typelevel.log4cats.noop.NoOpFactory
+import org.typelevel.otel4s.trace.Tracer.Implicits.noop
+
+import java.time.Instant
+
+object ShortCut9024App extends IOApp.Simple {
```
toddburnside, line 44: What is the purpose of this app? Just running for a manual test?

toddburnside (approved):
LGTM

modules/catalog-tests/jvm/src/main/scala/lucuma/ags/ShortCut9024App.scala
cquiroz: Yes, it matches exactly a test case in explore so I can use it as a before and after sanity check.

PR #1487, opened 2026-06-30
Add outline for the GMOS masks

cquiroz:
We can get the outline of the MOS masks for gmos from gmmps (barcode included). Note the outline is slightly different at GN and GS

image

### DIFF
--- a/modules/tests/jvm/src/main/scala/lucuma/core/geom/jts/demo/JtsDemo.scala
+++ b/modules/tests/jvm/src/main/scala/lucuma/core/geom/jts/demo/JtsDemo.scala
@@ -96,6 +96,28 @@ trait GmosImagingShapes extends InstrumentShapes:
       candidatesArea.candidatesAreaAt(posAngle, offsetPos)
     )
 
+trait GmosMosShapes extends InstrumentShapes:
+  import lucuma.core.geom.gmos.{scienceArea, candidatesArea}
+  import lucuma.core.geom.gmos.oiwfs.{probeArm, patrolField}
+
+  val posAngle: Angle         = 0.deg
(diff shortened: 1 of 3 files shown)


modules/tests/jvm/src/main/scala/lucuma/core/geom/jts/demo/JtsDemo.scala
```diff
@@ -96,6 +96,28 @@ trait GmosImagingShapes extends InstrumentShapes:
...
+
+  val posAngle: Angle         = 0.deg
+  val offsetPos: Offset       = Offset(-60.arcsec.p, 60.arcsec.q)
+  val guideStarOffset: Offset = Offset(-140.arcsec.p, 160.arcsec.q)
+  val port: PortDisposition   = PortDisposition.Side
+
+  override def coloredShapes: List[ColoredShape] =
+    List(
+      ColoredShape(scienceArea.imagingMode.shapeAt(posAngle, offsetPos),   new Color(144, 238, 144)),
+      ColoredShape(scienceArea.mosModeSouth.shapeAt(posAngle, offsetPos), Color.cyan, Some(new BasicStroke(2f)))
```
swalker2m, line 111: We just picked south for the demo right?  Or did you mean to show north too?

swalker2m (approved):
Looks good.  Do we need this?

modules/tests/jvm/src/main/scala/lucuma/core/geom/jts/demo/JtsDemo.scala
cquiroz: just for the demo

PR #1496, opened 2026-07-07
Gnirs imaging + IFU AGS and imaging geometry

rpiaggio:
~For imaging the geometry width is 4.33 arcsecs, which in OCS is the default width for FPUs that don't define an explicit width. This is the case for the acquisition FPU, which is the one GNIRS uses for imaging.~

The actual keyhole geometry is implemented, see comment below.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometry.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometry.scala
@@ -46,23 +46,22 @@ trait GnirsScienceAreaGeometry:
       case GnirsFilter.Y | GnirsFilter.J | GnirsFilter.K => true
       case _                                             => false
 
-  private def asec(a: Angle): Double = Angle.signedDecimalArcseconds.get(a).toDouble
-  private def arcsec(d: Double): Angle = Angle.fromDoubleArcseconds(d)
   private def offP(pArcsec: Double): Offset =
     Offset.signedMicroarcseconds.reverseGet(((pArcsec * 1e6).round, 0L))
 
   // A circular cap (segment) protruding toward +p from the +p edge of a bar of the
-  // given p-width (centered on the origin). `chord` is the cap's extent along q, and
-  // `sagitta` its protrusion along p; the radius is implied by the two. The long field
-  // axis runs along q (like the long slit), so the cap bumps out sideways.
+  // given p-width (centered on the origin). `capWidth` is the cap's extent along q, and
+  // `capHeight` its protrusion along p; the radius is implied by the two (the standard
+  // chord/sagitta relation for a circular segment). The long field axis runs along q
+  // (like the long slit), so the cap bumps out sideways.
   // See https://www.gemini.edu/sciops/instruments/nirs/filters/imaging_aps.jpg
-  private def cap(chord: Angle, sagitta: Angle, barWidth: Angle): ShapeExpression =
-    val c       = asec(chord)
-    val s       = asec(sagitta)
-    val baseP   = asec(barWidth) / 2                   // cap springs from the +p edge of the bar
-    val r       = (s * s + (c / 2) * (c / 2)) / (2 * s) // circle radius from chord & sagitta
-    val circle  = ShapeExpression.centeredEllipse(arcsec(2 * r), arcsec(2 * r)) ↗ offP(baseP - (r - s))
-    val clip    = ShapeExpression.centeredRectangle(arcsec(2 * s + 2), arcsec(c + 2)) ↗ offP(baseP + s + 1)
+  private def cap(capWidth: Angle, capHeight: Angle, barWidth: Angle): ShapeExpression =
+    val w       = capWidth.toSignedDoubleDecimalArcseconds
+    val h       = capHeight.toSignedDoubleDecimalArcseconds
+    val baseP   = barWidth.toSignedDoubleDecimalArcseconds / 2 // cap springs from the +p edge of the bar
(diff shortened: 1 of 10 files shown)


rpiaggio:
I just got the correct shape. Converting to draft until that is implemented.

https://www.gemini.edu/sciops/instruments/nirs/filters/imaging_aps.jpg

rpiaggio:
This PR is now ready for review, implementing the keyhole geometries for GNIRS imaging:

image

modules/core/shared/src/main/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometry.scala
```diff
@@ -38,6 +39,44 @@ trait GnirsScienceAreaGeometry:
...
 
+  // Y-MK, J-MK and K-MK filters see the smaller, round unvignetted field; every other
+  // imaging filter (order-blocking, narrow-band, H-MK) sees the full keyhole.
+  def isRoundImagingFilter(filter: GnirsFilter): Boolean =
+    filter match
+      case GnirsFilter.Y | GnirsFilter.J | GnirsFilter.K => true
+      case _                                             => false
+
+  private def asec(a: Angle): Double = Angle.signedDecimalArcseconds.get(a).toDouble
+  private def arcsec(d: Double): Angle = Angle.fromDoubleArcseconds(d)
```
cquiroz: I'd think these may be defined somewhere else but in general we may want to use BigDecimal for calculations.

modules/core/shared/src/main/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometry.scala
```diff
@@ -38,6 +39,44 @@ trait GnirsScienceAreaGeometry:
...
+  private def offP(pArcsec: Double): Offset =
+    Offset.signedMicroarcseconds.reverseGet(((pArcsec * 1e6).round, 0L))
+
+  // A circular cap (segment) protruding toward +p from the +p edge of a bar of the
+  // given p-width (centered on the origin). `chord` is the cap's extent along q, and
+  // `sagitta` its protrusion along p; the radius is implied by the two. The long field
```
cquiroz: What is `sagitta`?

modules/tests/shared/src/test/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometrySuite.scala
```diff
@@ -0,0 +1,64 @@
...
+import lucuma.core.enums.GnirsCamera
+import lucuma.core.enums.GnirsFilter
+import lucuma.core.enums.GnirsFpuIfu
+import lucuma.core.geom.jts.interpreter.given
+import lucuma.core.math.Angle
+import lucuma.core.math.Offset
+
+class GnirsScienceAreaGeometrySuite extends munit.FunSuite:
+
+  private def sides(shape: lucuma.core.geom.ShapeExpression): (Angle, Angle) =
```
cquiroz: `fqn`?

cquiroz:
We have been adding jtsdemo implementations for each mode, could you include one and a screenshot?

rpiaggio:
> We have been adding jtsdemo implementations for each mode, could you include one and a screenshot?

Done!

image

rpiaggio:
And this is with a Y/J/K filter:

image

cquiroz (commented):
In general seems good, it is an odd shape, some comments inline.

modules/core/shared/src/main/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometry.scala
rpiaggio:
```
They're standard geometry terminology, and I chose them myself — they're not from the OCS source, the Gemini docs, or anywhere else in the codebase (the grep confirms GnirsScienceAreaGeometry.scala is the only place either word appears).

For a circular segment (the piece of a disk cut off by a straight line — which is exactly the cap shape bumping out of the keyhole bar), the classical names are:

- chord — the straight line cutting the circle; here, the cap's extent along q (where it springs from the bar's edge).
- sagitta — the perpendicular distance from the midpoint of the chord out to the arc; here, the cap's protrusion along p. The name is Latin for "arrow": picture the chord as a bowstring and the sagitta as the arrow drawn back from it. It's the standard term in that segment-geometry vocabulary (chord / sagitta / versine).

The relationship encoded in line 63,

r = (s² + (c/2)²) / (2s)

is the textbook formula for a circle's radius given a chord and its sagitta — that's the whole reason those two names go together. So they describe the shape accurately, but to be clear about provenance: I picked them as the conventional math names for that construction, not from any Gemini/GNIRS-specific reference. The imaging_aps.jpg link is only for the shape of the field, not the naming.

If you'd prefer names that map more directly onto the instrument's own axes (e.g. qExtent/pProtrusion or capWidth/capHeight), I'm happy to rename — the call sites already use KeyholeCapWidth/KeyholeCapHeight, so there's a slight vocabulary mismatch between the constants and the cap parameters worth reconciling.
```

I'm renaming to our more familiar terminology.

modules/core/shared/src/main/scala/lucuma/core/geom/gnirs/GnirsScienceAreaGeometry.scala
rpiaggio: They weren't. 2 of them are now. The offP is specific enough to leave it here.

rpiaggio:
I'm taking the chance to bundle other trivial unrelated changes in this PR, so as to speedup the release and the start of the train:
- Gnirs acquisition constants.
- Lamp shortnames.

cquiroz (approved):
LGTM

PR #1503, opened 2026-07-15
Decouple visitor science fov and ags diameter

cquiroz:
we'll have two fields for diameters for visitors, this removes the need for a special maroonx ags case


modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
```diff
@@ -55,9 +53,9 @@ trait SingleProbeAgsParams:
 
   def probeArm(posAngle: Angle, guideStar: Offset, offset: Offset): ShapeExpression
 
-  def scienceRadius: Angle
+  def scienceDiameter: Angle
```
cquiroz, line 56: This was correctly used as a diameter but named radius

toddburnside (approved):
LGTM

PR #1504, opened 2026-07-16
ProgramTypes and ProgramReferences for Keck and Subaru

toddburnside:
Due to a desire to reserve existing ProgramTypes and ProgramReferences for gemini programs, new instances of each were added for Keck and Subaru.


cquiroz (approved):
LGTM

PR #1510, opened 2026-07-18
Unify telescope offsets across instruments

cquiroz:
Generalizes how longslit telescope-offset presets are modeled so instruments can declare different preset sets, and adds Flamingos2 longslit support. Also brings IGRINS2 onto the same SlitTelescopeConfigs architecture as GNIRS.


toddburnside (approved):
LGTM

PR #1511, opened 2026-07-19
Min exposure time for very bright should be 0.2s

cquiroz:


toddburnside (approved):
Excellent. One I can understand! LGTM

cquiroz:
And one we are not sure is the correct fix 😄

PR #1512, opened 2026-07-19
Add GOA client

cquiroz:
This is more ore less a one-to-one port of the PIT client code for the gemini archive


toddburnside (approved):
LGTM

PR #1515, opened 2026-07-22
Add width for gnirs imaging filters

cquiroz:


toddburnside (approved):
LGTM after base version bump

PR #1519, opened 2026-07-23
Add test mixin that retries flaky tests in CI

hugo-vrijswijk:
Local behaviour is the same, but useful for DB-related tests and such


cquiroz (approved):
Seems useful, LGTM

PR #1523, opened 2026-07-27
Use RetryFlakyTests trait

hugo-vrijswijk:


cquiroz:
Are we sure all these tests are flaky because of external reasons? I'm a bit wary of marking tests as flaky where they are really masking a code problem. I think all that require external data should be legitimately flaky also the shape expression has some approximation problems, not sure about coordinates and bounded interval

hugo-vrijswijk:
> Are we sure all these tests are flaky because of external reasons? I'm a bit wary of marking tests as flaky where they are really masking a code problem. I think all that require external data should be legitimately flaky also the shape expression has some approximation problems, not sure about coordinates and bounded interval

I think the best way to do that would be to remove `.flaky` everywhere and only add it back if it can't be fixed properly

cquiroz:
I'd leave coordinates and boundedinterval non flaky and keep the others, we know horizons and target imports are flaky as they call remote services and shape has some numerical edge cases

PR #1525, opened 2026-07-27
Support pending masks

cquiroz:
for MOS we may actually define the mask in phase2 but before that it is valid to have a MOS observation that doesn't have a fully defined mask id yet.

This PR models that adding an option for a `ToBeDefined `mask without a name

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/MaskDefinition.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/MaskDefinition.scala
@@ -5,6 +5,9 @@ package lucuma.core.model
 
 import cats.Eq
 import cats.derived.*
+import monocle.Iso
+import monocle.Prism
+import monocle.macros.GenPrism
 
 /**
  * Observations with a mask to be defined are valid.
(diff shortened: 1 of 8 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/MaskDefinition.scala
```diff
@@ -0,0 +1,16 @@
...
+import cats.derived.*
+
+/**
+ * Observations with a mask to be defined are valid.
+ * We'll represent that state with the ADT below
+ */
+sealed trait MaskDefinition derives Eq
+
+case object ToBeDefined extends MaskDefinition derives Eq
+case class Defined(id: Attachment.Id) extends MaskDefinition derives Eq
```
swalker2m, line 19: I suppose we could add an `Iso[MaskDefinition, Option[Attachment.Id]]`.

swalker2m (approved):
Looks great. :+1:

modules/core/shared/src/main/scala/lucuma/core/model/MaskDefinition.scala
cquiroz: Sounds good

PR #1531, opened 2026-07-31
Tracing improvements

cquiroz:
Adds a method to create trace attributes from ags stats and fixes an edge cats for catalog tracing


toddburnside (approved):
LGTM

PR #1534, opened 2026-08-02
Some tiny details missing

cquiroz:


modules/tests/shared/src/test/scala/lucuma/core/math/CoordinatesSuite.scala
```diff
@@ -127,16 +127,24 @@ final class CoordinatesSuite extends munit.DisciplineSuite {
...
+      val nearAntipodal = (µas180 - sep.toMicroarcseconds).abs < minGapµas
+      // Skip test cases very close to 180
+      !nearAntipodal ==> {
+        val Δs = (Range.BigDecimal(-1.0, 2.0, 0.1)).map { f =>
+          val stepSep  = c1.interpolate(c2, f.toDouble).angularDistance(c1).toMicroarcseconds
+          val fracSep  = (sep.toMicroarcseconds * f.abs).toLong
+          val fracSepʹ = if (fracSep <= µas180) fracSep else µas360 - fracSep
+          (stepSep - fracSepʹ).abs
+        }
+        Δs.max <= 55L
```
toddburnside: Don't we need an `assert` here somewhere?

toddburnside (approved):
LGTM w/ question on test.

modules/tests/shared/src/test/scala/lucuma/core/math/CoordinatesSuite.scala
cquiroz: It is implicit as we return a Prop boolean which is evaluated, OTOH in other places where we use filters we have used `assert` and it is equally valid. I'll change it

PR #1538, opened 2026-08-05
Add GNIRS imaging Faint sky offset default

toddburnside:
Adds `GnirsAcquisitionMode.Faint.DefaultImagingSkyOffset` alongside the existing slit and IFU defaults, for the GNIRS imaging acquisition, whose Faint field image doubles as the sky frame.

`forTypeAndOffset` now requires the sky offset rather than defaulting to the slit value, since the correct default differs per observing mode and silently picking the slit one is wrong for IFU and imaging.

Claude-Session: https://claude.ai/code/session_01Lyt6Er1f7k4DDcNBoZEUGS


cquiroz (approved):
LGTM

PR #1539, opened 2026-08-05
ToO Activation Update

swalker2m:
Core updates required for [the corresponding update to the ODB](https://github.com/gemini-hlsw/lucuma-odb/pull/2808).

Adds an `Interrupting` ToO enum value, a validation code for requesting a ToO activation that is more disruptive than the approved most disruptive status, and standardizes the case of `TooActivation`.


toddburnside (approved):
LGTM. I guess lucuma-apps will need to be updated for the name change.

swalker2m:
> LGTM. I guess lucuma-apps will need to be updated for the name change.

I'll have to hold off merging until the ODB PR is reviewed of course.

PR #1547, opened 2026-08-12
Make GnirsDisperserOrder constructible

toddburnside:
Orders One and Two passed zero as their delta wavelength, and Wavelength
requires a positive value, so every value of the enum threw on construction:
even touching one failed with "Cannot build a Wavelength with value 0". That
is why the enum had no callers.

The delta wavelength is a per-pixel span, so it is now an
Option[WavelengthDelta], None for the two orders where it is undefined —
which are also the two that cannot be cross-dispersed.

Note that this Enum was unused. It will be used in explore in an upcoming PR.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDisperserOrder.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDisperserOrder.scala
@@ -5,12 +5,23 @@ package lucuma
 package core
 package enums
 import lucuma.core.math.Wavelength
+import lucuma.core.math.WavelengthDelta
 import lucuma.core.util.Display
 import lucuma.core.util.Enumerated
 
 
 /**
  * Enumerated type for GNIRS Disperser Order.
+ *
+ * `deltaWavelength` is the wavelength covered by one pixel with the 32 l/mm disperser at
+ * the 0.15"/pix pixel scale; the other combinations are derived from it (111 l/mm divides
+ * by 3.49, 10 l/mm multiplies by 3, and the 0.05"/pix scale divides by 3).  It is `None`
+ * for the two orders where it is undefined, which are also the two that cannot be
+ * cross-dispersed.
+ *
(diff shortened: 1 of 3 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDisperserOrder.scala
```diff
@@ -21,17 +32,16 @@ enum GnirsDisperserOrder(
   val defaultWavelength: Wavelength,
   val minWavelength:     Wavelength,
   val maxWavelength:     Wavelength,
-  val deltaWavelength:   Wavelength,
+  val deltaWavelength:   Option[WavelengthDelta],
```
swalker2m, line 35: I knew about `WavelengthDither` but not `WavelengthDelta`.  I see the difference but they are very similar...  I suppose this could also have been a non-optional `WavelengthDither` where `None` maps to `WavelengthDither.Zero`, not that it would be any better that way.  I wonder if we really need both concepts.

swalker2m (approved):
This looks great. :+1:  I was just musing on something outside of the scope of the PR.

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsDisperserOrder.scala
toddburnside: Well, so far we haven't needed either concept it seems, since the enum was unused. I'm guessing this was just moved from the OCS at some point.

toddburnside: Oh, and my explore PR only uses `minWavelength` and `maxWavelength`, not any of the other fields.

swalker2m: `WavelengthDither` is used in the ODB in sequence generation, for long slit at least.  In fact, it's part of the mode configuration.

PR #1548, opened 2026-08-12
UDF parsing

cquiroz:
Read-only support for parsing Gemini MOS mask ODF/MDF files

It includes a streaming fits reader api that can produce set of header units HDU and also tables. containing in mask information.
It also includes a higher level API for MOS mask files to read information like slits. Instrument position angle, etc. 

It is based on GMMPS thus masks produced that way should be handled by this lib. Tests files also come from gmmps


modules/catalog/shared/src/main/scala/lucuma/catalog/fits/FitsHeader.scala
```diff
@@ -0,0 +1,156 @@
...
+ */
+case class FitsHeader(cards: List[FitsHeader.Card]):
+
+  private lazy val byKeyword: Map[String, FitsHeader.Card] =
+    cards.foldLeft(Map.empty[String, FitsHeader.Card]): (m, c) =>
+      if m.contains(c.keyword) then m else m.updated(c.keyword, c)
+
+  /** The first card with this keyword, if any. */
+  def card(keyword: String): Option[FitsHeader.Card] =
+    byKeyword.get(keyword)
```
swalker2m, line 35: It has to create a `Map` of all the entries for each keyword lookup?  I wonder if a find on the list would be better, or maybe just keep up with a `cards` list and a key/value `Map`.

swalker2m, line 35: It has to create a `Map` of all the entries for each keyword lookup?  I wonder if a find on the list would be better, or maybe just keep up with a `cards` list and a key/value `Map`.

modules/catalog/shared/src/main/scala/lucuma/catalog/fits/FitsHeader.scala
```diff
@@ -0,0 +1,156 @@
...
+   *
+   * Returns `None` for the `END` card, for commentary keywords, and for blank padding — none of
+   * which carry a value.
+   */
+  def parseCard(record: String): Option[Card] =
+    val keyword = record.take(8).trim
+    if keyword.isEmpty || keyword === "END" then None
+    else if record.length < 10 || record.charAt(8) =!= '=' then
+      // COMMENT, HISTORY and other commentary keywords have no value indicator.
+      None
```
swalker2m, line 126: What if a COMMENT has an `=` character at position 8?

swalker2m (approved):
FITS is fun.  We're pulling this in from GMMPS?

cquiroz:
> FITS is fun. We're pulling this in from GMMPS?

Not exactly, the fits parsing part is generic but how to understand the mask is taken from gmmps, we also got the exact geometry directly from gmmps earlier

PR #1549, opened 2026-08-13
ToO Refactor

swalker2m:
This PR offers support for a corresponding ODB update, which in turn is working toward implementing [ToOs in GPP](https://gist.github.com/swalker2m/50ba9ff79a03a91edb5a9f8c8bc59351).  It combines two concepts, each of which could probably have been a stand-alone PR (apologies for that):

* **Scheduling Mode**. Replaces `ExecutionRequirement` with a `SchedulingMode`.  Here I'm trying to simplify ToO configuration relative to my initial attempt.  Before `ExecutionRequirement` and `TooActivation` were independently configurable and yet interacted in that some `TooActivation` options raised the floor on acceptable `ExecutionRequirements`.  Now, a PI only needs to set the `SchedulingMode`. The `TooActivation` is then _derived_ from the mode and the fact that a ToO `Opportunity` target exists in the asterism.
* **Forever Opportunity**. Reimagines `Opportunity` targets as persisting even after having been resolved.  That is, an `Opportunity` target marks the observation as a `ToO` and continues to be an `Opportunity` target even after the real celestial target is identified and associated.


modules/core/shared/src/main/scala/lucuma/core/enums/SchedulingMode.scala
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * The mode says nothing about urgency. *When* an observation must happen is a
+ * timing window question.
+ */
+enum SchedulingMode(
+  val tag:             String,
+  val isSplittable:    Boolean,
+  val isInterruptible: Boolean,
+  val mayInterrupt:    Boolean
+) derives Enumerated:
```
cquiroz, line 36: it seems there should be an order defined for power, but maybe we won't be using it anywhere?

toddburnside: It is ordered by power, isn't it?

toddburnside (approved):
Seems fine to me.

modules/core/shared/src/main/scala/lucuma/core/enums/SchedulingMode.scala
swalker2m: Yes, the natural enumeration order is progressively more powerful.

modules/core/shared/src/main/scala/lucuma/core/enums/SchedulingMode.scala
```diff
@@ -0,0 +1,59 @@
...
+ * `Uninterruptible`.
+ *
+ * The mode says nothing about urgency. *When* an observation must happen is a
+ * timing window question.
+ */
+enum SchedulingMode(
+  val tag:             String,
+  val isSplittable:    Boolean,
+  val isInterruptible: Boolean,
+  val mayInterrupt:    Boolean
```
rpiaggio, line 35: I suggest using `NewBoolean`s here.

modules/core/shared/src/main/scala/lucuma/core/enums/SchedulingMode.scala
rpiaggio, line 36: I'm guessing these values will be shown as a dropdown somewhere in a UI. If so, can we consider `derive Display` ?

modules/core/shared/src/main/scala/lucuma/core/enums/TooActivation.scala
```diff
@@ -8,7 +8,12 @@ import lucuma.core.util.Enumerated
...
  * both as a proposal-level ceiling (the maximum an observation in the program
- * may declare) and as the observation's own value.
+ * may reach) and as the observation's own value.
+ *
+ * An observation does not declare its activation. It is derived from whether the
+ * observation is a Target of Opportunity -- that is, whether its asterism holds
+ * an opportunity target -- together with its [[SchedulingMode]]; see
+ * [[TooActivation.fromSchedulingMode]].
  */
 enum TooActivation(val tag: String, val label: String) derives Enumerated:
```
rpiaggio, line 18: `derive Display` too ?

modules/core/shared/src/main/scala/lucuma/core/model/Target.scala
```diff
@@ -25,16 +25,46 @@ import scala.collection.immutable.SortedMap
...
+
+  /**
+   * This target as a sidereal target, if it tracks siderally -- whether because it is one, or
+   * because it is a Target of Opportunity that resolved to one.
+   *
+   * The result is a projection for computation (coordinates, ITC, guide stars) and not an
+   * identity: it has dropped the region and the fact that this was ever a Target of Opportunity,
+   * so it must not be persisted or used to decide whether a target is a ToO.
+   */
+  def asSidereal: Option[Target.Sidereal] =
```
rpiaggio, line 44: Is it worthy to return an opaque `ResolvedSidereal` type instead, to make it more difficult to perform an unwanted operation on it? (eg persistence)

rpiaggio (approved):
LGTM, with a couple of suggestions

swalker2m:
> LGTM, with a couple of suggestions

Thanks for all of the suggestions!  Let me come back to this when the dust settles.  Since this is the first of a chain of PRs that depend on it and which are finally all ready, I'm going to go ahead and merge this one.

PR #1550, opened 2026-08-13
Always find a GNIRS acquisition filter, and acquire on the blue camera

toddburnside:
A GNIRS observation at 5µm could not generate a sequence: it failed with "No Gnirs spectroscopy filter available for wavelength: 5000000". The acquisition filters cover 1.030-2.490µm plus PAH's 3.266-3.321µm, and fromAcquisitionWavelength required strict containment, so every M-band config and every L-band config outside the PAH band had no acquisition filter at all. Gaps at 1.370-1.490 and 1.800-1.910µm failed the same way — the latter reachable from a cross-dispersed config, whose LC SXD optimal wavelength is 1.85µm.

fromAcquisitionWavelength is now total: containment first, then ThermalAcquisitionFilter above the 2.5µm blue/red boundary, then the nearest broadband order filter for the remaining gaps. Acquisition must always be possible for a wavelength the instrument can observe.

Per the GNIRS instrument scientists, thermal-IR science is acquired with the blue camera at the same pixel scale (short and long are never mixed) in H, or in H2 when the target is very bright, which the existing brightness rule already handles. GnirsCamera gains blue and red accessors for that; red is not used yet but is what a manual override to acquire on the red camera will want. PAH drops out of automatic selection, since a 3.3µm filter has no place on the blue camera; it remains available as an explicit choice, and Explore's dropdown is unaffected because it uses AcquisitionFilters.

Returning a filter rather than an Either is binary incompatible, hence the tlBaseVersion bump.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsCamera.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GnirsCamera.scala
@@ -21,3 +21,18 @@ enum GnirsCamera(
   case LongRed extends GnirsCamera("LongRed", "LR", "Long Red", GnirsPixelScale.PixelScale_0_05)
   case ShortBlue extends GnirsCamera("ShortBlue", "SB", "Short Blue", GnirsPixelScale.PixelScale_0_15)
   case ShortRed extends GnirsCamera("ShortRed", "SR", "Short Red", GnirsPixelScale.PixelScale_0_15)
+
+  // The blue and red cameras come in pairs sharing a pixel scale, and short and long are
+  // never mixed within an observation, so changing colour always keeps the pixel scale.
+
+  /** The blue camera with this camera's pixel scale; itself, when already blue. */
+  def blue: GnirsCamera =
+    pixelScale match
+      case GnirsPixelScale.PixelScale_0_05 => GnirsCamera.LongBlue
+      case GnirsPixelScale.PixelScale_0_15 => GnirsCamera.ShortBlue
+
+  /** The red camera with this camera's pixel scale; itself, when already red. */
+  def red: GnirsCamera =
+    pixelScale match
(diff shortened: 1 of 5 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/GnirsCamera.scala
```diff
@@ -20,4 +20,19 @@ enum GnirsCamera(
...
   case ShortBlue extends GnirsCamera("ShortBlue", "SB", "Short Blue", GnirsPixelScale.PixelScale_0_15)
-  case ShortRed extends GnirsCamera("ShortRed", "SR", "Short Red", GnirsPixelScale.PixelScale_0_15)
\ No newline at end of file
+  case ShortRed extends GnirsCamera("ShortRed", "SR", "Short Red", GnirsPixelScale.PixelScale_0_15)
+
+  // The blue and red cameras come in pairs sharing a pixel scale, and short and long are
+  // never mixed within an observation, so changing colour always keeps the pixel scale.
+
+  /** The blue camera with this camera's pixel scale; itself, when already blue. */
+  def blue: GnirsCamera =
```
cquiroz, line 29: seems odd, like it were a new camera but i guess that's not the intent. is this what clients would use rather than the enums directly?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/enums/GnirsCamera.scala
toddburnside: Not exactly. The ODB will use this to get the blue version of the red camera at the same pixel scale. It needs to use the blue camera for acquisitions for red cameral science. But it can't mix long and short.

PR #1551, opened 2026-08-13
Revert the shortName changes from PR 1546

toddburnside:
The shortNames are used to parse the phase 0 tables in the odb. Create a manual Display instance, instead.


cquiroz (approved):
LGTM

PR #1557, opened 2026-08-19
Mos mask geometry

cquiroz:
With the mask information we can generate a geometry to be used for visualization and/or ags
The process was reversed engineered from gmmps and it is documented in an md file.

image

image

image


toddburnside (approved):
I don't really understand any of this, but it looks like scala, so 🤷

PR #1558, opened 2026-08-20
generic warning, for_review status

tpolecat:
This adds an observation validation code for warnings, and a `ForReview` workflow state. This new state will break things downstream so avoid updating until [the companion ODB PR](https://github.com/gemini-hlsw/lucuma-odb/pull/2872) is merged.


modules/core/shared/src/main/scala/lucuma/core/model/ObservationValidation.scala
```diff
@@ -34,3 +34,5 @@ object ObservationValidation:
     fromMsgs(ConfigurationRequestPending, ConfigurationRequestPending.description)
   def tooActivationUnapproved(msg: String, moreMsgs: String*): ObservationValidation =
     fromMsgs(TooActivationUnapproved, msg, moreMsgs*)
+  def genericWaning(msg: String, moreMsgs: String*): ObservationValidation =
```
toddburnside: Typo

toddburnside (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/model/ObservationValidation.scala
tpolecat: doh

PR #1563, opened 2026-08-22
GMOS IFU enums

rpiaggio:
Gnirs has separate enums for Slit and IFU FPUs and then unifies them for the DynamicConfig. I would like to mimic that design throughout but the migration impact is huge by this point.

Therefore, I'm going for tight specific types for IFU FPUs that can be used in the observing modes and map back to the general `Gmos*Ifu` instances.

`GmosIfuAnalysis` is a new value that needs to be defined in the observing mode.


modules/core/shared/src/main/scala/lucuma/core/geom/gmos/GmosScienceAreaGeometry.scala
```diff
@@ -70,6 +71,55 @@ trait GmosScienceAreaGeometry {
...
+
+  /**
+   * The sky lenslet field, half as wide as the target field because one pseudo-slit samples half of
+   * each. It does not sit on the base: see [[ifuSkyOffset]].
+   */
+  def ifuSkyFov(fieldWidth: Angle, site: Site): ShapeExpression =
+    ShapeExpression.centeredRectangle(ifuSkyFieldWidth(fieldWidth), IfuFieldHeight) ↗
+      ifuSkyOffset(fieldWidth, site)
+
+  def ifuSkyFieldWidth(fieldWidth: Angle): Angle =
```
cquiroz: Seem this could be inlined

modules/core/shared/src/main/scala/lucuma/core/model/GmosIfuAnalysis.scala
```diff
@@ -0,0 +1,48 @@
...
+import monocle.Focus
+import monocle.Iso
+import monocle.Prism
+import monocle.macros.GenPrism
+
+/**
+ * How to sample a GMOS IFU field. The number of fibres on sky is deliberately not part of this: it
+ * follows from the focal plane unit, so the sampling geometry is all the caller chooses.
+ */
+enum GmosIfuAnalysis:
```
cquiroz, line 17: lucuma-itc has the analysis method defined for the legacy ocs code and it doesn't match this shape. Why the difference? why should it be in lucuma-core and not in itc?

modules/tests/shared/src/test/scala/lucuma/core/geom/gmos/GmosScienceAreaGeometrySuite.scala
```diff
@@ -0,0 +1,67 @@
...
+import lucuma.core.enums.GmosSouthIfuFpu
+import lucuma.core.enums.Site
+import lucuma.core.geom.ShapeExpression
+import lucuma.core.geom.jts.interpreter.given
+import lucuma.core.math.Angle
+import lucuma.core.math.Offset
+
+class GmosScienceAreaGeometrySuite extends munit.FunSuite:
+
+  private def sides(shape: ShapeExpression): (Angle, Angle) =
```
cquiroz: I think this method exist already in the codebase

modules/core/shared/src/main/scala/lucuma/core/model/GmosIfuAnalysis.scala
rpiaggio: The user will have to specify this and the ~natural~ only possible place is to put it on the ObservingMode. For that, it needs to be seen from ODB and Explore and so the decision to put it here in core.

ITC will then translate this and other information into the legacy invocation. The number of fibers that the legacy shape uses can be inferred from the FPU, that's why the shape is different. This one has only the information from the radio buttons in the web ITC:

image

The middle option will not be supported, that's why it's missing.

modules/tests/shared/src/test/scala/lucuma/core/geom/gmos/GmosScienceAreaGeometrySuite.scala
rpiaggio: Good catch. It was already defined for Gnirs. Unified now.

swalker2m (approved):
Assuming Carlos approves as well, :+1:

cquiroz (approved):
Seems quite a few different set of changes conflated in one PR but it's ok

PR #1571, opened 2026-08-27
GMOS IFU offset presets

rpiaggio:
https://app.shortcut.com/lucuma/story/10045/gmos-ifu-offset-presets#activity-10061


toddburnside (approved):
LGTM

PR #1573, opened 2026-08-27
Add ProgramStatus enum

cquiroz:


build.sbt
```diff
@@ -2,7 +2,7 @@ import org.scalajs.linker.interface.ESVersion
 import org.typelevel.sbt.gha.PermissionValue
 import org.typelevel.sbt.gha.Permissions
 
-ThisBuild / tlBaseVersion                         := "0.238"
+ThisBuild / tlBaseVersion                         := "0.239"
```
toddburnside, line 5: Does this really need a base version jump? You're just adding an enum.

modules/core/shared/src/main/scala/lucuma/core/enums/ProgramStatus.scala
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Enumerated
+
+enum ProgramStatus(val tag: String, val name: String) derives Enumerated:
```
toddburnside: Should we derive Display as well?

toddburnside (approved):
LGTM

build.sbt
cquiroz: ci complained otherwise

modules/core/shared/src/main/scala/lucuma/core/enums/ProgramStatus.scala
cquiroz: yeah, i'll add it

build.sbt
toddburnside: Odd. 🤷

swalker2m:
There was no description, but it seems like a useful idea.  What is the plan for this?

cquiroz:
It was requested that programs will get a status flag, it will be only setteable by staff and used by the scheduler

PR #1575, opened 2026-08-28
Add ProposalSubmissionError

toddburnside:
Proposal submission is validated twice: in Explore, so that a submission is not attempted while it would fail, and again in the ODB, which is the authority.  The two rule sets had drifted -- each enforced rules the other did not, and two were implemented with different semantics.

Enumerate the rules here so both sides describe the same thing, and so that adding a rule is visible to both.  Each case carries the canonical message; consumers may decorate it (the ODB appends the program id) but should not restate the rule in their own words.


swalker2m (commented):
I approve posthumously.  Seems like a good idea.

PR #1576, opened 2026-08-29
Add GmosIfuAcquisitionRoi

rpiaggio:
A new enum is needed for GMOS IFU Acquisition ROI options: https://app.shortcut.com/lucuma/story/10044/gmos-ifu-acquisition#activity-10180

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/enums/GmosIfuAcquisitionRoi.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/enums/GmosIfuAcquisitionRoi.scala
@@ -1,0 +1,57 @@
+// Copyright (c) 2016-2025 Association of Universities for Research in Astronomy, Inc. (AURA)
+// For license information see LICENSE or https://opensource.org/licenses/BSD-3-Clause
+
+package lucuma.core.enums
+
+import lucuma.core.util.Display
+import lucuma.core.util.Enumerated
+
+/**
+ * GMOS IFU acquisition ROI options.
+ *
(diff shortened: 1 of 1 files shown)


modules/core/shared/src/main/scala/lucuma/core/enums/GmosIfuAcquisitionRoi.scala
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * Unlike the long slit these widen rather than narrow: the image through the IFU is always Full
+ * Frame, because both bundles have to be visible and the sky bundle sits 60" from the target.
+ *
+ * @param tag database tag
+ * @param name display name
+ * @param imagingRoi ROI to use for the imaging step
+ * @param ifuRoi ROI to use for the steps taken through the IFU
+ */
+enum GmosIfuAcquisitionRoi(
```
cquiroz, line 20: In ocs ROI was common for all gmos modes, why not in gpp?

cquiroz: I suppose that would limit assiging a roi not meant for a mode?

rpiaggio: > In ocs ROI was common for all gmos modes, why not in gpp?

The ROI enum is common for all modes; but the combinations available for acquisition steps differ for slit and IFU.

PR #1578, opened 2026-08-31
GMOS IFU AGS support

rpiaggio:


modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
```diff
@@ -201,6 +204,66 @@ object AgsParams:
...
+    private def fieldWidth: Angle = fpu.fold(_.fieldWidth, _.fieldWidth)
+
+    override def patrolFieldAt(
+      posAngle: Angle,
+      offset:   Offset,
+      pivot:    Offset = Offset.Zero
+    ): ShapeExpression =
+      probe match
+        case GuideProbe.GmosOIWFS =>
+          oiwfs.patrolField.longSlitMode.patrolFieldAt(posAngle, offset, gmosFpu, port, pivot)
```
cquiroz: So this is the same as longslit  but the field width is for the IFU area?

rpiaggio: Yes, that's the case. Only longslit allows specifying an FPU. I left the naming to preserve compatibility downstream, but I realize now this can get confusing. I'll rename.

cquiroz (approved):
LGTM

PR #1579, opened 2026-08-31
Flip GMOS IFU offset signs

rpiaggio:
The signs for the GMOS IFU FPU `xOffset`s seemed to be flipped WRT OCS.

OCS computes them rather than storing literals, I defer to Claude for details:

→ The arithmetic, North — getWFSOffset(), returns negative:

https://github.com/gemini-hlsw/ocs/blob/main/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosNorthType.java#L446

```
double w = IFU_FOV[IFU_FOV_LARGER_RECT_INDEX].width / 2.0;    // 7.0 / 2 = 3.5
double offset = -IFU_FOV_OFFSET - w;                          // -30 - 3.5 = -33.5
if (this == IFU_2)      offset += w / 2.0;                    // blue  -> -31.75
else if (this == IFU_3) offset -= w / 2.0;                    // red   -> -35.25
```

→ The arithmetic, South — returns positive, and uses the smaller rectangle:

https://github.com/gemini-hlsw/ocs/blob/main/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosSouthType.java#L471

```
double w = IFU_FOV[IFU_FOV_SMALLER_RECT_INDEX].width / 2.0;   // 3.5 / 2 = 1.75
double offset = IFU_FOV_OFFSET + w;                           // 30 + 1.75 = 31.75
if (IFU_2 || IFU_N_B) offset -= w / 2.0;                      // blue  -> 30.875
else if (IFU_3 || IFU_N_R) offset += w / 2.0;                 // red   -> 32.625
```

→ The constants both depend on:

https://github.com/gemini-hlsw/ocs/blob/main/bundle/edu.gemini.pot/src/main/java/edu/gemini/spModel/gemini/gmos/GmosCommonType.java#L65-L79

```
IFU_FOV_OFFSET = 30.
IFU_FOV[0] = 3.5″ // wide (sky); IFU_FOV_SMALLER_RECT_INDEX = 0.
IFU_FOV[1] = 7.0″ // wide (target); IFU_FOV_LARGER_RECT_INDEX = 1
```


swalker2m (approved):
I don't know which is correct, but I'll take it you've verified these.  One confusing part of OCS code is that it uses screen coordinates in a lot of places which are flipped wrt to both RA and Dec.

rpiaggio:
> I don't know which is correct, but I'll take it you've verified these. One confusing part of OCS code is that it uses screen coordinates in a lot of places which are flipped wrt to both RA and Dec.

That's a piece of information I didn't have. I pushed Claude to investigate whereas this is the case with these coordinates in the OCS and the answer is not nice:

---

I found a real coordinate-frame problem — and it's worse than a sign choice: OCS itself applies this offset with opposite p-signs in its two consumers, so no value of xOffset makes both correct.

→ OCS PatrolField shapes really are in a flipped frame. PatrolField.checkBoundaries says so outright at PatrolField.java:119-123: "Get offset and switch it to be defined in the same coordinate system as the shape", then p = -dis.p(), q = -dis.q(). Same as FeatureGeometry.toPoint = Point2D(-p, -q). So shape x = −p.

→ In the patrol field, OCS applies the IFU offset in that flipped frame. GmosOiwfsGuideProbe.java:133 does getTranslateInstance(ifuXOffset, 0.0) on the shape, i.e. −getWFSOffset() in p.

→ In the probe arm, OCS applies it in sky coordinates. GmosOiwfsProbeArm.scala:80 builds Offset(ifu.arcsecs[OffsetP], OffsetQ.Zero) — explicitly OffsetP, not shape x — i.e. +getWFSOffset() in p.

→ lucuma-core applies the same sign in both places, because both go through the one ifuOffset helper. So it can match one OCS site or the other, never both. My flip makes the probe arm match OCS and the patrol field disagree; before the flip it was the reverse. Neither state is fully right.

→ Both lucuma ports are otherwise faithful, which is how I'm confident the frames are what I say. The patrol-field rectangle is a clean port: OCS's Rectangle2D(-11.4, -34.92, 212.7, 249.6) centres at (94.95, 89.88) in shape coords, and lucuma writes exactly −Offset(94950.mas.p, 89880.mas.q) — the flip already applied. The probe arm is line-for-line: same T = (427.52 p, 101.84 q), same (offset − ifuOffset).rotate(posAngle), and lucuma's atan2(x, -y) after negating p equals OCS's atan2(-p, -q) once you unfold toPoint.

→ Physically the two must shift together, since the patrol field is where the probe can reach and the probe arm is where it is. Shifting them oppositely puts them ~64″ apart for the IFU. That's a strong hint one of OCS's two sites is a long-standing bug — but I cannot tell you which one is astronomically right from the code alone. That needs an astronomer or a visual check against the TPE.

rpiaggio:
Ok, upon closer inspection there's no bug in OCS: the offsets have to cancel each other. I'm adjusting the logic in this PR to mimic what OCS does.

rpiaggio:
Point 3 in https://app.shortcut.com/lucuma/story/10048/gmos-ifu-aladin-overlay#activity-10236 confirms this fix as correct, merging.

PR #1580, opened 2026-08-31
Add TelluricType.None, a PI can use it to request no telluric generation

cquiroz:


modules/core/shared/src/main/scala/lucuma/core/model/TelluricType.scala
```diff
@@ -21,6 +21,7 @@ object TelluricType:
   case object A0V                                     extends TelluricType("A0V")
   case object Solar                                   extends TelluricType("Solar")
   case class  Manual(starTypes: NonEmptyList[String]) extends TelluricType("Manual") derives Eq
+  case object None                                    extends TelluricType("None")
```
rpiaggio: I suggest renaming to `NoTelluric` or similar, to avoid confusion with `Option.None`.

rpiaggio (approved):
LGTM if renamed

modules/core/shared/src/main/scala/lucuma/core/model/TelluricType.scala
cquiroz: Good idea

PR #1581, opened 2026-08-31
sc-10025 remove `for_review` workflow state

tpolecat:
We decided to get rid of `for_review` in favor of just blocking the transition to `ready` when warnings exist, and giving staff a way to dismiss classes of warnings on a per-program basis. This is step 1 of this change.

After this change the transition set will be accurate (no "secret" transition available to staff users) so that special case goes away in the UI.


toddburnside (approved):
Sounds simpler, I guess. I had not yet added that special case to explore so there should be nothing to change there.

PR #1583, opened 2026-09-01
Telluric type encoder fix

cquiroz:


modules/catalog/shared/src/main/scala/lucuma/catalog/telluric/codecs.scala
```diff
@@ -27,8 +27,8 @@ object codecs:
   given Encoder[TelluricType] =
     Encoder[String].contramap:
       case TelluricType.Hot               => "hot"
-      case TelluricType.A0V               => "A0V"
-      case TelluricType.Solar             => "Solar"
+      case TelluricType.A0V               => "A0"
```
rpiaggio, line 30: Was removing the `V` intentional? Not sure what this PR fixes.

modules/catalog/shared/src/main/scala/lucuma/catalog/telluric/codecs.scala
```diff
@@ -27,8 +27,8 @@ object codecs:
...
+      case TelluricType.Solar             => "solar"
       case TelluricType.NoTelluric        => "NoTelluric"
```
rpiaggio, line 32: Do we want to keep UpperCamel for NoTelluric?

rpiaggio (approved):
LGTM

modules/catalog/shared/src/main/scala/lucuma/catalog/telluric/codecs.scala
cquiroz: yes, there is a mismatch with the remote value used

modules/catalog/shared/src/main/scala/lucuma/catalog/telluric/codecs.scala
cquiroz: It doesn't matter, `NoTelluric` by definition will never be sent to to the backend

PR #1586, opened 2026-09-01
Don't assign a default SED on CSV target import

toddburnside:
The CSV target importer has no SED column, but the surface-brightness branch of TargetCsvRow.sourceProfile hard-coded
StellarLibrary(O5V) while the integrated branch used None. Targets imported with surface brightness units therefore came back with a Stellar Library SED they never asked for.

Use None in both branches, and add a regression test built from a reported file of galaxies mixing surface and integrated units.


modules/catalog/shared/src/main/scala/lucuma/catalog/csv/TargetImport.scala
```diff
@@ -86,7 +84,7 @@ private case class TargetCsvRow(
     if (surfaceUnits.nonEmpty)
       SourceProfile.Uniform(
         SpectralDefinition.BandNormalized(
-          UnnormalizedSED.StellarLibrary(StellarLibrarySpectrum.O5V).some,
+          None,
```
cquiroz, line 87: Should we call simbad to infer the SED?

toddburnside: We do that if only the name is provided.

cquiroz: so if the name is in the csv we'd query simbad and get SED if not, it will be unset rather than assigned arbitrarily to 05V?

toddburnside: If they are providing brightnesses, etc., we do not query simbad and the user will need to manually set the SED. The arbitrary StellarLibrary(05V) is often wrong. Andy requested the SED be set to None, so that's what I did. I think we should stick with that until `they` request something else.

cquiroz (approved):
LGTM

PR #1588, opened 2026-09-02
SC-10025 break validation errors and warnings into distinct enums

tpolecat:
This is a mostly mechanical change. `ObservationValidationCode` exists as before and is source-compatible, but there are now also sub-enums `ObservationValidationCode.Error` and `ObservationValidationCode.Warning` for code that needs to distinguish them.


cquiroz (approved):
LGTM

PR #1591, opened 2026-09-03
Fixes the warnings for unary conversions

cquiroz:
scala 3.9.0 added a warning about negative literals that breaks lucuma-core, see https://github.com/scala/scala3/pull/24163


toddburnside (approved):
LGTM

PR #1592, opened 2026-09-03
Interval seq union all

swalker2m:
Adds a more efficient `unionAll` method to our `IntervalSeq.type` syntax and uses it where before we were doing a fold (with much repetitive copying) over a list of `IntervalSeq`.


toddburnside (approved):
LGTM

PR #1597, opened 2026-09-04
Change Gmos IFU offset default, renames

rpiaggio:
A couple of renames and changing the default offset for GMOS IFU.

Downstream changes are breaking but minimal, I'll send PRs as soon as this is published.


toddburnside (approved):
LGTM

PR #1607, opened 2026-09-08
Probe rules

cquiroz:
These rules were defined on https://github.com/gemini-hlsw/lucuma-odb/pull/3001 but we'll need them in explore


rpiaggio (approved):
LGTM

PR #1610, opened 2026-09-08
Require a complete profile of every investigator

toddburnside:
Only the PI had to have an email address, and only the partner link of
each investigator had to be set.  Nothing required a co-investigator to
have an email, and nothing required anyone to have a name, an
educational status or an institutional affiliation.

Generalize the two PI email rules to every investigator -- the PI, the
co-investigators and the read-only co-investigators alike -- and add the
three that were missing.  The new affiliation rule is about the
free-text institutional affiliation, which nothing validated before; the
partner link keeps its own rule, whose message already says
"partnership" rather than "affiliation".

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import org.scalajs.linker.interface.ESVersion
 import org.typelevel.sbt.gha.PermissionValue
 import org.typelevel.sbt.gha.Permissions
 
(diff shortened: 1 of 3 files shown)


cquiroz (approved):
LGTM

PR #1611, opened 2026-09-09
Replaces recursion with a loop when evaluating a chain of geometric operations

cquiroz:
In explore, the AGS worker failed in Chrome with `RangeError: Maximum call stack size exceeded` at about 200 science offsets. `AgsParams.posCalculations` combines the per-offset patrol fields with `reduce(using _ ∩ _)`, giving a left-nested tree with one Intersection node per offset.

In `toGeometry`, `case Intersection(a, b) => toGeometry(a).intersection(...)` recurses into the left child before computing anything, so stack depth grows with the offset count.

Fix

`ShapeExpression.leftSpine` collapses a run of the same operation into a flat list of operands (sort of replacing recursion with a loop). It follows left children from the root and stops at the first node of a different kind, so nesting of a different operator stays a single operand.

```scala
Intersection(Intersection(Intersection(pf1, pf2), pf3), pf4)
  -> (Intersection, List(pf1, pf2, pf3, pf4))
```

Tested it in explore and the error disappears, tested with up to 500 offsets, it is still very slow but at least it converges

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import org.scalajs.linker.interface.ESVersion
 import org.typelevel.sbt.gha.PermissionValue
 import org.typelevel.sbt.gha.Permissions
 
-ThisBuild / tlBaseVersion                         := "0.248"
+ThisBuild / tlBaseVersion                         := "0.247"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
--- a/build.sbt
+++ b/build.sbt
@@ -2,7 +2,7 @@ import org.scalajs.linker.interface.ESVersion
 import org.typelevel.sbt.gha.PermissionValue
 import org.typelevel.sbt.gha.Permissions
 
-ThisBuild / tlBaseVersion                         := "0.247"
+ThisBuild / tlBaseVersion                         := "0.248"
 ThisBuild / tlCiReleaseBranches                   := Seq("master")
 ThisBuild / githubWorkflowEnv += "MUNIT_FLAKY_OK" -> "true"
 
--- a/modules/core/shared/src/main/scala/lucuma/core/geom/ShapeExpression.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/geom/ShapeExpression.scala
@@ -3,6 +3,8 @@
 
 package lucuma.core.geom
 
+import cats.Eq
+import cats.syntax.eq.*
 import lucuma.core.math.Angle
 import lucuma.core.math.Offset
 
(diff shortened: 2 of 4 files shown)


rpiaggio:
It caught my attention that we don't go `@tailrec`, I asked AI and this is it's review:

---

Yes, `@tailrec` does the same job with no vars or the descending flag, and I'd push for it. Scala compiles it to the exact same loop, so stack safety is identical, and it reads like the rest of the codebase.

```
private[geom] def leftSpine(e: ShapeExpression, op: BinaryOp): List[ShapeExpression] =
  @tailrec
  def go(cur: ShapeExpression, acc: List[ShapeExpression]): List[ShapeExpression] =
    cur match
      case Difference(a, b)   if op == BinaryOp.Difference   => go(a, b :: acc)
      case Intersection(a, b) if op == BinaryOp.Intersection => go(a, b :: acc)
      case Union(a, b)        if op == BinaryOp.Union        => go(a, b :: acc)
      case _                                                 => cur :: acc
  go(e, Nil)
```

→ The `Eq[BinaryOp]` + `===` is noise. A Scala 3 enum already gets `CanEqual`, so plain `==` is safe even under strict equality. Or drop the enum entirely and pass a splitter, `ShapeExpression => Option[(ShapeExpression, ShapeExpression)]`, then go is one split(cur) match with no per-iteration op check.

→ Correctness is fine. Operands come out as [leftmost-leaf, b1 … bn] in evaluation order, and foldLeft reproduces ((a - b) - c) - d, so Difference (non-associative) is preserved. head/tail in foldChain are safe since the list is never empty, but returning (head, rest) from leftSpine would make that explicit and avoid partial calls.

→ It only fixes left-nested chains, which is the honest scope. `reduceRight`, a deep Union sitting under an Intersection, or a 200-long `↗` chain still recurse. That's OK for AGS's reduce, but a stack-safe interpreter (explicit stack or `cats.Eval`) would be the general fix if this comes back in another shape.

→ Check the test actually fails without the fix on JVM. 2000 levels of `toGeometry` may fit in a default JVM stack; the Chrome failure happened at ~200 because the worker stack was already deep. If it passes pre-fix on JVM it's not a regression test there, only on JS. Also 2000 JTS overlays at "a few ms each" is several seconds added to a shared suite, so 500 would likely prove the point just as well.

Also found: the `tlBaseVersion` bump to 0.248 isn't needed, adding a private[geom] member is binary compatible.

cquiroz:
> It caught my attention that we don't go `@tailrec`, I asked AI and this is it's review:
> 
> Yes, `@tailrec` does the same job with no vars or the descending flag, and I'd push for it. Scala compiles it to the exact same loop, so stack safety is identical, and it reads like the rest of the codebase.
> 
> ```
> private[geom] def leftSpine(e: ShapeExpression, op: BinaryOp): List[ShapeExpression] =
>   @tailrec
>   def go(cur: ShapeExpression, acc: List[ShapeExpression]): List[ShapeExpression] =
>     cur match
>       case Difference(a, b)   if op == BinaryOp.Difference   => go(a, b :: acc)
>       case Intersection(a, b) if op == BinaryOp.Intersection => go(a, b :: acc)
>       case Union(a, b)        if op == BinaryOp.Union        => go(a, b :: acc)
>       case _                                                 => cur :: acc
>   go(e, Nil)
> ```

Yeah it is fine, i don't see a big difference either way, vars are confined to the method but we can use the tail recursive version too, both versions perform the same

> 
> → The `Eq[BinaryOp]` + `===` is noise. A Scala 3 enum already gets `CanEqual`, so plain `==` is safe even under strict equality. Or drop the enum entirely and pass a splitter, `ShapeExpression => Option[(ShapeExpression, ShapeExpression)]`, then go is one split(cur) match with no per-iteration op check.

Right we could use `==`

> 
> → Correctness is fine. Operands come out as [leftmost-leaf, b1 … bn] in evaluation order, and foldLeft reproduces ((a - b) - c) - d, so Difference (non-associative) is preserved. head/tail in foldChain are safe since the list is never empty, but returning (head, rest) from leftSpine would make that explicit and avoid partial calls.
> 
> → It only fixes left-nested chains, which is the honest scope. `reduceRight`, a deep Union sitting under an Intersection, or a 200-long `↗` chain still recurse. That's OK for AGS's reduce, but a stack-safe interpreter (explicit stack or `cats.Eval`) would be the general fix if this comes back in another shape.

This is true and it is intentional by design as ags uses left nested chains, there is no claim otherwise.

> 
> → Check the test actually fails without the fix on JVM. 2000 levels of `toGeometry` may fit in a default JVM stack; the Chrome failure happened at ~200 because the worker stack was already deep. If it passes pre-fix on JVM it's not a regression test there, only on JS. Also 2000 JTS overlays at "a few ms each" is several seconds added to a shared suite, so 500 would likely prove the point just as well.
The test works either way on the jvm, the stack is bigger, but it demonstrably fails in explore



> 
> Also found: the `tlBaseVersion` bump to 0.248 isn't needed, adding a private[geom] member is binary compatible.

hugo-vrijswijk (approved):
LGTM

rpiaggio (approved):
LGTM

PR #1612, opened 2026-09-09
Add pdf generation state enum

cquiroz:


rpiaggio (approved):
LGTM, a nit about naming: `Pending` conveys the idea that the process hasn't started yet. I suggest `Rendering` or `Working` instead.

cquiroz:
Pending and Rendering are different states, in fact they are different in the odb but we don't need as much detail exposed to explore, perhaps `Generating` is a better choice

PR #1614, opened 2026-09-09
Add configuration warning

rpiaggio:
First step to implement https://app.shortcut.com/lucuma/story/6632/implement-checks-for-gnirs-ls.

### DIFF
--- a/modules/core/shared/src/main/scala/lucuma/core/model/ObservationValidation.scala
+++ b/modules/core/shared/src/main/scala/lucuma/core/model/ObservationValidation.scala
@@ -47,3 +47,6 @@ object ObservationValidation:
 
     def lowTotalSignalToNoise(extra: Option[String], minRecommended: TotalSN, actual: TotalSN): ObservationValidation =
       fromMsgs(LowTotalSignalToNoise, f"Total S/N ${extra.foldMap(s => s"($s) ")} is ${actual.value.toBigDecimal}%4.3f (min. ${minRecommended.value.toBigDecimal}%4.3f recommended)")
+
+    def configuration(msg: String, moreMsgs: String*): ObservationValidation =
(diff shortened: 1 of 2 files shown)


modules/core/shared/src/main/scala/lucuma/core/model/ObservationValidation.scala
```diff
@@ -46,4 +46,7 @@ object ObservationValidation:
       fromMsgs(ConditionsUnlikely, s"Conditions likelihood is ${actual.toPercent.toInt}%; minimum recommended is ${recommended.toPercent.toInt}%.")
 
     def lowTotalSignalToNoise(extra: Option[String], minRecommended: TotalSN, actual: TotalSN): ObservationValidation =
-      fromMsgs(LowTotalSignalToNoise, f"Total S/N ${extra.foldMap(s => s"($s) ")} is ${actual.value.toBigDecimal}%4.3f (min. ${minRecommended.value.toBigDecimal}%4.3f recommended)")
\ No newline at end of file
+      fromMsgs(LowTotalSignalToNoise, f"Total S/N ${extra.foldMap(s => s"($s) ")} is ${actual.value.toBigDecimal}%4.3f (min. ${minRecommended.value.toBigDecimal}%4.3f recommended)")
+
+    def configuration(msg: String, moreMsgs: String*): ObservationValidation =
```
cquiroz, line 51: are we going to have more specific warnings or the same for all configuration errors?

rpiaggio: A single warning can have multiple messages.

I evaluated going more granular but leaned towards a single one. This way explore will show a single warning on the observation, holding all the messages. Also, there can be a large number of cases across all modes.

If needed, we can split into specific codes in the future.

cquiroz (approved):
LGTM

PR #1619, opened 2026-09-12
Separate Gnirs acquisition filter lists for spectroscopy and imaging

rpiaggio:
The selectable acquitision filters for GNIRS spectroscopy and imaging are different.


cquiroz (approved):
LGTM

PR #1620, opened 2026-09-14
Get R band from Gaia, needed for Altair

rpiaggio:
Altair needs the R band brightness of the guidestars, which we were not querying until now.

Gaia doesn't have the R information explicitly but publishes a method to comput it from its BP and RP values: https://gea.esac.esa.int/archive/documentation/GDR3/Data_processing/chap_cu5pho/cu5pho_sec_photSystem/cu5pho_ssec_photRelations.html.

The method is slightly different from the one used in OCS since ESA changed it starting with EDR3. This PR implements the latest (since we use DR3) instead of porting the OCS's, which is the one recommended for DR2.


cquiroz:
Does altair need R as defined in the OCS? Could it instead use G directly? I guess that's a question for science and maybe it is not worth doing but we could ask anyway.

rpiaggio:
I had asked Claude for evidence that the OCS only uses R for Altair, and the choice for different method, before sending this PR. This is what it had to say:

---

It derives R from Gaia G and BP−RP with a colour polynomial, exactly the approach I took, and falls back to UCAC4. And "R only" is mostly true for AGS and ITC, but not for the P2 checks or the NGS web docs.

→ How OCS gets R. The OCS Gaia adapter fetches only phot_g_mean_mag and bp_rp, then synthesizes V, R, I, r', i', g', J, H, K for every star using Gaia DR2 polynomials in BP−RP. For R: R = G + 0.003226 − 0.3833 x + 0.1345 x². For Altair only, AGS additionally queries UCAC4, whose fit magnitude is band "UC", a rough R.

→ Same method, newer coefficients. My branch uses the EDR3 relation, which agrees with the DR2 one to about 0.01 mag at BP−RP=1. The DR2 one is a quadratic, EDR3 a quartic with a declared validity range.

** Evidence Altair uses R**

- AltairAowfsGuider.getBands returns RBandsList, the probe's declared band.
- AGS limits CSV has rows Altair NGS,R,17.0 and Altair LGS,R,5.0; the band column drives the magnitude test.
- ITC extractMagnitude for Altair takes r', then R, then UC, and errors with "No r, R or UC magnitude defined for guide star".
- Gemini docs: GNIRS guiding table gives every Altair row in R; LGS limits (17.5/18/18.5 by sky background) are R.

→ Where it is not only R.

- P2 checks use getMinMaxMagnitudeInVorR: R if present, else V.
- NGS web docs give the limiting magnitude table in V by spectral type (11.0 to 15.1 for full/partial correction), while the GNIRS page says R ≲ 12 optimal, R ≲ 15 faint.

Also found: in OCS every single-probe strategy, including PWFS and GMOS OIWFS, used RBandsList. Lucuma moved PWFS/OIWFS to Gaia G; Altair would be the first probe back on a derived band.

---

Furthermore:

❯ give me a link to an explicit doc page where R band is specified for altair

The clearest one is the GNIRS capability page, "Guiding Options" section: https://www.gemini.edu/instrumentation/gnirs/capability

→ What it says, verbatim table rows (Guide Probe | Separation | Guide star brightness, Vega magnitude):

- ALTAIR (NGS) | < 25″ | R <~12 mag (optimal), R <~15 mag (faint limit)
- ALTAIR (LGS) | < 25″ | R <~17.5 mag (bright time), R <~18.5 mag (dark time)
- LGS+PWFS1 | 3.5′ to 7′ | R <~14 mag

Plus in the prose: "ALTAIR can guide on fainter stars (to R~15 mag for A0 stars), but this will result in poor correction."

→ Second source, Altair modes page, LGS section: https://www.gemini.edu/instrumentation/altair/modes

"Stars down to R < 15 within 15 arcsec for high-Strehl LGS correction and R < 18.5 up to 25 arcseconds away," and the sky-background table is titled "LGS tip/tilt guide star R-band magnitude" (R<17.5 Any, 17.5–18 SB80, 18–18.5 SB50).

→ Caveat, same page, NGS section. The NGS limiting-magnitude table by spectral type is given in V (11.0–12.6 full correction, 14.9–16.5 partial), with a "V−R color" row. So for NGS the Altair page itself speaks V while the GNIRS page speaks R. The OCS limit tables and probe band are R throughout, which is what the implementation follows.

modules/catalog/shared/src/main/scala/lucuma/catalog/votable/CatalogAdapter.scala
```diff
@@ -388,20 +388,21 @@ object CatalogAdapter {
 
     protected def brightnessQuery(constraints: Option[BrightnessConstraints]): String = {
       val conditions = constraints.toList.flatMap { bc =>
-        val faintness  = bc.faintnessConstraint.brightness.value.value.toDouble
-        val saturation = bc.saturationConstraint.map(_.brightness.value.value.toDouble)
+        val faintness                = bc.faintnessConstraint.brightness.value.value.toDouble
```
cquiroz, line 391: Will the faintness constraints be applied to R or to each gaia band separately?

modules/catalog-tests/shared/src/test/scala/lucuma/catalog/votable/AdaptersSuite.scala
```diff
@@ -160,4 +165,68 @@ class AdaptersSuite extends CatsEffectSuite with VoTableParser with VoTableSampl
...
+    )
+    val query             = QueryByADQL(
+      Coordinates.Zero,
+      ShapeExpression.centeredEllipse(10.arcseconds, 10.arcseconds),
+      constraints.some
+    )
+    val adql              = summon[ADQLInterpreter].buildQueryString(CatalogAdapter.Gaia3Esa, query)
+    assert(
+      adql.contains(
+        "(phot_rp_mean_mag < 15.000) or (phot_g_mean_mag < 15.000) or (phot_bp_mean_mag < 15.000)"
```
cquiroz, line 232: Why is the same limit applied the same for all bands? Should we have a per band limit instead?

modules/catalog/shared/src/main/scala/lucuma/catalog/GaiaPhotometry.scala
```diff
@@ -0,0 +1,50 @@
...
+ * https://gea.esac.esa.int/archive/documentation/GDR3/Data_processing/chap_cu5pho/cu5pho_sec_photSystem/cu5pho_ssec_photRelations.html
+ *
+ * G - R = -0.02275 + 0.3961 x - 0.1243 x^2 - 0.01396 x^3 + 0.003775 x^4, x = BP - RP
+ */
+object GaiaPhotometry:
+
+  private val GMinusRCoefficients: List[Double] =
+    List(-0.02275, 0.3961, -0.1243, -0.01396, 0.003775)
+
+  val MinBpMinusRp: Double = 0.0
```
cquiroz, line 23: Should these be private?

modules/catalog/shared/src/main/scala/lucuma/catalog/GaiaPhotometry.scala
```diff
@@ -0,0 +1,50 @@
...
+      val (c, i) = ci
+      acc + c * math.pow(bpMinusRp, i.toDouble)
+
+  /**
+   * Smallest and largest G - R over the validity range. A query on G that must not drop any star
+   * whose R falls in a given range has to widen that range by these bounds.
+   */
+  val GMinusRBounds: (Double, Double) =
+    val samples: List[Double] =
+      (0 to 400).toList.map(i => gMinusR(MinBpMinusRp + i * (MaxBpMinusRp - MinBpMinusRp) / 400))
```
cquiroz: Where does the 400 come from?

cquiroz:
Would it make sense to always read the g/rp/bp bands and only calculate R when needed locally? Explore caches the candidates and it seems we wouldn't be able to reuse the cached candidates for both altair and non-altair observations

modules/catalog/shared/src/main/scala/lucuma/catalog/votable/CatalogAdapter.scala
```diff
@@ -388,20 +388,21 @@ object CatalogAdapter {
...
-            saturation match {
-              case Some(sat) => f"($bid between $sat%.3f and $faintness%.3f)"
-              case None      => f"($bid < $faintness%.3f)"
-            }
-          }
+          .collect:
+            case Band.Gaia   => (gMagField.id, faintness, saturation)
+            case Band.GaiaBP => (bpMagField.id, faintness, saturation)
+            case Band.GaiaRP => (rpMagField.id, faintness, saturation)
+            case Band.R      => (gMagField.id, faintness + maxGMinusR, saturation.map(_ + minGMinusR))
```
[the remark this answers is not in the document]
rpiaggio: Valid. Fixed in new commit.

modules/catalog/shared/src/main/scala/lucuma/catalog/GaiaPhotometry.scala
```diff
@@ -0,0 +1,50 @@
...
+    (samples.min, samples.max)
```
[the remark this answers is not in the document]
rpiaggio: Valid. Fixed in new commit.

modules/catalog/shared/src/main/scala/lucuma/catalog/votable/CatalogAdapter.scala
```diff
@@ -388,20 +388,21 @@ object CatalogAdapter {
...
+          .map: (bid, faint, sat) =>
+            sat match
+              case Some(sat) => f"($bid between $sat%.3f and $faint%.3f)"
+              case None      => f"($bid < $faint%.3f)"
```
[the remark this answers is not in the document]
rpiaggio: Valid. Fixed in new commit.

modules/catalog/shared/src/main/scala/lucuma/catalog/votable/CatalogAdapter.scala
rpiaggio: To whatever bands the constraint's list names, with one limit per constraint.

An R constraint (`RBandsList`) gets translated to a widened G range plus the colour predicate; the exact R test happens client side.

modules/catalog/shared/src/main/scala/lucuma/catalog/GaiaPhotometry.scala
rpiaggio: They should have been in the commit you reviewed, but not anymore after the fixes to copilot's comments.

modules/catalog/shared/src/main/scala/lucuma/catalog/GaiaPhotometry.scala
rpiaggio: The colour (spread between BP and RP mags) is limited to 4. This sampled every 0.01, so that's 400 samples. It has changed now to address one of copilot's comments.

modules/catalog-tests/shared/src/test/scala/lucuma/catalog/votable/AdaptersSuite.scala
rpiaggio: I'll let Claude answer this one:

That's pre-existing behaviour, the test just makes the query visible. The AGS faint limit is a single number computed in Gaia G, and a candidate may carry RP, G or BP depending on what Gaia has for it, so the query applies the one limit to whichever column is present (OR) rather than dropping stars missing a band. Per-band limits would need both a per-band structure in BrightnessConstraints and per-band formulas, and we only have G (and now R, handled on its own path).

We could address this in another PR. It's not trivial, so only if we really need it.

rpiaggio:
> Would it make sense to always read the g/rp/bp bands and only calculate R when needed locally? Explore caches the candidates and it seems we wouldn't be able to reuse the cached candidates for both altair and non-altair observations

This is a good idea. I think nobody uses the brightnesses apart from AGS, so I think it can work. I'll add a commit.

rpiaggio:
> > Would it make sense to always read the g/rp/bp bands and only calculate R when needed locally? Explore caches the candidates and it seems we wouldn't be able to reuse the cached candidates for both altair and non-altair observations
> 
> This is a good idea. I think nobody uses the brightnesses apart from AGS, so I think it can work. I'll add a commit.

Note that current cache entries will still be invalidated, since we need now the BP and RP bands which were not present before.

rpiaggio:
> > > Would it make sense to always read the g/rp/bp bands and only calculate R when needed locally? Explore caches the candidates and it seems we wouldn't be able to reuse the cached candidates for both altair and non-altair observations
> > 
> > 
> > This is a good idea. I think nobody uses the brightnesses apart from AGS, so I think it can work. I'll add a commit.
> 
> Note that current cache entries will still be invalidated, since we need now the BP and RP bands which were not present before.

Also, this will be implemented as a new commit in https://github.com/gemini-hlsw/lucuma-core/pull/1621, since that's were we changed the logic to store the value, not here.

modules/catalog-tests/shared/src/test/scala/lucuma/catalog/votable/AdaptersSuite.scala
cquiroz: Now i remember why, we had a discussion about it but it is no obvious what are the limits per band and 15 is conservative anyway

cquiroz:
> > This is a good idea. I think nobody uses the brightnesses apart from AGS, so I think it can work. I'll add a commit.
> 
> Note that current cache entries will still be invalidated, since we need now the BP and RP bands which were not present before.

That's fine we should invalidate the cache and use an all-bands cache instead

cquiroz (approved):
Thanks for checking this, looks great. I'm not sure if we have yet addressed the use of a single or multiple caches but perhaps that's something to be done in explore?

At any rate i don't want to hold this any longer. LGTM

PR #1621, opened 2026-09-14
ALTAIR AGS

rpiaggio:
Includes the broadening of the logic to consider R band, thus requiring https://github.com/gemini-hlsw/lucuma-core/pull/1620.


modules/ags/src/main/scala/lucuma/ags/Ags.scala
```diff
@@ -40,13 +40,22 @@ object Ags {
     calcsNanos:           Long // time spent in posCalculations
   )
 
+  private def satisfies(constraint: BrightnessConstraints, candidate: GuideStarCandidate): Boolean =
+    candidate
+      .brightnessIn(constraint.searchBands)
```
cquiroz, line 45: I guess this allows to search either by the R band or the G bands?

modules/ags/src/main/scala/lucuma/ags/Ags.scala
```diff
@@ -131,13 +142,13 @@ object Ags {
       )
     }
 
-    // Do we have a g magnitude
-    guideStar.gBrightness match {
-      case Some((_, g)) =>
-        guideSpeedFor(speeds, g)
+    // Do we have a magnitude in the probe's band
```
cquiroz, line 145: In this case we'd say that the altair probe uses the R band? while let's say gmos oi uses the G bands?

modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
```diff
@@ -522,6 +527,35 @@ object AgsParams:
     override def probeArm(posAngle: Angle, guideStar: Offset, offset: Offset): ShapeExpression =
       lucuma.core.geom.pwfs.probeArm.vignettedAreaAt(probe, guideStar, offset)
 
+  trait AltairSupport[A]:
+    def withAltair(mode: AltairMode): A
+
+  /**
+   * Instruments behind Altair guide with a PWFS or, through Altair, its AOWFS. The AOWFS picks the
+   * star off behind a dichroic, so it never vignettes the science field.
```
cquiroz, line 540: If it never vignettes we could skip the calculation right?

modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
```diff
@@ -543,16 +577,22 @@ object AgsParams:
...
-      with PwfsSupport[GnirsLongSlit] derives Eq:
+      with AltairCapableParams
+      with PwfsSupport[GnirsLongSlit]
+      with AltairSupport[GnirsLongSlit] derives Eq:
+
+    protected def withPWFSProbe(probe: PWFSGuideProbe): GnirsLongSlit =
+      copy(probe = probe, altair = None)
 
-    protected def withPWFSProbe(probe: PWFSGuideProbe): GnirsLongSlit = copy(probe = probe)
+    def withAltair(mode: AltairMode): GnirsLongSlit =
```
cquiroz, line 599: neat

modules/ags/src/main/scala/lucuma/ags/GuideStarCandidate.scala
```diff
@@ -34,16 +35,22 @@ import java.time.ZoneId
...
+ * Brightnesses always include the R estimated from the Gaia bands when the colour allows it, so a
+ * candidate is a fixed point of that estimation.
  */
 case class GuideStarCandidate private (
-  id:          Long,
-  tracking:    SiderealTracking,
-  gBrightness: Option[(Band, BrightnessValue)]
+  id:           Long,
+  tracking:     SiderealTracking,
+  brightnesses: SortedMap[Band, BrightnessValue]
```
cquiroz, line 45: The original reason to do this was to save memory (and cache space) not storing every band when we only care about one. We could store just one though having two is perhaps more efficient for the cache?

modules/catalog/shared/src/main/scala/lucuma/catalog/GaiaPhotometry.scala
```diff
@@ -48,3 +52,16 @@ object GaiaPhotometry:
     Option
       .when(colour > MinBpMinusRp && colour < MaxBpMinusRp)(colour)
       .flatMap(c => BrightnessValue.from(g.value.value - BigDecimal(gMinusR(c))).toOption)
+
+  /**
+   * Adds an estimated R to a set of brightnesses that has G, BP and RP but no R of its own. A
+   * catalog R is never overwritten.
+   */
+  def withEstimatedR(
```
cquiroz: 👍

modules/tests/shared/src/test/scala/lucuma/core/geom/altair/AltairPatrolFieldSuite.scala
```diff
@@ -0,0 +1,45 @@
...
+package lucuma.core.geom.altair
+
+import lucuma.core.enums.GuideProbe
+import lucuma.core.geom.ScienceAreaGeometrySuite
+import lucuma.core.geom.jts.interpreter.given
+import lucuma.core.geom.syntax.all.*
+import lucuma.core.math.Angle
+import lucuma.core.math.Offset
+
+class AltairPatrolFieldSuite extends ScienceAreaGeometrySuite:
```
cquiroz, line 13: These seems kind of trivial tests, could we rather do a jtsdemo for altair to display it graphically?

modules/ags/src/main/scala/lucuma/ags/GuideStarCandidate.scala
rpiaggio: Well, we could either store G, RP and BP and compute R on the fly; or we can store G and the computed R.

modules/ags/src/main/scala/lucuma/ags/Ags.scala
rpiaggio: Yes

modules/ags/src/main/scala/lucuma/ags/Ags.scala
rpiaggio: Exactly

modules/ags/src/main/scala/lucuma/ags/AgsParams.scala
rpiaggio: Good idea. Sending new commit.

modules/tests/shared/src/test/scala/lucuma/core/geom/altair/AltairPatrolFieldSuite.scala
rpiaggio: Added to the demo, but I'd rather keep the test just in case.

modules/ags/src/main/scala/lucuma/ags/GuideStarCandidate.scala
rpiaggio: We could store `Int` millimagnitudes instead of `BigDecimal`s. This can save a few bytes per candidate (~20 bytes for the 3 values). Anyway, the gain would be very small since what dominates the storage of each candidate is all the tracking infromation. Not sure it's worth it.

cquiroz: if we store only either G or R we can't reuse the cache, if the difference isn't that big maybe storing the map is ok

modules/tests/shared/src/test/scala/lucuma/core/geom/altair/AltairPatrolFieldSuite.scala
cquiroz: Can you put a screenshot of the demo?

cquiroz (approved):
LGTM

cquiroz (approved):
LGTM

modules/tests/shared/src/test/scala/lucuma/core/geom/altair/AltairPatrolFieldSuite.scala
rpiaggio: image

→ What each shape is (GNIRS long slit behind Altair NGS, PA 0 plus its 180 flip, ABBA nods of ±5" along the slit):

1 → Black circle, 27" radius. The AOWFS candidates area: a circle covering the patrol oval at every PA, used for catalog searches. Same 27" as the oval's positive-q limit.

2 → Two magenta solid ovals. The patrol field at the two base positions (PA 0 and PA 180). The oval is 25" wide in p, 27" toward +q, 23" toward -q, so the flip shifts it visibly up and down.

3 → Blue dashed ovals, six of them. The patrol field at each guided science offset: +5" and -5" q, at both PAs. Several overlap, so they read as three or four.

4 → Green thin oval. The intersection of all patrol fields for one PA, the region where a guide star stays reachable through the whole nod sequence. Spans about -22" to +22" in q.

5 → Black vertical bar. GNIRS science area: the 0.30" long slit, 49" long, at PA 0.

6 → Red square at p=12", q=18". The guide star, drawn as a 1" box.

Also found: the guide star sits outside the green intersection. It is inside the base ovals but not reachable at all nod positions, so if you expect AGS to accept it, it won't. I'm sending a new PR to fix this so that it renders a GS within the patrol field.

PR #1623, opened 2026-09-15
Add calibrationCount field

cquiroz:
We want to expose how many calibrations (that are charged) will be in the total time estimate.
We need to eventually update the total estimate but not yet


modules/tests/shared/src/test/scala/lucuma/core/model/sequence/ExecutionDigestSuite.scala
```diff
@@ -4,13 +4,19 @@
...
-
-  checkAll("Eq[ExecutionDigest]", EqTests[ExecutionDigest].eqv)
-
-}
+  checkAll("Eq[ExecutionDigest]",              EqTests[ExecutionDigest].eqv)
+  checkAll("ExecutionDigest.setup",            LensTests(ExecutionDigest.setup))
+  checkAll("ExecutionDigest.setupCount",       LensTests(ExecutionDigest.setupCount))
+  checkAll("ExecutionDigest.calibrationCount", LensTests(ExecutionDigest.calibrationCount))
+  checkAll("ExecutionDigest.acquisition",      LensTests(ExecutionDigest.acquisition))
+  checkAll("ExecutionDigest.science",          LensTests(ExecutionDigest.science))
```
rpiaggio, line 22: The lens tests are not necessary for generated lenses.

Or for a derived `Eq` (not the case but we could derive it).

rpiaggio (approved):
LGTM, I'd remove the unnecessary tests

modules/tests/shared/src/test/scala/lucuma/core/model/sequence/ExecutionDigestSuite.scala
cquiroz: Yeah I think your' right

PR #1625, opened 2026-09-16
Model additions the ODB needs for Altair on GNIRS

rpiaggio:
- AltairNdFilter enum (In, Out): Altair's neutral density filter position, stored per observation.
- probes: Altair-aware overloads of allowedProbes, isProbeAllowed and defaultGuideProbe taking an Option[AltairMode]. With Altair on a GNIRS mode the only allowed probe is the mode's wavefront sensor (AltairAOWFS for NGS/LGS, PWFS1 for LGS+P1). Altair on any other instrument allows no probe. None keeps the existing behaviour.
- ProbesSuite cases for the new overloads.


modules/core/shared/src/main/scala/lucuma/core/model/probes.scala
```diff
@@ -34,9 +36,24 @@ trait probes:
...
+      ObservingModeType.toFacility.getOption(observingMode).exists(_.instrument === Instrument.Gnirs)
+    altair match
+      case Some(mode) if isGnirs => SortedSet(mode.guideProbe)
+      case Some(_)               => SortedSet.empty
+      case None                  => allowedProbes(observingMode)
+
   def isProbeAllowed(observingMode: ObservingModeType, probe: GuideProbe): Boolean =
     allowedProbes(observingMode).contains(probe)
 
+  def isProbeAllowed(observingMode: ObservingModeType, altair: Option[AltairMode], probe: GuideProbe): Boolean =
```
cquiroz, line 54: Should we have just a single method with altair defaulting to None? just a small detail

cquiroz (approved):

modules/core/shared/src/main/scala/lucuma/core/model/probes.scala
rpiaggio: I saw this comment after merging. I agree. Fixing in new PR.

PR #1626, opened 2026-09-18
Use Map for Enumerated `fromTag` and `Decoder` lookup

hugo-vrijswijk:


modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
```diff
@@ -82,9 +86,8 @@ object Enumerated:
   private def enumeratedImpl[E: Type](using Quotes): Expr[Enumerated[E]] =
     '{
       Enumerated
-        .fromNEL(NonEmptyList.fromList(${ enumValuesImpl[E] }.toList).get)
+        .fromNEL(NonEmptyList.fromListUnsafe(${ enumValuesImpl[E] }.toList))
```
cquiroz, line 89: Could this unsafe ever fail or is the list never empty?

cquiroz (approved):
LGTM

modules/core/shared/src/main/scala/lucuma/core/util/Enumerated.scala
hugo-vrijswijk: Maybe for an Enumerated without instances? If possible at all it would fail compile-time, which might be wanted for an Enumerated without values anyway

PR #1627, opened 2026-09-18
Add Altair to configuration request

rpiaggio:
Adding or removing Altair or changing its parameters will change the requested time, so it invalidates an accepted configuration request.


cquiroz (approved):
LGTM

